Issue #43, opened 2023-05-27
Choose the serialization method based on content type


czechboy0:
Currently, the generator chooses the serialization method (JSON, text-based, or raw data) based on the type used in the request/response body. Most of the time, that works correctly, but there are some corner cases, such as:
- content type: application/json, the body is a fragment string (as opposed to a JSON container, like an object or an array).

In this case, we today serialize it using `LosslessStringConvertible`, and end up _without_ the quotes. That's only correct if the content type is `text/plain` (and in fact was just fixed in https://github.com/apple/swift-openapi-runtime/pull/9).

Action: add logic for choosing the serialization method based on the content type, instead of trying to deduce it from the Swift type.

> Whilst implementing this change, I think that we may want to revisit the encoding logic, since I think the knowledge of whether to encode a value as binary data, JSON, or plain text, is at the place where the content-type is set, but we'll defer that to a subsequent PR.

From: https://github.com/apple/swift-openapi-runtime/pull/10

PR #50, opened 2023-06-06
[Docs] Document the Converter type and all the cases it needs to support

czechboy0:
### Motivation

I wrote down how the Converter works and all the cases we need to support. That'll help prepare ground for the refactor needed for #43.

Also created space for maintainer-focused docs.

### Modifications

Added docs, which I'll use in a subsequent PR to ensure all the cases listed are implemented and tested.

### Result

Converter now has higher level docs, and maintainers have a place to add more docs.

### Test Plan

Previewed locally.

### DIFF
--- a/Sources/swift-openapi-generator/Documentation.docc/Development/Converting-between-data-and-Swift-types.md
+++ b/Sources/swift-openapi-generator/Documentation.docc/Development/Converting-between-data-and-Swift-types.md
@@ -19,7 +19,7 @@ Some helper methods can be reused between client and server code, such as header
 
 As outlined in <doc:Project-scope-and-goals>, we aim to minimize the complexity of the generator and rely on the Swift compiler to help ensure that if generated code compiles, it's likely to work correctly.
 
-To that end, if the input OpenAPI document contains an input that Swift OpenAPI Generator doesn't support, our first preference is to catch it in the generator and emit a descriptive error. However, there are cases where that is prohibitively complex, and we let the Swift compiler ensure that, for example, an array of strings cannot be used as a path parameter. In this example case, the generator emits code with the path parameter being of Swift type `[String]`, but since there doesn't exist a converter method for it, it will fail to build. This is considered expected behavior.
+To that end, if the input OpenAPI document contains an input that Swift OpenAPI Generator doesn't support, our first preference is to catch it in the generator and emit a descriptive diagnostic. However, there are cases where that is prohibitively complex, and we let the Swift compiler ensure that, for example, an array of strings cannot be used as a path parameter. In this example case, the generator emits code with the path parameter being of Swift type `[String]`, but since there doesn't exist a converter method for it, it will fail to build. This is considered expected behavior.
 
 In the case of the converter, it contains helper methods for all the supported combinations of an HTTP location, a "content type family" and a Swift type.
 
@@ -58,7 +58,7 @@ func headerFieldGetTextRequired( // <<< 1.
 ) throws -> Date
 ```
 
-In `1.`, notice that the method name contains which schema location, content type family, and optionality; whilie in `2.` it contains the Swift type.
+In `1.`, notice that the method name contains which schema location, content type family, and optionality; while in `2.` it contains the Swift type.
 
 ### Helper method variants
 
--- a/Sources/swift-openapi-generator/Documentation.docc/Development/Converting-between-data-and-Swift-types.md
+++ b/Sources/swift-openapi-generator/Documentation.docc/Development/Converting-between-data-and-Swift-types.md
@@ -1,10 +1,10 @@
 # Converting between data and Swift types
 
-Learn about the type responsible for convertering between raw data and Swift types.
+Learn about the type responsible for convertering between binary data and Swift types.
 
 ## Overview
 
-The [`Converter`](https://github.com/apple/swift-openapi-runtime/blob/main/Sources/OpenAPIRuntime/Conversion/Converter.swift) type is a structure defined in the runtime library and is used by both the client and server generated code to perform conversions between raw data and Swift types.
+The [`Converter`](https://github.com/apple/swift-openapi-runtime/blob/main/Sources/OpenAPIRuntime/Conversion/Converter.swift) type is a structure defined in the runtime library and is used by both the client and server generated code to perform conversions between binary data and Swift types.
 
 > Note: `Converter` is one of the SPI types, not considered part of the public API of the runtime library. However, because generated code relies on it, SPI stability needs to be considered when making changes to it and to the generator.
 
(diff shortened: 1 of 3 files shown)


Sources/swift-openapi-generator/Documentation.docc/Development/Converting-between-data-and-Swift-types.md
```diff
@@ -0,0 +1,234 @@
...
+
+### Generated code and generics interaction
+
+As outlined in <doc:Project-scope-and-goals>, we aim to minimize the complexity of the generator and rely on the Swift compiler to help ensure that if generated code compiles, it's likely to work correctly.
+
+To that end, if the input OpenAPI document contains an input that Swift OpenAPI Generator doesn't support, our first preference is to catch it in the generator and emit a descriptive error. However, there are cases where that is prohibitively complex, and we let the Swift compiler ensure that, for example, an array of strings cannot be used as a path parameter. In this example case, the generator emits code with the path parameter being of Swift type `[String]`, but since there doesn't exist a converter method for it, it will fail to build. This is considered expected behavior.
```
simonjbeaumont, line 22: > "...emit a descriptive error..."

Don't we usually emit a diagnostic and try to continue?

Sources/swift-openapi-generator/Documentation.docc/Development/Converting-between-data-and-Swift-types.md
```diff
@@ -0,0 +1,234 @@
...
+
+```swift
+func headerFieldGetTextRequired( // <<< 1.
+    in headerFields: [HeaderField],
+    name: String,
+    as type: Date.Type // <<< 2.
+) throws -> Date
+```
+
+In `1.`, notice that the method name contains which schema location, content type family, and optionality; whilie in `2.` it contains the Swift type.
```
simonjbeaumont, line 61:
```suggestion
In `1.`, notice that the method name contains which schema location, content type family, and optionality; while in `2.` it contains the Swift type.
```

Sources/swift-openapi-generator/Documentation.docc/Development/Converting-between-data-and-Swift-types.md
```diff
@@ -0,0 +1,234 @@
...
+Together, the schema location, the content type family, and the Swift type is enough to unambiguously decide which helper method on the converter should be used.
+
+For example, to use the converter to get a required response header of type `Foundation.Date` using the `text` content type family, look for a method (exact spelling is subject to change) that looks like:
+
+```swift
+func headerFieldGetTextRequired( // <<< 1.
+    in headerFields: [HeaderField],
+    name: String,
+    as type: Date.Type // <<< 2.
+) throws -> Date
```
simonjbeaumont, line 58: While we might want to have the combinatorial explosion of methods internally, could they all be dispatched from one to help reduce the complexity in the generator.

I haven't fully thought this through, but something like:

```swift
func getParameter<T>(
    name: String,    
    location: ParameterLocation,
    required: Bool,
    as type: T.Type = T.self
) throws -> T

func getBody<T>(
    name: String,
    required: Bool,
    contentType: ContentType,
) throws -> T
```

Where `ParameterLocation` and `ContentType` are enum-like types?

simonjbeaumont (commented):
Thanks for starting this document @czechboy0, it's great to get some of these maintainer-focused docs going.

I had some questions about whether we need to have the combinatorial explosion of methods that all need to be individually targeted by the generated code or whether we could manage to have something more general-purpose, which might farm out to other private things. If it doesn't have too much of an overhead, it might simplify the generated code and the generator a bit?

Sources/swift-openapi-generator/Documentation.docc/Development/Converting-between-data-and-Swift-types.md
czechboy0: Good point, let me update the wording to reflect that.

Sources/swift-openapi-generator/Documentation.docc/Development/Converting-between-data-and-Swift-types.md
czechboy0: 🎡 😄

Sources/swift-openapi-generator/Documentation.docc/Development/Converting-between-data-and-Swift-types.md
czechboy0: Yeah, this is similar to the approach I started with in https://github.com/apple/swift-openapi-runtime/pull/12 with the "coding strategy" approach, however there are a few issues I ran into:

1. `required` can't be a parameter, because the return value has to be `T?` for optional and `T` for required parameters/bodies. I played with it today and couldn't find a way to express that in Swift in a way that's clearer than just having two variants.
2. Bodies are tricky - they're optional or required in requests, but always required in responses.
3. There are several invalid combinations of inputs here, that we don't want to have to emit errors for, or add precondition failures, and if uncovered by tests, they'll make it difficult for us to have high test coverage.

For example, given:

```swift
struct Foo: Codable, ... { }
```

You could call:

```swift
getParameter(name: "mumble", location: .path, required: ..., as: Foo.self)
```

But path + structured isn't a supported combination.

Where possible, I think it's best to not have codepaths that will never be hit, as it has a pretty high cost on testing, creating error types that'll never be thrown, and it makes the methods seem more flexible than they really are (if half of the codepaths in the method above throw/crash, it won't be fun to use).

That said – this is because we don't support every possible combination. If we did, using more parameters than specialized methods would make sense, so maybe if we fill in that support matrix in the future, we could consider something like this.

One extra benefit of the combinatorial explosion of methods: more room for the compiler to optimize them away.

Now, this is going too much into the proposed implementation, which is actually independent of this PR. This PR is mostly about the requirements (and yes, this example above references the existing implementation), which I actually think is harder to nail down than the implementation. 🙂 

So looking at the support matrix, does that cover our needs, you think? If so, we can iterate on the proposed implementations separately (as mentioned, #48 was my... err... second attempt at this, so I figured we first agree on requirements, that'll help constrain the solution).

simonjbeaumont: > Now, this is going too much into the proposed implementation, which is actually independent of this PR. This PR is mostly about the requirements (and yes, this example above references the existing implementation), which I actually think is harder to nail down than the implementation. 🙂

It wasn't clear to me that this was _only_ requirements based on the way the documentation is written.here's a note that the spelling could change but the idea that all this needed to be in the runtime SPI, and that the generator needing to find the precise sounded final, which I think does warrant some discussion.

> So looking at the support matrix, does that cover our needs, you think? If so, we can iterate on the proposed implementations separately (as mentioned, #48 was my... err... second attempt at this, so I figured we first agree on requirements, that'll help constrain the solution).

Let me read through the list below and comment there.

 > Where possible, I think it's best to not have codepaths that will never be hit, as it has a pretty high cost on testing, creating error types that'll never be thrown, and it makes the methods seem more flexible than they really are (if half of the codepaths in the method above throw/crash, it won't be fun to use).
> 
> That said – this is because we don't support every possible combination. If we did, using more parameters than specialized methods would make sense, so maybe if we fill in that support matrix in the future, we could consider something like this.

I think whatever design we come up with should be forward looking. On the subject of paths you cannot reach, I'm all for that where practical. However, I don't see too much harm in us having some common _entry point_ to this functionality to simplify the generator. Any combination of inputs that's invalid will be a programmer error on the part of the generator (distinct from a runtime error) and so can be checked with a `precondition`. This pattern seems common in the state machine code in many of the NIO projects.
 
> One extra benefit of the combinatorial explosion of methods: more room for the compiler to optimize them away.

This is a good motivation for our requirements 👍. That said, we should understand the tradeoff more qualitatively. In general I've found the compiler is smarter than me. Be good to understand the cost here before we eagerly optimise at the cost of complexity.

Sources/swift-openapi-generator/Documentation.docc/Development/Converting-between-data-and-Swift-types.md
```diff
@@ -0,0 +1,234 @@
...
+               - optional/required
+           - array of string-convertibles
+               - optional/required
+           - date
+               - optional/required
+           - array of dates
+               - optional/required
+       - structured
+           - codable
+               - optional/required
```
simonjbeaumont, line 101: Should I always read `codable` as JSON? 

Are there other ways we can structure data that we want to support? If so, should we spell our API more explicitly to accommodate this, even if it's not going to be in the initial supported things?

czechboy0: We only support JSON today, but OpenAPI also supports XML in a very similar capacity. The good news is that we should be able to use Codable regardless of whether the type is serialized as JSON or XML, thus the terminology being focused on the Codable type, rather than the serialized content type.

simonjbeaumont: OK, I suppose this is more a question of the API of a `ContentType`, if we have one. It should probably be expressed as a `.json` and not `.codable` if we intend to be more expressive?

czechboy0: One more thought: if we do want to reuse the same method with a parameter, maybe the methods _should_ have the word "structured" in them, so that we can later go from:

```swift
getParamStructured<T: Codable>(...) -> T
```

to

```swift
getParamStructured<T: Codable>(format: .json, ...) -> T
getParamStructured<T: Codable>(format: .xml, ...) -> T
```

So the full content type family would be:

- JSON
- XML
- text
- raw

But JSON and XML would share the helper methods. Does that make sense?

simonjbeaumont (approved):

PR #59, opened 2023-06-09
Support empty string OpenAPI identifiers

### LINKED ISSUE #33: Empty string enum value breaks the generator
We've seen a string enum definition like:

```yaml
type: string
enum:
  - ""
  - foo
  - bar
```

Note the empty string as the first case.

Currently, the generator breaks on it, but we should be able to handle it. Somewhat related to #21.

simonjbeaumont:
### Motivation

The OpenAPI specification allows the use of the empty string in some places, for example in the following enum, which we've seen in some real-world OpenAPI documents:

```yaml
type: string
enum:
  - ""
  - foo
  - bar
```

We have code that converts an OpenAPI identifier to a suitable string that can be used as a Swift identifier. It currently returns an empty string when given an empty string as input, which is not a valid Swift identifier. We already have some logic there to avoid clashing with some Swift keywords, so this can be extended to handle the empty string.

### Modifications

- Map the empty string OpenAPI identifier to "_empty"

### Result

We can generate compiling code when presented with empty string OpenAPI identifiers.

### Test Plan

- Added unit test.
- Extended reference test.

### Resolves

- Resolves #33.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Extensions/String.swift
+++ b/Sources/_OpenAPIGeneratorCore/Extensions/String.swift
@@ -63,7 +63,7 @@ fileprivate extension String {
     /// ensures that the identifier starts with a letter and not a number.
     var sanitizedForSwiftCode: String {
         guard !isEmpty else {
-            return self
+            return "_empty"
         }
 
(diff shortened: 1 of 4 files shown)


czechboy0 (approved):
🙏

PR #63, opened 2023-06-12
Relax minimum macOS requirement to 10.15

### LINKED ISSUE #62: Support platform versions back to macOS 10.15,  iOS 13.0, tvOS 13.0, and watchOS 6.0
Let's see how much work it'd be to reduce the deployment targets, macOS 13->11, and so on.

This would be all of the repos - generator, runtime, and transports.

// Edit: Possibly we could go as low as `.macOS(.v10_15), .iOS(.v13), .tvOS(.v13), .watchOS(.v6)`.

andrewse02:
### Motivation

Reduce deployment target by 3 versions to gain more adoption.

### Modifications

- Reduced deployment targets by 3 versions
- Modified `client.Package.x` and `server.Package.x` files to reflect new deployment targets
- Modified documentation to reflect new deployment targets

### Result

Deployment target will be reduced by 3 versions and documentation will reflect such changes.

### Resolves

- Resolves #62

### Related PRs

- Paired with https://github.com/apple/swift-openapi-runtime/pull/18

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -61,7 +61,7 @@ let package = Package(
         ),
 
         // Tests-only: Runtime library linked by generated code
-        .package(url: "https://github.com/apple/swift-openapi-runtime", .upToNextMinor(from: "0.1.1")),
+        .package(url: "https://github.com/apple/swift-openapi-runtime", .upToNextMinor(from: "0.1.3")),
 
         // Build and preview docs
         .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"),
--- a/Package.swift
+++ b/Package.swift
@@ -17,7 +17,7 @@ import PackageDescription
 let package = Package(
     name: "swift-openapi-generator",
     platforms: [
-        .macOS(.v13)
+        .macOS(.v10_15)
     ],
     products: [
         .executable(name: "swift-openapi-generator", targets: ["swift-openapi-generator"]),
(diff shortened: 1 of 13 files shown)


czechboy0 (approved):

czechboy0:

czechboy0 (changes requested):
Actually, please also update the README and Documentation.md with these new deployment versions.

andrewse02:
Will do. Also noticed that the `client.Package.x` files' deployment targets can be reduced as well. Adding that too.

czechboy0 (approved):

czechboy0 (changes requested):
Please also bump the dependency on the runtime library to 0.1.3, which also has the reduced versions.

andrewse02:
@czechboy0 Sure! Just for future PRs, is that something I should have had in there from the start? Since this one was paired with a runtime PR should I have known to increase the version number based on the `semver/X` label?

simonjbeaumont:
> Just for future PRs, is that something I should have had in there from the start? Since this one was paired with a runtime PR should I have known to increase the version number based on the `semver/X` label?

Assuming you mean the bumping of the runtime dependency version, then you couldn't have done it from the start because we needed to publish the release first, so you didn't miss any steps :)

It's masked because the dependency on the runtime is only consumed in a test target within the generator, but if it was consumed in a library or executable target then the CI would have failed.

We're just wanting to make sure we do a coordinated rollout in this instance.

As to whether you'll need to always bump the version of the runtime library: not really, it depends on the change.

Thanks for your contribution!

czechboy0 (approved):

simonjbeaumont (approved):

PR #67, opened 2023-06-14
Use operation summary and description in generated function docs

simonjbeaumont:
### Motivation

For generated functions for each operations we generate reference
documentation. Currently this only states the HTTP information and the
operation ID, and doesn't include either the `summary` or the `description`
that may be present in the OpenAPI document.

### Modifications

Change the reference documentation to include both the summary and the
description if they are provided.

### Result

Reference documentation for operation functions contains the documentation from
the OpenAPI document.

### Test Plan

The OpenAPI document used in the reference test includes operations with and
without these fields.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/CommonTypes/CommentExtensions.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/CommonTypes/CommentExtensions.swift
@@ -133,31 +133,26 @@ extension ResponseKind {
 
 extension Comment {
 
-    /// Returns a documentation comment for an OpenAPI operation.
+    /// Returns a reference documentation string to attach to the generated function for an operation.
     ///
-    /// For example: "Operation `getPet` performs `GET` on `/pets/{petId}`".
     /// - Parameters:
-    ///   - operationID: The identifier of the OpenAPI operation.
-    ///   - method: The HTTP method of the OpenAPI operation.
-    ///   - path: The URL path of the OpenAPI operation.
-    static func operation(
-        operationID: String?,
-        method: OpenAPI.HttpMethod,
-        path: OpenAPI.Path
-    ) -> Self {
-        let operationName: String
-        if let operationID {
-            operationName = "`\(operationID)` "
-        } else {
-            operationName = ""
-        }
+    ///   - operationDescription: The OpenAPI operation description.
+    init(from operationDescription: OperationDescription) {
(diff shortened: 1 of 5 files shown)


Sources/_OpenAPIGeneratorCore/Translator/CommonTypes/CommentExtensions.swift
```diff
@@ -133,31 +133,26 @@ extension ResponseKind {
...
-        path: OpenAPI.Path
-    ) -> Self {
-        let operationName: String
-        if let operationID {
-            operationName = "`\(operationID)` "
-        } else {
-            operationName = ""
-        }
+    ///   - operationDescription: The OpenAPI operation description.
+    init(from operationDescription: OperationDescription) {
```
czechboy0, line 140: I'd prefer if we provided the necessary values explicitly, i.e. by adding `summary` and `description` as optional String parameters, instead of requiring a whole `OperationDescription`. That'll preserve its ability to be easily unit tested, whereas providing a whole `OperationDescription` would require bringing up a bunch of unnecessary state in tests.

But happy to go from a static method to an initializer, that part is fine.

czechboy0 (changes requested):

Sources/_OpenAPIGeneratorCore/Translator/CommonTypes/CommentExtensions.swift
simonjbeaumont: Hm, I was trying to optimise the call site but empathise with making it more testable. How about we provide the method you suggest, but then keep the type conversion initializer as a convenience:

```swift
init(summary: String?, description: String?, path: OpenAPI.Path, httpMethod: OpenAPI.HttpMethod, openAPIDocPath: String) { ... }

init(from operationDescription: OperationD) {
    .init(...)
}
```

czechboy0: Sure, sounds good. I previously didn't optimize the call site as there's only one place we call it.

simonjbeaumont: Addressed in https://github.com/apple/swift-openapi-generator/pull/67/commits/0a594413fe9c7722083e549dba82f869eefa7c5b

Issue #51, opened 2023-06-07
Issue with paths ending with a slash


miguel-arrf:
**Context:**
Hi! I was testing creating a Client with the following `openapi.yaml` file:

```
openapi: 3.0.2
info:
  title: FastAPI
  description: Example app
  version: 0.1.0
servers:
  - url: http://127.0.0.1:8000
    description: Localhost deployment.
paths:
  /items/:
    get:
      tags:
        - Items
      summary: Read Items
      operationId: read_items_items__get
      responses:
        "200":
          description: Successful Response
          content:
            application/json:
              schema:
                title: Response Read Items Items  Get
                type: array
                items:
                  $ref: "#/components/schemas/Item"
    post:
      tags:
        - Items
      summary: Create Item
      operationId: create_item_items__post
      requestBody:
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/Item"
        required: true
      responses:
        "200":
          description: Successful Response
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Item"
        "422":
          description: Validation Error
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/HTTPValidationError"
components:
  schemas:
    HTTPValidationError:
      title: HTTPValidationError
      type: object
      properties:
        detail:
          title: Detail
          type: array
          items:
            $ref: "#/components/schemas/ValidationError"
    Item:
      title: Item
      required:
        - name
        - price
      type: object
      properties:
        name:
          title: Name
          type: string
        description:
          title: Description
          type: string
        price:
          title: Price
          type: number
        tax:
          title: Tax
          type: number
        tags:
          title: Tags
          type: array
          items:
            type: string
          default: []
    ValidationError:
      title: ValidationError
      required:
        - loc
        - msg
        - type
      type: object
      properties:
        loc:
          title: Location
          type: array
          items:
            anyOf:
              - type: string
              - type: integer
        msg:
          title: Message
          type: string
        type:
          title: Error Type
          type: string

```

Here's my `main.swift` (this is for a Swift Package):

```
import OpenAPIRuntime
import OpenAPIURLSession

let client = Client(
    serverURL: try Servers.server1(),
    transport: URLSessionTransport()
)

let response = try await client.read_items_items__get(
    .init()
)


switch response {
case .ok(let okResponse):
    switch okResponse.body {
    case .json(let greeting):
        print(greeting)
    }
case .undocumented(statusCode: let statusCode, _):
[...]

PR #69, opened 2023-06-15
Bumps OpenAPIKit to version 3.0.0-alpha.7 fixing an issue with trailing slashes.

miguel-arrf:
### Motivation
As discussed in #51, there was an issue with paths containing trailing slashes. The problem was identified in the [OpenAPIKit](https://github.com/mattpolzin/OpenAPIKit) dependency. It was fixed in the PR https://github.com/mattpolzin/OpenAPIKit/pull/274.

### Modifications
This PR updates the OpenAPIKit dependency version from `3.0.0-alpha.6` to `3.0.0-alpha.7` ([available here](https://github.com/mattpolzin/OpenAPIKit/releases/tag/3.0.0-alpha.7)), which includes the mentioned PR.

### Result
If a path ends with a trailing slash, such as `foo/bar/`, the generated functions will now correctly call the path `foo/bar/` instead of (as previously) `foo/bar`.

### Test Plan
Replaced an endpoint in the reference test to use a path with a trailing slash.

### Resolves
- Resolves https://github.com/apple/swift-openapi-generator/issues/51

### DIFF
--- a/Tests/PetstoreConsumerTests/Test_Client.swift
+++ b/Tests/PetstoreConsumerTests/Test_Client.swift
@@ -344,7 +344,7 @@ final class Test_Client: XCTestCase {
     func testProbe_204() async throws {
         transport = .init { request, baseURL, operationID in
             XCTAssertEqual(operationID, "probe")
-            XCTAssertEqual(request.path, "/probe")
+            XCTAssertEqual(request.path, "/probe/")
             XCTAssertNil(request.query)
             XCTAssertEqual(baseURL.absoluteString, "/api")
             XCTAssertEqual(request.method, .post)
--- a/Package.swift
+++ b/Package.swift
@@ -47,7 +47,7 @@ let package = Package(
         // Read OpenAPI documents
         .package(
             url: "https://github.com/mattpolzin/OpenAPIKit.git",
-            exact: "3.0.0-alpha.6"
+            exact: "3.0.0-alpha.7"
         ),
         .package(
             url: "https://github.com/jpsim/Yams.git",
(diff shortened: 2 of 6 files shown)


simonjbeaumont:

czechboy0:
You could just change one of the paths in the integration test to end with a slash instead of adding another path+operation. But it'd be good for our tests to include this case.

simonjbeaumont (changes requested):
Thanks for the fix. Based on the discussion here and on the issue, I'll mark this as request changes, until we get the test changed.

czechboy0 (changes requested):
Thanks @miguel-arrf for updating the tests, but I don't think we need a new operation for it, as those are pretty expensive to maintain (as you just saw when adding a new one), so we add new ones sparingly, only when we can't extend existing ones. An easier alternative is to modify the path of one of the existing operations, and update the corresponding test.

miguel-arrf:
@czechboy0 Ah, I see! Thank you, I'll update it accordingly :)!

miguel-arrf:
(I still have some more tests to change, because of the Server!)

Tests/PetstoreConsumerTests/Test_Client.swift
```diff
@@ -344,7 +344,7 @@ final class Test_Client: XCTestCase {
     func testProbe_204() async throws {
         transport = .init { request, baseURL, operationID in
             XCTAssertEqual(operationID, "probe")
-            XCTAssertEqual(request.path, "/probe")
+            XCTAssertEqual(request.path, "/probe/")
```
czechboy0, line 347: Great, thanks for updating the `PetstoreConsumerTests` target as well – this is the important bit that verifies that the trailing slash is actually passed to the transport, and not stripped between OpenAPIKit and the generated code.

czechboy0 (approved):
Looks great, thank you, @miguel-arrf! 🙏

simonjbeaumont:
@miguel-arrf Looks great!

Would you like to update the PR description given you've gone to the effort of sorting the testing out?

> ⚠️ No steps have been taken yet to test this PR. The existing tests should be updated to include paths with trailing slashes.

simonjbeaumont (approved):

miguel-arrf:
Thanks @simonjbeaumont ! ☺️

Issue #40, opened 2023-05-25
Use the same .swift-format in the repo and in the generated code


czechboy0:
We use swift-format both for formatting the generator code, and the generated code. However, the configuration is duplicated in both places, we should have one source of truth for it instead.

denil-ct:
I've created a [PR](https://github.com/apple/swift-openapi-generator/pull/73) for this. Please have a look.

czechboy0:
Well, this is good timing: https://github.com/apple/swift-package-manager/pull/6067

Unfortunately, it's only supported in Swift 5.9, so I propose we punt this issue until later, and adopt it when we bump the minimum tools version to 5.9.

denil-ct:
Sounds good to me. So sometime in the fall when Xcode 15 comes out of beta I assume?

czechboy0:
To be determined, but marking as blocked for now. You can close your other PR for now, but we'll definitely appreciate the contribution when Swift OpenAPI Generator requires Swift tools version 5.9!

czechboy0:
Filed #75 to properly relate the two.

denil-ct:
@czechboy0 if #75 is 1.0, can't this be under the same milestone as well?

czechboy0:
@denil-ct the milestone doesn't mean when it'll land, it means which release it's blocking. It's likely this will land around the same time as #75, the milestone just means this wouldn't be considered a 1.0 blocker.

czechboy0:
We don't use swift-format as a library anymore, so this is obsolete.

PR #73, opened 2023-06-18
Add common configuration for swift-format

denil-ct:
### Motivation

Fixes this [issue](https://github.com/apple/swift-openapi-generator/issues/40)

### Modifications

- Removed the configuration defined in `SwiftFormat.swift`.
- Used the json configuration file in the project root to generate the configuration object.

### Result

A single configuration file for swift-format, which will be easier to handle.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Extensions/SwiftFormat.swift
+++ b/Sources/_OpenAPIGeneratorCore/Extensions/SwiftFormat.swift
@@ -11,27 +11,27 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 //===----------------------------------------------------------------------===//
+import Foundation
 import SwiftFormat
 import SwiftFormatConfiguration
 
 extension String {
+    private static let configurationURL = {
+        let configurationFilePath = #file
+            .components(separatedBy: "/")
+            .prefix(while: { $0 != "Sources" })
+            .joined(separator: "/")
+            .appending("/.swift-format")
+        guard #available(macOS 13.0, *) else {
+            return URL(fileURLWithPath: configurationFilePath)
+        }
+        return URL(filePath: configurationFilePath)
+    }()
     /// A copy of the string formatted using swift-format.
     var swiftFormatted: Self {
         get throws {
             var formattedString = ""
-            // TODO: Should be loaded from a swift-format file that we also use to format our own code.
-            var configuration = Configuration()
-            configuration.rules["OrderedImports"] = false
-            configuration.rules["NoAccessLevelOnExtensionDeclaration"] = false
-            configuration.rules["UseLetInEveryBoundCaseVariable"] = false
-            configuration.indentation = .spaces(4)
-            configuration.respectsExistingLineBreaks = false
-            configuration.lineBreakBeforeEachArgument = true
-            configuration.lineBreakBeforeControlFlowKeywords = false
-            configuration.lineBreakBeforeEachGenericRequirement = true
-            configuration.lineBreakAroundMultilineExpressionChainComponents = true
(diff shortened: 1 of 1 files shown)


Sources/_OpenAPIGeneratorCore/Extensions/SwiftFormat.swift
```diff
@@ -11,27 +11,27 @@
...
 import SwiftFormat
 import SwiftFormatConfiguration
 
 extension String {
+    private static let configurationURL = {
+        let configurationFilePath = #file
+            .components(separatedBy: "/")
+            .prefix(while: { $0 != "Sources" })
+            .joined(separator: "/")
+            .appending("/.swift-format")
```
denil-ct, line 24: I feel like this is quite susceptible to be triggered by other folders that have a `Sources` in their name. I started with it because it seemed the simplest. What is your opinion, am I being overly cautions here?

Maybe something like this might work better as it narrows down the search criteria?
```suggestion
        let configurationFilePath = #file
            .components(separatedBy: "swift-openapi-generator/Sources")[0]
            .appending("swift-openapi-generator/.swift-format")
```

Or is there some other way of getting the file location, that I am missing?

czechboy0:
Hi @denil-ct, thanks for picking this one up!

While the approach of finding the format file on disk would work when the source package is cloned in some local directory, but we also have to consider the use case where the CLI is built as a binary, transferred to another machine, and executed there. In that scenario, we still need to support formatting as part of the code generation pipeline.

What might be worth investigating is how we could capture the format file at _compile_ time and embed it in the binary, so that it's then always available, even when not distributed as a source package.

There's some precedent in embedding an Info.plist for a CLI into the binary, so maybe that's a place to start looking. Possibly having an internal build tool plugin that reads the file and generates a variable that contains the string representation is another. I'll let you investigate these and propose which one you think will work best.

czechboy0 (changes requested):
Marking as needs work as this doesn't cover the use case where the CLI is invoked outside of the generator source package.

simonjbeaumont:
Thanks @denil-ct for jumping in again!

> What might be worth investigating is how we could capture the format file at _compile_ time and embed it in the binary, so that it's then always available, even when not distributed as a source package.
> 
> There's some precedent in embedding an Info.plist for a CLI into the binary, so maybe that's a place to start looking. Possibly having an internal build tool plugin that reads the file and generates a variable that contains the string representation is another. I'll let you investigate these and propose which one you think works best.

Can't we use Swift PM for this. IIRC you can provide a `resources:` parameter in the target definition? Something like this:

```swift
...
    targets: [
...
        .executableTarget(
            dependencies: [ ... ],
            resources: [
                .process(".swift-format"),
            ],
        ),
...
    ]
```

czechboy0:
@simonjbeaumont IIRC, that'll create a resource bundle next to the binary, and if you transfer the binary to another machine without explicitly also transferring the resource bundle, it won't find it at runtime. But might require a confirmation in case I misremembered how it works.

So it'd work for a framework or an app target (it'd get embedded in that bundle), but not for an executable binary CLI that can be distributed on its own, as it doesn't get distributed as a bundle.

czechboy0:
While trying @simonjbeaumont's suggestion, I stumbled upon a new SwiftPM 5.9 feature that allows embedding files in code: https://github.com/apple/swift-openapi-generator/issues/40#issuecomment-1596942582

denil-ct:
Shall I close this PR, then? @czechboy0

denil-ct:
Closing this PR for now, as we are blocked on [this](https://github.com/apple/swift-openapi-generator/issues/40#issuecomment-1596942582).

PR #74, opened 2023-06-19
Create an Unresolved* typealias for all supported OpenAPI types

### LINKED ISSUE #12: Create an Unresolved* typealias for all supported OpenAPI types
For easier handling of potentially referenceable OpenAPI types in the generator, e.g. `Either<JSONReference<JSONSchema>, JSONSchema>?`, we should create a typealias called `Unresolved*` (e.g. `UnresolvedSchema`). This is already present for some of the types, but not all.

The full list right now is:
- schemas
- parameters
- (response) headers
- request bodies
- responses

andrewse02:
### Motivation

Currently, when referencing OpenAPI types that may not be referenceable, it requires a very long and unintuitive type name (e.g., `Either<JSONReference<OpenAPI.Request>, OpenAPI.Request>`). We can clean this up, by exposing more usable,  and practical typealiases to improve quality of life and speed while working on the generator.

### Modifications

**Added and used `Unresolved*` typaliases for:**
- OpenAPI.Request
- OpenAPI.Response
- OpenAPI.Header
- JSONSchema

### Result

New typealiases will be added, decreasing confusion, and increasing symmetry when using OpenAPI types.

### Resolves

Resolves #12 

### Edits

1. Removed additions of `Resolved*` typealiases

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/Content/SchemaContent.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Content/SchemaContent.swift
@@ -24,7 +24,7 @@ struct SchemaContent {
     ///
     /// Can be nil for unstructured JSON payloads, or for unstructured
     /// content types such as binary data.
-    var schema: Either<JSONReference<JSONSchema>, JSONSchema>?
+    var schema: UnresolvedSchema?
 }
 
 /// A type grouping schema content and its computed Swift type usage.
@@ -44,3 +44,8 @@ struct TypedSchemaContent {
         typeUsage ?? TypeName.valueContainer.asUsage
     }
 }
+
+/// An unresolved OpenAPI schema.
+///
+/// Can be either a reference or an inline schema.
+typealias UnresolvedSchema = Either<JSONReference<JSONSchema>, JSONSchema>
--- a/Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
@@ -20,7 +20,7 @@ struct TypedParameter {
     var parameter: ResolvedParameter
 
     /// The underlying schema.
-    var schema: Either<JSONReference<JSONSchema>, JSONSchema>
+    var schema: UnresolvedSchema
 
     /// The computed type usage.
     var typeUsage: TypeUsage
@@ -66,7 +66,7 @@ extension TypedParameter {
     }
 }
 
-extension Either where A == JSONReference<JSONSchema>, B == JSONSchema {
+extension UnresolvedSchema {
 
     /// A schema to be inlined.
     ///
@@ -128,7 +128,7 @@ extension FileTranslator {
         let locationTypeName = parameter.location.typeName(in: parent)
         let foundIn = "\(locationTypeName.description)/\(parameter.name)"
 
-        let schema: Either<JSONReference<JSONSchema>, JSONSchema>
+        let schema: UnresolvedSchema
         let codingStrategy: CodingStrategy
         switch parameter.schemaOrContent {
         case let .a(schemaContext):
--- a/Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
@@ -17,7 +17,7 @@ import OpenAPIKit30
 struct TypedParameter {
 
     /// The OpenAPI parameter.
-    var parameter: ResolvedParameter
+    var parameter: OpenAPI.Parameter
 
     /// The underlying schema.
     var schema: UnresolvedSchema
@@ -54,7 +54,7 @@ extension TypedParameter {
     }
 
     /// The location of the parameter in the HTTP request.
-    var location: ResolvedParameter.Context.Location {
+    var location: OpenAPI.Parameter.Context.Location {
         parameter.location
     }
 
@@ -117,7 +117,7 @@ extension FileTranslator {
     ) throws -> TypedParameter? {
 
         // Collect the parameter
-        let parameter: ResolvedParameter
+        let parameter: OpenAPI.Parameter
         switch unresolvedParameter {
         case let .a(ref):
             parameter = try components.lookup(ref)
@@ -228,10 +228,7 @@ extension FileTranslator {
 /// Can be either a reference or an inline parameter.
 typealias UnresolvedParameter = Either<JSONReference<OpenAPI.Parameter>, OpenAPI.Parameter>
 
-/// A resolved OpenAPI parameter.
-typealias ResolvedParameter = OpenAPI.Parameter
-
-extension ResolvedParameter.Context.Location {
+extension OpenAPI.Parameter.Context.Location {
 
     /// A name of the location usable as a Swift type name.
     var shortTypeName: String {
--- a/Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
@@ -17,7 +17,7 @@ import OpenAPIKit30
 struct TypedParameter {
 
     /// The OpenAPI parameter.
-    var parameter: OpenAPI.Parameter
+    var parameter: ResolvedParameter
 
     /// The underlying schema.
     var schema: Either<JSONReference<JSONSchema>, JSONSchema>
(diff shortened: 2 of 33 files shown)


Sources/_OpenAPIGeneratorCore/Translator/Content/SchemaContent.swift
```diff
@@ -44,3 +44,8 @@ struct TypedSchemaContent {
         typeUsage ?? TypeName.valueContainer.asUsage
     }
 }
+
+/// An unresolved OpenAPI schema.
+///
+/// Can be either a reference or an inline schema.
+typealias UnresolvedSchema = Either<JSONReference<JSONSchema>, JSONSchema>
```
andrewse02, line 51: Not sure about the location of this typealias. Since this type isn't the same as `OpenAPI.*`, there didn't seem to be a very obvious place to put it.

Sources/_OpenAPIGeneratorCore/Translator/RequestBody/TypedRequestBody.swift
```diff
@@ -113,3 +113,11 @@ extension FileTranslator {
...
     }
 }
+
+/// An unresolved OpenAPI request.
+///
+/// Can be either a reference or an inline request.
+typealias UnresolvedRequest = Either<JSONReference<OpenAPI.Request>, OpenAPI.Request>
+
+/// A resolved OpenAPI request.
+typealias ResolvedRequest = OpenAPI.Request
```
andrewse02, line 123: The `Resolved*` typealiases weren't in the [issue's](#12) scope, but there was an existing `ResolvedParameter` typealias, so I made these to keep it symmetric. I honestly would be totally fine if we removed all of these. To me, they don't seem very useful since `OpenAPI.Request`, etc., is pretty reasonable and intuitive.

Sources/_OpenAPIGeneratorCore/Translator/Content/SchemaContent.swift
andrewse02, line 51: I didn't create a `ResolvedSchema` typealias, which breaks the symmetry. It felt very unnecessary to typealias `JSONSchema`.

Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeAssigner.swift
```diff
@@ -381,13 +381,13 @@ struct TypeAssigner {
...
+    /// - `#/components/responses` -> `ResolvedResponse`
     /// - `#/components/callbacks` -> `OpenAPI.Callbacks`
-    /// - `#/components/parameters` -> `OpenAPI.Parameter`
+    /// - `#/components/parameters` -> `ResolvedParameter`
     /// (includes request headers)
     /// - `#/components/examples` -> `OpenAPI.Example`
-    /// - `#/components/requestBodies` -> `OpenAPI.Request`
-    /// - `#/components/headers` -> `OpenAPI.Header` (response headers)
+    /// - `#/components/requestBodies` -> `ResolvedRequest`
+    /// - `#/components/headers` -> `ResolvedHeader` (response headers)
```
andrewse02, line 390: Wasn't sure about including the documentation comments here, would love to hear some thoughts on this. (Assuming we keep the `Resolved*` typealiases)

Sources/_OpenAPIGeneratorCore/Translator/Content/SchemaContent.swift
czechboy0: I think the location is fine.

Sources/_OpenAPIGeneratorCore/Translator/RequestBody/TypedRequestBody.swift
czechboy0: I think I agree - can you get rid of all the Resolved* typealiases? They seem like an unnecessary layer of indirection.

czechboy0 (changes requested):
Making all the Unresolved* aliases is great, but as you suggested, let's get rid of the Resolved* aliases, as they create an extra layer of indirection to clear types, like JSONSchema/OpenAPI.Parameter, and so on. Thanks 🙏

czechboy0 (approved):
Looks great, thanks!

Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
```diff
@@ -228,9 +228,6 @@ extension FileTranslator {
 /// Can be either a reference or an inline parameter.
 typealias UnresolvedParameter = Either<JSONReference<OpenAPI.Parameter>, OpenAPI.Parameter>
 
-/// A resolved OpenAPI parameter.
-typealias ResolvedParameter = OpenAPI.Parameter
```
simonjbeaumont, line 232: Thanks for clearing up the asymmetry here. 👍

Sources/_OpenAPIGeneratorCore/Translator/Responses/TypedResponseHeader.swift
```diff
@@ -164,3 +164,8 @@ extension FileTranslator {
         )
     }
 }
+
+/// An unresolved OpenAPI response header.
+///
+/// Can be either a reference or an inline response header.
+typealias UnresolvedHeader = Either<JSONReference<OpenAPI.Header>, OpenAPI.Header>
```
simonjbeaumont, line 171: Nit: The OpenAPI spec has the headers at a scope not specific to requests, e.g. when used in `#components/headers`.

```suggestion
/// An unresolved OpenAPI header.
///
/// Can be either a reference or an inline header.
typealias UnresolvedHeader = Either<JSONReference<OpenAPI.Header>, OpenAPI.Header>
```

simonjbeaumont (changes requested):

simonjbeaumont:
@andrewse02 thanks for taking the time to clean this up. Just one minor nit from me—otherwise LGTM!

Sources/_OpenAPIGeneratorCore/Translator/Responses/TypedResponseHeader.swift
czechboy0: Hmm I think @andrewse02 had it correctly, `Header Object` ([OpenAPI specification link](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#header-object)) _must not_ have `in` and `name` fields, which makes it invalid to provide in a place of a parameter. So my understanding has always been that they only apply to response headers, and request headers are just parameters. Are you reading the specification differently?

simonjbeaumont: Yep. I checked the spec before commenting. I was considering parity with how it’s spelled out in the spec and with the type name chosen here. 

Seemed reasonable to keep them aligned.

czechboy0: I see – we kind of have two options, because we use the explicit name "response header" all over the generator codebase, to explicitly disambiguate it from request headers, which are just parameters. So the question is whether we want to stick with the spec spelling, or with the spelling we use in the generator business logic. Either is fine, I guess, but we should probably rename things holistically across the codebase to avoid drift.

simonjbeaumont: Ah, do we. I hadn't remembered that and didn't notice it from the context of the diff. Let's just leave this as it is then.

simonjbeaumont (approved):

Sources/_OpenAPIGeneratorCore/Translator/Responses/TypedResponseHeader.swift
simonjbeaumont: Oh @czechboy0—were you potentially suggesting we actually rename the type alias here to match the rest of the generator? i.e. this:
```suggestion
/// An unresolved OpenAPI response header.
///
/// Can be either a reference or an inline response header.
typealias UnresolvedResponseHeader = Either<JSONReference<OpenAPI.Header>, OpenAPI.Header>
```

czechboy0: Might be worth to go all the way and disambiguate everywhere, but I'd also be happy with that being a separate issue, to avoid further expanding this PR. It'll likely be more invasive than the scope of this PR.

czechboy0: Broke it out: https://github.com/apple/swift-openapi-generator/issues/78

PR #81, opened 2023-06-20
Emit YAML processing errors that Xcode can display

### LINKED ISSUE #65: Emit errors processing .yaml files that Xcode can display
I got the following error in the build transcript:

```
Showing All Messages
Error: dataCorrupted(Swift.DecodingError.Context(codingPath: [], debugDescription: "The given data was not valid YAML.", underlyingError: Optional(19:3: error: scanner: while scanning for the next token in line 19, column 3
found character that cannot start any token:
  	post:
  ^)))
```

It would be great if this were something Xcode could parse and show as an error in the error navigator.

macblazer:
### Motivation

If there are any errors in the openapi.yaml file currently the errors show up in the build transcript but are not shown in the Xcode issue sidebar and are also not linked to the openapi.yaml file in the sidebar. This makes it hard to know what has failed, and what to do to fix the file.

### Modifications

The YAML parser now catches decoding errors, and if they are common parsing or scanning errors it will emit a diagnostic message then throw a failure exit code.  This required importing the swift-argument-parser module into the main _OpenAPIGeneratorCore.

Added `absoluteFilePath` and `lineNumber` optional properties to the `Diagnostics` struct to capture more info. Modified the `Diagnostics.description` method to build a properly formatted message that Xcode can parse to point out the file and line causing issues.

### Result

In consuming projects, invalid yaml in the openapi.yaml file is now highlighted in the Xcode Issue Navigator after a build with this plugin.

### Test Plan

Added unit tests to verify emitted diagnostic and thrown error.  Also tested manually with an example project.

### Resolves

- Resolves #65.


macblazer:
I'm not particularly happy with adding `ArgumentParser` as a dependency into the `_OpenAPIGeneratorCore`.

Without it, any thrown error is caught by `ArgumentParser` and the thrown error's localizedDescription is used with a prefix of "Error: " which completely blows away any other diagnostic that I was able to emit.

The code now throws an `ArgumentParser.ExitCode.failure` after emitting a diagnostic with the proper information for Xcode to pick up the filepath and line number.  Using `ArgumentParser.ExitCode` was the only way I found to have the correct diagnostic message.

Package.swift
```diff
@@ -116,8 +117,7 @@ let package = Package(
         .executableTarget(
             name: "swift-openapi-generator",
             dependencies: [
-                "_OpenAPIGeneratorCore",
-                .product(name: "ArgumentParser", package: "swift-argument-parser"),
+                "_OpenAPIGeneratorCore"
```
czechboy0, line 120: It'd be great to avoid making the library depend on swift-argument-parser, see comments below how we should be able to remove it again.

Sources/_OpenAPIGeneratorCore/Diagnostics.swift
```diff
@@ -37,6 +37,12 @@ public struct Diagnostic: Error, Codable {
     /// A user-friendly description of the diagnostic.
     public var message: String
 
+    /// The absolute path to a specific source file that triggered the diagnostic.
+    public var absoluteFilePath: URL?
+
+    /// The line number within the specific source file that triggered the diagnostic.
+    public var lineNumber: Int?
```
czechboy0, line 44: Since we'll either always have location, or not have location, let's represent it as 

```swift 
struct Diagnostic {
  //...
  struct Location {
    var filePath: String
    var lineNumber: Int
  }
  var location: Location?
}
```

Sources/_OpenAPIGeneratorCore/Diagnostics.swift
```diff
@@ -101,8 +107,20 @@ extension Diagnostic.Severity: CustomStringConvertible {
 
 extension Diagnostic: CustomStringConvertible {
     public var description: String {
+        var prefix = ""
+        if let filePath = absoluteFilePath {
+            if #available(macOS 13.0, *) {
+                prefix = "\(filePath.path(percentEncoded: false)):"
+            } else {
+                prefix = "\(filePath.path):"
```
czechboy0, line 115: Just use this codepath that works on all platforms, no need to branch out to the new API on newer OS versions.

Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
```diff
@@ -58,9 +65,63 @@ struct YamsParser: ParserProtocol {
...
+                from: input.contents
+            )
+        } catch DecodingError.dataCorrupted(let errorContext) {
+            try possibly_throw_parsing_error(context: errorContext, input: input, diagnostics: diagnostics)
+            throw DecodingError.dataCorrupted(errorContext)
+        }
+    }
+
+    /// Detect specific YAML parsing errors to emit nicely formatted errors for IDEs
+    private func possibly_throw_parsing_error(
```
czechboy0, line 80:
```suggestion
    private func checkParsingError(
```

A naming suggestion - let's use camelCase, and suggested slightly different phrase.

Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
```diff
@@ -58,9 +65,63 @@ struct YamsParser: ParserProtocol {
...
+                try throw_parsing_error(
+                    context: yamlContext,
+                    problem: yamlProblem,
+                    lineNumber: yamlMark.line - 1,
+                    input: input,
+                    diagnostics
+                )
+            }
+        } else if let openAPIError = context.underlyingError as? OpenAPIError {
+            var problem = Diagnostic.error(message: openAPIError.localizedDescription)
```
czechboy0, line 104: Can you add `location: Location? = nil` to the convenience initializer so that we can create the Diagnostic in one step here, instead of modifying it after creation?

Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
```diff
@@ -58,9 +65,63 @@ struct YamsParser: ParserProtocol {
...
+            problem.absoluteFilePath = input.absolutePath
+            diagnostics.emit(problem)
+            throw ExitCode.failure
```
czechboy0, line 107: The Diagnostic itself conforms to error, so you can emit it into the DiagnosticCollector, and then also throw it. This will allow us to drop the dependency on swift-argument-parser.

Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
```diff
@@ -58,9 +65,63 @@ struct YamsParser: ParserProtocol {
...
+        }
+    }
+
+    private func throw_parsing_error(
```
czechboy0, line 111:
```suggestion
    private func throwParsingError(
```

CamelCase, and please add a documentation comment for this method.

Kyle-Ye, line 111: Nit
```suggestion
    private func throwParsingError(
```

Tests/OpenAPIGeneratorCoreTests/Parser/Test_YamsParser.swift
```diff
@@ -25,24 +26,151 @@ final class Test_YamsParser: Test_Core {
...
+        let yaml = """
+            openapi: "3.0.0"
+            info:
+             title: "Test"
+              version: 1.0.0
+            paths: {}
+            """
+
+        XCTAssertThrowsError(try _test(yaml, diagnostics: collector)) { error in
+            if let exitError = error as? ExitCode {
```
czechboy0, line 54: Here, you'd check for the Diagnostic being thrown, after you remove the swift-argument-parser dependency.

czechboy0 (changes requested):
This is great, thanks for working on it, @macblazer! I think we can do this without the swift-argument-parser dependency, so I added a few comments.

Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
macblazer: When throwing the Diagnostic after emitting it, Xcode shows duplicate errors because ArgumentParser adds the thrown error (as a string) with `Error: ` in front of the description like so:

> /Users/<redacted>/GreetingServiceClient/Sources/openapi.yaml: error: Expected to find `info` key in the root Document object but it is missing.
> Error: /Users/<redacted>/GreetingServiceClient/Sources/openapi.yaml: error: Expected to find `info` key in the root Document object but it is missing.

Duplicate errors are maybe acceptable, but it's a tradeoff with adding the ArgumentParser dependency.

If the call to `emit()` is removed, then Xcode only gets the message with `Error: ` in front of it which does not link to the file and line number.  If the throw is removed then it crashes because of course it is an error and something should be thrown.

I don't see any way short of using `ExitCode.failure` to have ArgumentParser _not_ add `Error: ` to the front of a thrown instance's localizedDescription.  Maybe this should be a feature request on the ArgumentParser repo.

czechboy0: I think it's probably fine to have the duplicate error in the log then, and let's file a separate issue to investigate if we can remove the duplication. It's more important that we don't introduce this dependency.

Tests/OpenAPIGeneratorCoreTests/Parser/Test_YamsParser.swift
macblazer: Done.

Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
macblazer: This method was completely removed in favor of throwing the Diagnostic directly.

Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
macblazer: Done

Sources/_OpenAPIGeneratorCore/Diagnostics.swift
macblazer: Done

Sources/_OpenAPIGeneratorCore/Diagnostics.swift
macblazer: The lineNumber is still optional for some error cases, so I left that as optional inside the `Location` struct.

Package.swift
macblazer: Done.

Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
macblazer: Done

Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
macblazer: Done

Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
macblazer: I found an alternate way to handle this de-duplication in the GenerateOptions+runGenerator.swift file (which is part of the executable target that already imports ArgumentParser) to catch **any** thrown `Diagnostic`, emit the nice error message from the diagnostic, then throw `ExitCode.failure` so ArgumentParser can terminate cleanly without any other messages.

The parser no longer calls `emit()` but simply throws the `Diagnostic`.

Sources/swift-openapi-generator/GenerateOptions+runGenerator.swift
```diff
@@ -65,13 +66,19 @@ extension _GenerateOptions {
...
+                configs: configs,
+                isPluginInvocation: isPluginInvocation,
+                outputDirectory: outputDirectory,
+                diagnostics: diagnostics
+            )
+        } catch let error as Diagnostic {
+            // Emit our nice Diagnostics message instead of relying on ArgumentParser output.
+            diagnostics.emit(error)
+            throw ExitCode.failure
+        }
```
czechboy0, line 81: Oh cool, does this avoid the duplication?

Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
```diff
@@ -30,37 +30,84 @@ struct YamsParser: ParserProtocol {
...
+                message: openAPIError.localizedDescription,
+                location: .init(filePath: input.absolutePath.path)
+            )
         }
+    }
+}
+
+extension Diagnostic {
+    /// Use when the document is an unsupported version.
+    static func openAPIVersionError(versionString: String, location: Location) -> Diagnostic {
```
czechboy0, line 97: Please also document parameters

Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
```diff
@@ -30,37 +30,84 @@ struct YamsParser: ParserProtocol {
...
+                "Unsupported document version: \(versionString). Please provide a document with OpenAPI versions in the 3.0.x set.",
+            location: location
+        )
+    }
 
-        return try decoder.decode(
-            OpenAPI.Document.self,
-            from: input.contents
+    // Use when the YAML document is completely missing the `openapi` version key.
+    static func openAPIMissingVersionError(location: Location) -> Diagnostic {
```
czechboy0, line 106: Please also document parameters

Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
```diff
@@ -30,37 +30,84 @@ struct YamsParser: ParserProtocol {
...
+                from: input.contents
+            )
+        } catch DecodingError.dataCorrupted(let errorContext) {
+            try checkParsingError(context: errorContext, input: input, diagnostics: diagnostics)
+            throw DecodingError.dataCorrupted(errorContext)
+        }
+    }
+
+    /// Detect specific YAML parsing errors to throw nicely formatted diagnostics for IDEs
+    private func checkParsingError(
```
czechboy0, line 69: Please also document parameters

Sources/_OpenAPIGeneratorCore/Diagnostics.swift
```diff
@@ -85,11 +94,16 @@ public struct Diagnostic: Error, Codable {
     public static func unsupported(
```
czechboy0, line 94: Please add documentation for the location parameter

Sources/_OpenAPIGeneratorCore/Diagnostics.swift
```diff
@@ -67,9 +75,10 @@ public struct Diagnostic: Error, Codable {
     /// - Returns: An error diagnostic.
```
czechboy0, line 75: Please add documentation for the location parameter

Sources/_OpenAPIGeneratorCore/Diagnostics.swift
```diff
@@ -54,9 +61,10 @@ public struct Diagnostic: Error, Codable {
     /// - Returns: A warning diagnostic.
     public static func warning(
```
czechboy0, line 62: Please add documentation for the location parameter

Sources/_OpenAPIGeneratorCore/Diagnostics.swift
```diff
@@ -37,12 +37,19 @@ public struct Diagnostic: Error, Codable {
     /// A user-friendly description of the diagnostic.
     public var message: String
 
+    /// The absolute path to a specific source file and optional line number within that file that triggered the diagnostic.
+    public struct Location: Codable {
+        public var filePath: String
+        public var lineNumber: Int?
+    }
+    public var location: Location?
```
czechboy0, line 45: Please add documentation for the properties inside Location, and the location property itself.

czechboy0 (changes requested):
Looks great! Just added a few requests for missing documentation comments, otherwise it's almost ready to go.

Sources/swift-openapi-generator/GenerateOptions+runGenerator.swift
macblazer: Yes, this fixes the duplication.  Any time the code throws a `Diagnostic` it will be emitted here, and ArgumentParser will exit with a failure code but no other output which is exactly what we want.

Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
macblazer: Done

Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
macblazer: Done

Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
macblazer: Done

Sources/_OpenAPIGeneratorCore/Diagnostics.swift
macblazer: Done

Sources/_OpenAPIGeneratorCore/Diagnostics.swift
macblazer: Done

Sources/_OpenAPIGeneratorCore/Diagnostics.swift
macblazer: Done

Sources/_OpenAPIGeneratorCore/Diagnostics.swift
macblazer: Done

czechboy0 (approved):
Looks great, thanks!

czechboy0:

simonjbeaumont (approved):
Thanks @macblazer. This is a great step forward, here ❤️

PR #87, opened 2023-06-21
Emit a descriptive error when running the generator on iOS, tvOS, or watchOS

### LINKED ISSUE #86: Emit a descriptive error when running the generator on iOS, tvOS, or watchOS
The `_OpenAPIGeneratorCore` library and `swift-openapi-generator` executable are mostly an implementation detail of the `OpenAPIGenerator` plugin, all vended by the `swift-openapi-generator` package.

The generator runs at _build_ time to generate code, but the `_OpenAPIGeneratorCore` library shouldn't actually be linked by adopters's target, which is documented correctly here (Step 7 of https://swiftpackageindex.com/apple/swift-openapi-generator/0.1.3/tutorials/swift-openapi-generator/clientxcode#Configuring-your-target-to-use-the-Swift-OpenAPI-Generator-plugin), but it's also easy to get it wrong and link it when adding the package in Xcode.

We should make it so that a clear, descriptive error is emitted when someone links the generator module itself to a non-macOS/Linux target.

czechboy0:
### Motivation

Resolves #86, which some adopters are hitting by accident.

### Modifications

Makes the generator itself build successfully on iOS, tvOS, watchOS, but not really, as I just added an explicit, clear error on those platforms pointing out that the adopter probably did this by accident, and how to fix it.

### Result

A clear, descriptive error is now emitted when the adopter links the `_OpenAPIGeneratorCore` target from e.g. an iOS app (most likely by accident):

```
.../swift-openapi-generator/Sources/_OpenAPIGeneratorCore/PlatformChecks.swift:4:8 Running the generator tool itself is not supported on iOS, tvOS, and watchOS. Check that your app is not linking the generator directly. For details, check out: https://github.com/apple/swift-openapi-generator/issues/86
```

### Test Plan

Tested locally by creating an iOS app with a lower deployment target and linked the generator directly. The error goes away when the linkages is removed, as expected.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -19,11 +19,9 @@ let package = Package(
     platforms: [
         .macOS(.v10_15),
 
-        // The platforms below are not currently supported for running the
-        // generator itself (only the generated code), they are here
-        // to allow the generator to emit a more descriptive error,
-        // see the file PlatformChecks.swift for details.
-        .iOS(.v13), .tvOS(.v13), .watchOS(.v6),
+        // The platforms below are not currently supported for running the generator itself.
+        // We include them here to allow the us to emit a more descriptive compiler error.
+        .iOS("99"), .tvOS("99"), .watchOS("99"),
     ],
     products: [
         .executable(name: "swift-openapi-generator", targets: ["swift-openapi-generator"]),
--- a/Package.swift
+++ b/Package.swift
@@ -17,7 +17,13 @@ import PackageDescription
 let package = Package(
     name: "swift-openapi-generator",
     platforms: [
-        .macOS(.v10_15)
+        .macOS(.v10_15),
+
+        // The platforms below are not currently supported for running the
+        // generator itself (only the generated code), they are here
+        // to allow the generator to emit a more descriptive error,
+        // see the file PlatformChecks.swift for details.
+        .iOS(.v13), .tvOS(.v13), .watchOS(.v6),
     ],
     products: [
         .executable(name: "swift-openapi-generator", targets: ["swift-openapi-generator"]),
--- a/Package.swift
+++ b/Package.swift
@@ -19,8 +19,9 @@ let package = Package(
     platforms: [
         .macOS(.v10_15),
 
-        // The platforms below are not currently supported for running the generator itself.
-        // We include them here to allow the us to emit a more descriptive compiler error.
+        // The platforms below are not currently supported for running
+        // the generator itself. We include them here to allow the generator
+        // to emit a more descriptive compiler error.
         .iOS("99"), .tvOS("99"), .watchOS("99"),
     ],
     products: [
--- a/Sources/_OpenAPIGeneratorCore/PlatformChecks.swift
+++ b/Sources/_OpenAPIGeneratorCore/PlatformChecks.swift
@@ -1,3 +1,16 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftOpenAPIGenerator open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftOpenAPIGenerator project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
 
 // Details: https://github.com/apple/swift-openapi-generator/issues/86
 #if os(iOS) || os(tvOS) || os(watchOS)
--- a/Sources/_OpenAPIGeneratorCore/PlatformChecks.swift
+++ b/Sources/_OpenAPIGeneratorCore/PlatformChecks.swift
@@ -1,0 +1,5 @@
+
+// Details: https://github.com/apple/swift-openapi-generator/issues/86
+#if os(iOS) || os(tvOS) || os(watchOS)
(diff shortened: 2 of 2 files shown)


Sources/_OpenAPIGeneratorCore/PlatformChecks.swift
```diff
@@ -0,0 +1,5 @@
+
+// Details: https://github.com/apple/swift-openapi-generator/issues/86
+#if os(iOS) || os(tvOS) || os(watchOS)
+#error("Running the generator tool itself is not supported on iOS, tvOS, and watchOS. Check that your app is not linking the generator directly. For details, check out: https://github.com/apple/swift-openapi-generator/issues/86")
```
czechboy0, line 4: Feedback on the wording is welcome.

denil-ct: _The generator tool cannot run on iOS, tvOS, or watchOS. Make sure your app does not link against the generator directly. See https://github.com/apple/swift-openapi-generator/issues/86 for details._

I feel like this removes redundant words from the first sentence and on the second, it adds a more polite and imperative tone.

Sources/_OpenAPIGeneratorCore/PlatformChecks.swift
```diff
@@ -0,0 +1,18 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+// Details: https://github.com/apple/swift-openapi-generator/issues/86
+#if os(iOS) || os(tvOS) || os(watchOS)
```
simonjbeaumont, line 16: I think this is what we want. Especially with upcoming `visionOS`.

```suggestion
#if !os(macOS) || !os(Linux)
```

Sources/_OpenAPIGeneratorCore/PlatformChecks.swift
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+// Details: https://github.com/apple/swift-openapi-generator/issues/86
```
simonjbeaumont, line 15: I'd rather us carry informative comments _here_ than link out to Github.

```suggestion
// Emit a compiler error if this library is linked with a target in an adopter project.
```

Sources/_OpenAPIGeneratorCore/PlatformChecks.swift
```diff
@@ -0,0 +1,18 @@
...
+#if os(iOS) || os(tvOS) || os(watchOS)
+#error("Running the generator tool itself is not supported on iOS, tvOS, and watchOS. Check that your app is not linking the generator directly. For details, check out: https://github.com/apple/swift-openapi-generator/issues/86")
```
simonjbeaumont, line 17:
```suggestion
#error("_OpenAPIGeneratorCore is only to be used by swift-openapi-generator itself—your target should not link this library directly.")
```

Package.swift
```diff
@@ -17,7 +17,13 @@ import PackageDescription
...
     name: "swift-openapi-generator",
     platforms: [
-        .macOS(.v10_15)
+        .macOS(.v10_15),
+
+        // The platforms below are not currently supported for running the
+        // generator itself (only the generated code), they are here
+        // to allow the generator to emit a more descriptive error,
+        // see the file PlatformChecks.swift for details.
+        .iOS(.v13), .tvOS(.v13), .watchOS(.v6),
```
simonjbeaumont, line 26: I'm really not a fan of us having to do this and would rather us consider removing `_OpenAPIGeneratorCore` from `products`.

If we do have to do it, we can make use of string version numbers for unreasonably high version numbers, e.g.

```suggestion
        .macOS(.v10_15),

        // The platforms below are not currently supported for running the generator itself.
        // We include them here to allow the us to emit a more descriptive compiler error.
        .iOS("99"), .tvOS("99"), .watchOS("99"),
```

simonjbeaumont (changes requested):
If we removed `_OpenAPIGeneratorCore` from the `products` then I think this is less likely to trip folks up. When I tested that approach locally, it is no longer offered by Xcode when adding the package dependency.

Sources/_OpenAPIGeneratorCore/PlatformChecks.swift
czechboy0: They might have linked the CLI though, not the Core library, so it might be confusing why we're warning about the library.

Package.swift
czechboy0: I'll do the high version numbers, are 9999 instead also good? I've seen that used elsewhere.

simonjbeaumont: While they work in `@available` annotations, I tried them in this setting and it caused less readable errors because there are some other macro expansions which seem to rely on the major component being 2 digits. You can try and see for yourself what happens if you use 9999.

Sources/_OpenAPIGeneratorCore/PlatformChecks.swift
simonjbeaumont: OK, well we can append a little to the message?

```suggestion
#error("_OpenAPIGeneratorCore is only to be used by swift-openapi-generator itself—your target should not link this library or the plugin directly.")
```

PR #88, opened 2023-06-21
Add `await` and `Protocol` to the identifiers escaping rules

### LINKED ISSUE #34: Add "Protocol" to the list of reserved terms
A property in the OpenAPI doc can be called "Protocol", which breaks the Swift build.

Add it to the `String.keywords` set, should be a quick fix.

denil-ct:
### Motivation

Resolves https://github.com/apple/swift-openapi-generator/issues/34

### Modifications

- Added `Protocol` to the keywords set.
- Added `await` to the keywords set.

### Result

- Would be able to support `Protocol` and `await` as property names.

### Test Plan

Tested and verified for both names. They are properly renamed to `_Protocol` and `_await`

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Extensions/String.swift
+++ b/Sources/_OpenAPIGeneratorCore/Extensions/String.swift
@@ -217,5 +217,7 @@ fileprivate extension String {
         "Array",
         "Type",
         "type",
+        "Protocol",
+        "await",
     ]
 }


czechboy0 (changes requested):
Hi @denil-ct, thanks for another contribution!

Adding the two keywords is the quick fix and covers the original issues, so let's keep this PR to just that change, and please add to the unit tests to have some automation testing.

Could you break out the hex encoding experiment into a separate PR? I'd like us to discuss it more, but I don't want to block fixing #34 for it. Thanks! 🙏

denil-ct:
Sure, that makes sense. 
Quick question @czechboy0 , the initial issue mentions only the `Protocol` keyword. Should there be a new issue for `await` or can it be added on to that?

czechboy0:
You can add await as well in this PR, the change is well understood. 

The hex stuff is new and we'll want to think about the implications to existing code etc, thus the separate PR.

czechboy0:

czechboy0 (approved):
Looks good, thanks! I checked the unit tests and we already cover the "keyword" case, so no real need to list all the keywords there.

Sources/_OpenAPIGeneratorCore/Extensions/String.swift
```diff
@@ -217,5 +217,7 @@ fileprivate extension String {
         "Array",
         "Type",
         "type",
+        "Protocol",
```
simonjbeaumont, line 220: Do we need this one? This works fine:

```swift
let Protocol = 1
print(Protocol)  // prints 1
```

Sources/_OpenAPIGeneratorCore/Extensions/String.swift
```diff
@@ -217,5 +217,7 @@ fileprivate extension String {
...
+        "await",
```
simonjbeaumont, line 221: This one is fun... TIL this fails only at use, but not on the variable declaration:

```swift
let await = 1  // this is fine
print(await)  // this is the problem
```

simonjbeaumont (changes requested):
@denil-ct thank you for once again for diving in 💪 

I left a comment about `Protocol`—not sure if we need it.

Additionally, would you mind adopting a PR title that would make a good entry in the release notes. For this PR, something like:

```
Add `await` to the identifiers escaping rules 
```

Sources/_OpenAPIGeneratorCore/Extensions/String.swift
denil-ct: Unfortunately, we have a `Self` before it, which makes it a valid keyword in that context. If there is no self, then its not needed.

```swift
public init(
                Protocol: Swift.String? = nil,
                message: OpenAPIRuntime.OpenAPIValueContainer
            ) {
                self.Protocol = Protocol
                self.message = message
}
```

Sources/_OpenAPIGeneratorCore/Extensions/String.swift
denil-ct: So, this one is even more quirky. In the generated type init since await would be the RHS without any expression after it, I guess swift-format tries to apply a rule to move the next line over.

What we think will happen
```swift
public init(await: Swift.String? = nil, message: OpenAPIRuntime.OpenAPIValueContainer)
{
                self.await = await
                self.message = message
}
```

But what actually happens is
```swift
public init(await: Swift.String? = nil, message: OpenAPIRuntime.OpenAPIValueContainer) {
                self.await = await self.message = message
            }
```

Sources/_OpenAPIGeneratorCore/Extensions/String.swift
czechboy0: Yeah, `Protocol` came from a real OpenAPI doc, and it broke the build, so I think we need to add it as a reserved keyword, just like `Self` and `Type`.

simonjbeaumont: All makes perfect sense, thanks for clarifying.

simonjbeaumont (approved):

Issue #21, opened 2023-05-25
Converting strings into Swift safe names can lead to conflicts


czechboy0:
An example of a schema that today the generator produces non-compiling Swift code for, as both enum cases are named `_1`.

```yaml
type: string
enum:
  - "+1"
  - "-1"
```

Some ideas of solving this:
- word-ify common symbols, so instead of turning `+` into `_`, turn it into `plus`, ending up with `plus1` and `minus1` here

It's likely that no solution will completely solve this problem, as it's always going to be possible to create a conflict, as not all characters that are valid as an OpenAPI name are valid as Swift identifiers, so some mapping is necessary.

czechboy0:
For folks interested in this issue being addressed, please chime in this issue: https://github.com/apple/swift-openapi-generator/issues/112

czechboy0:
[SOAR-0001](https://github.com/apple/swift-openapi-generator/blob/main/Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0001.md) landed in 0.1.6 and can be tried out by enabling the feature flag `proposal0001`. It should address this issue by producing much fewer conflicts even when the identifier differs by an illegal character.

PR #89, opened 2023-06-22
[SOAR-0001] Implement Encoding for Property Names

denil-ct:
## Motivation

Encoding would help distinguish types properly. For eg. both `a b` and `a_b` would be rendered as `a_b`. This change would make it render `a20b` and `a_b` respectively.

## Modifications

Added a hexadecimal encoding for un-supported characters rather than having just an underscore for everything.
Removed redundant keywords.

## Result
Proper differentiation between names as explained in the motivation, along with that, it also reduces the reserved keywords set by removing symbols and any keywords that prepend such symbols.

## Test Plan
Tested and verified for several combinations of special characters in various positions in-between alphanumeric characters.


czechboy0:
Hey @denil-ct, thanks for breaking this out.

I'm on the fence about this.

On the plus side, it helps disambiguate special symbols, so it even helps with #21.

On the other hand, the hex code can seem a little arbitrary.

To help us see what these will look like in practice, can you add a bunch of examples to this unit tests? Like 10 different ones with various special symbols in different parts of the name, to help us see the impact. https://github.com/apple/swift-openapi-generator/blob/main/Tests/OpenAPIGeneratorCoreTests/Extensions/Test_String.swift#L19

Thanks 🙏

denil-ct:
Sure, will add some. I went with hex because I was influenced by the comment stating similar to url encoding. If the numbers seem a bit arbitrary we can explore some other form as well.

czechboy0:
Yeah, it's a fair starting point. @simonjbeaumont recently added a special case for empty strings (`_empty`), and we could even have a hard-coded map for common symbols, e.g. `+` -> `plus`, `#` -> `pound`, etc, so that the hex would only happen for rarely used symbols.

One thing to keep in mind - this is an API-breaking change, so will have to be in a minor bump, as if people update the generator, they'd get different symbol names. That's okay, just making it explicit here.

simonjbeaumont:
@denil-ct thanks for thinking about this one. 

One thing I _do_ like about this approach is it generalises. For replacements within a string we might also benefit from some delimiting to make it clearer it's the result of mangling.

> Yeah, it's a fair starting point. @simonjbeaumont recently added a special case for empty strings (_empty), and we could even have a hard-coded map for common symbols, e.g. + -> plus, # -> pound, etc, so that the hex would only happen for rarely used symbols.

We could find a middle ground. I think that the empty string probably deserves its own special case because there's nothing else to meaningfully identify it—i.e. having just `_20` is pretty uninformative.

It makes me wonder if the heuristic might be to only make a "wordy" replacement when there's nothing else other than the symbol? 

This one is definitely a tradeoff and anything we do will have some drawback. 

Another compromise here could be to generate the hex-encoded thing always and a type alias and or function wrapper with a nice name for a few cases, e.g. `+1` would generate `_2b1` and an alias `_plus1`. We can then have something that's uniform with some opt-in syntactic sugar?

czechboy0:
Mostly on the same page with you both, but since this will have a pretty visible impact on the naming logic, could you do a proposal (see our docs for the template)? Doesn't have to be long, just describe the overall naming logic:
- use as-is if all characters are legal for a Swift identifier, and doesn't start with a number 
- prefixed with an underscore when starts with a number 
- each character is checked, if it's not safe in a Swift identifier, then it's looked up in a hard-coded map (list the hard-coded characters and their word replacements), and if it's not in the map, is converted to its hex code

The above with a bunch of diverse examples of inputs and outputs.

Doing the typealias would be a bit tricky due to namespacing and uniqueness, so I'd prefer if there was exactly one Swift-safe term for an input string.

Again, this _will_ be a breaking change, so we want to make sure we understand the full implications.

denil-ct:
Sure, that sounds good. Will create one and we'll proceed from there.

czechboy0:
Can you put the PR into a draft state (or close it) until we have a proposal? Just to make clear that you're not waiting on us to take any action, like a review - thanks. 🙂

denil-ct:
Ah sure, no issues.

denil-ct:
Hey @czechboy0 @simonjbeaumont 

I have updated the PR description with the proposal. Do let me know if anything is missing, and what the next steps are. Also should I pull this out of draft state?

czechboy0:
Thanks for writing this up, @denil-ct! Could you add the proposal as a markdown file to the Proposals directory, next to the template? That'll allow us to comment on specific lines, thanks! You can then remove it from the PR description, to avoid having to update two places as we iterate.

Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0001.md
```diff
@@ -0,0 +1,52 @@
...
+
+The implementation for this is quite simple as you can see in https://github.com/apple/swift-openapi-generator/pull/89, we just made changes to the substitution logic where it used to substitute with `_`. We now add an additional encoding to the special character before substituting it. Contributors should be aware of this change and should review the places where they use this extension and evaluate if its suitable for them with this change.
+
+### API stability
+
+This is an API breaking change, as it will produce different symbol names than before. Other components such as the runtime and transports should not have any impacts.
+
+### Future directions
+
+The encoding strategy is open for further discussion. As a starting point, we have chosen the most simplest encoding format of hex. One of the reasons for this the hex encoding adds quite arbitrary symbols to the property name, which is not ideal. We could go towards a middle of the road approach where we have wordified versions of the special characters which we can map to. For example `a+b` can be `aplusb` or `a_plus_b` to add some kind of delimiter to specify the replaced portion.
```
czechboy0, line 52: I'd like to see the map of special characters to names in this proposal, not just as a future direction. I like that you started without it, as it's simpler, but since this is a breaking change anyway, we might as well make the experience nicer for a hard-coded set of characters.

I'd start with: +, -, #, and also take a look at identifiers used in some large OpenAPI docs, such as GitHub's.

I'll also let @simonjbeaumont suggest other ones that come to mind.

denil-ct: Sure, that makes sense. We'll land on a list of symbols to replace and I'll start working on it after.

denil-ct: I have added some of the common ones I can think of. Feel free to add on to the list. For any other character not in the list, we can map it to raw hex value. We can maybe prefix it with `x` like `x20` to indicate its a hex value.

```swift
let specialCharacters: [String: String] = [
    "+": "plus",
    "-": "minus",
    "*": "asterisk",
    "/": "slash",
    "\\": "backSlash",
    "=": "equals",
    "<": "lessThan",
    ">": "greaterThan",
    "!": "exclamationMark",
    "?": "questionMark",
    "@": "atSign",
    "#": "hashSign",
    "$": "dollarSign",
    "%": "percentSign",
    "&": "ampersand",
    "'": "apostrophe",
    "\"": "quotationMark",
    "(": "leftParenthesis",
    ")": "rightParenthesis",
    "[": "leftBracket",
    "]": "rightBracket",
    "{": "leftBrace",
    "}": "rightBrace",
    ".": "period",
    ",": "comma",
    ":": "colon",
    ";": "semicolon",
    "`": "graveAccent",
]
```

simonjbeaumont: > And I don't see a meaningful difference between the likelihood of conflicts for hex and pretty names, and I'd prefer to use the more readable option, so that adopters don't have to jump to definition to see which code maps to which symbol.

Hm, I think it's much more likely that someone would use e.g. `plus` and `+` in the same doc than using `2b` and `+` in the same doc.

But overall I'm sympathetic to the desire for nicer names so I'm willing to go for it.

I'm not sure about the mapping proposed in this thread, however. Maybe we should use this established convention for the mapping for brevity and familiarity: https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references.

denil-ct: Ah great, didn't know this existed. We can use this as its a standardized approach.

simonjbeaumont: > Is the symbol list suggested above exhaustive enough, or do we need to add more to it?

IMO this is why we should have these pretty names be aliases so we can add to them in the future without worrying about breaking APIs. For example, I think we should be generating `_2b1` and `plus1` where `plus1` is a wrapper around the escaped name.

This way we allow ourselves room to evolve—we can start with a small list and add more as the need arises.

However, if we really wanted to go full-tilt and add then I think we need to do some more research.

Looking at the OpenAPI spec, they do mention the notion of "reserved characters" which can be enabled in parameter names: https://swagger.io/specification#parameter-object, https://tools.ietf.org/html/rfc3986#section-2.2.

@czechboy0 might know if this should be read as the exhaustive set or whether there are more we need to support?

czechboy0: The reserved characters section is just related to query parameters, as those appear in the URL. But we have the need to create safe Swift identifiers in a few more places: type names, property names, property names (that aren't query), and operation ids.

The list doesn't contain `-`, which is legal in a URL without escaping, but isn't a legal character in a Swift identifier, so we'll still have to have our own list. But, it might be a good starting point that we can add to.

Alternatively, while that might be a bit more work, but at least is pretty scientific - we could consider the ASCII printable range for a pretty name only, and hex anything we don't choose to give a pretty name to. The range is 20 to 7E.

Skipping digits and letters, we get:

` !"#$%&'()*+,-./:;<=>?@[\]^{}|~` and backtick (I couldn't type it here, as it's the char to end code font 🙂)

So the starting point could be - let's use the HTML naming Si suggested and find a pretty name for all of these, and when we're asked, we can say that we handle all printable ASCII in a nice way, and everything else gets a hex.

denil-ct: So, this would be the list for now then.

| Character | Name |
|:---:|:-----------:|
| | space |
| ! | excl |
| " | quot |
| # | num |
| $ | dollar |
| % | percnt |
| & | amp |
| ' | apos |
| ( | lpar |
| ) | rpar |
| * | ast |
| + | plus |
| , | comma |
| - | hyphen |
| . | period |
| / | sol |
| : | colon |
| ; | semi |
| < | lt |
| = | equals |
| > | gt |
| ? | quest |
| @ | commat |
| [ | lbrack |
| \ | bsol |
| ] | rbrack |
| ^ | hat |
| ` | grave |
| { | lcub |
| \| | verbar |
| } | rcub |
| ~ | tilde |

czechboy0: Yup the set looks good to me. The names aren't great, but they're standard so I'm fine with them, as we can defend where we got them.

simonjbeaumont: Yeah, these names are pretty janky but it's nice to stick with something where there's precedence.

denil-ct:
Have added the proposed changes, along with updated examples.

czechboy0:
Great! One more thing, we just recently updated the proposal process a tiny bit, see https://github.com/apple/swift-openapi-generator/pull/91

So can you split the proposal into a separate PR from the code change, please? (Thanks for helping us figure out this overall process by test-driving it 🙂)

simonjbeaumont:
> Great! One more thing, we just recently updated the proposal process a tiny bit, see #91
> 
> So can you split the proposal into a separate PR from the code change, please? (Thanks for helping us figure out this overall process by test-driving it 🙂)

The docs you have here (about how we handle special characters) are pretty useful to folks generally beyond those reviewing the code change. @czechboy0 maybe the proposal should be replaced by an addition to the DocC?

Sources/_OpenAPIGeneratorCore/Extensions/String.swift
```diff
@@ -56,18 +56,19 @@ fileprivate extension String {
 
     /// Returns a string sanitized to be usable as a Swift identifier.
     ///
-    /// For example, the string `$nake` would be returned as `_nake`, because
-    /// the dollar sign is not a valid character in a Swift identifier.
+    /// For example, the string `$nake…` would be returned as `_dollar_nake_x2026_`, because
+    /// both the dollar and ellipsis sign are not valid characters in a Swift identifier.
+    /// So, it replaces such characters with their html enity equivalents or unicode hex representation,
```
siemensikkema, line 61:
```suggestion
    /// So, it replaces such characters with their html entity equivalents or unicode hex representation,
```

Sources/_OpenAPIGeneratorCore/Extensions/String.swift
```diff
@@ -56,18 +56,19 @@ fileprivate extension String {
...
+    /// in case its not present in the `specialCharsMap`. It marks this replacement with `_` as a delimiter.
```
siemensikkema, line 62:
```suggestion
    /// in case it’s not present in the `specialCharsMap`. It marks this replacement with `_` as a delimiter.
```

denil-ct:
@czechboy0 @simonjbeaumont 

Should I proceed with splitting up the proposal and implementation?

simonjbeaumont:
@denil-ct First off, thank you for jumping in—on this and other issues!

We'll want to proceed with a bit more caution with this change compared to the other contributions, as the choices we make here—and any subsequent changes to the area—will have an impact on the generated API.

This might mean waiting for us to have tests around the API changes in generated code, so I just wanted to manage expectations on your side.

> Should I proceed with splitting up the proposal and implementation?

Yes, let's first land this as a proposal and let others chime in with any feedback before we land the implementation.

When we do land the implementation, we'll want to have some good testing around this too.

denil-ct:
Marking as draft until the proposal is approved.

denil-ct:
@czechboy0 please add a blocked tag here as well

czechboy0:
Done ✅

czechboy0:
@denil-ct Ok, you're unblocked now, please provide the new implementation in [this](https://github.com/apple/swift-openapi-generator/blob/0fc97c53ca497c35aee620135a3f469dda61ae8d/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/SwiftSafeNames.swift#L79) method, and add a bunch of unit tests for it please.

czechboy0:
Also, changing the label from `semver/minor` to `semver/patch`, because this will be hidden behind a feature flag in 0.1.x. Only once we enable the feature flag unconditionally in 0.2.x will it be a breaking change, so we'll mark `semver/minor` _that_ PR, not this one.

czechboy0 (approved):
Awesome, thank you!

Just a reminder, it's currently disabled behind the `proposal0001` feature flag, but once it lands, anyone can test it out by enabling the feature flag (in CLI or in their config file).

The feature will be enabled unconditionally in the next breaking version, i.e. 0.2.0.

PR #98, opened 2023-07-02
Command Plugin for generating the source code

MahdiBM:
### Motivation

This PR adds the option to use the package as a Command plugin instead of a BuildTool plugin.

This benefits those who use heavy OpenAPI documents, and prefer not to have to wait for an extra round of OpenAPI code generation which can be accidentally triggered at times, for example if you clean your build folder.

The whole idea of creating this Command plugin came after @czechboy0 's comment here: https://github.com/apple/swift-openapi-generator/pull/96#issuecomment-1615969962

### Modifications

Generally, add a Command plugin target to the package, plus modifying the functions etc... to match/allow this addition.

### Result

There is a new Command plugin, and users can choose between the Command plugin and the BuildTool plugin at will.

### Test Plan

As visible in the PR discussions below, we've done enough manual-testing of the Command plugin.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -156,7 +156,7 @@ let package = Package(
         // Command Plugin
         .plugin(
             name: "OpenAPIGeneratorCommand",
-            capability: Target.PluginCapability.command(
+            capability: .command(
                 intent: .custom(
                     verb: "generate-openapi-code",
                     description: "Generates Swift code from an OpenAPI document."
--- a/Package.swift
+++ b/Package.swift
@@ -158,7 +158,7 @@ let package = Package(
             name: "OpenAPIGeneratorCommand",
             capability: .command(
                 intent: .custom(
-                    verb: "generate-openapi-code",
+                    verb: "generate-code-from-openapi",
                     description: "Generate Swift code from an OpenAPI document."
                 ),
                 permissions: [.writeToPackageDirectory(reason: "To write the generated Swift files back into the source directory of the package.")]
--- a/Package.swift
+++ b/Package.swift
@@ -33,7 +33,7 @@ let package = Package(
         // The platforms below are not currently supported for running
         // the generator itself. We include them here to allow the generator
         // to emit a more descriptive compiler error.
-        .iOS("99"), .tvOS("99"), .watchOS("99"),
+        .iOS(.v13), .tvOS(.v13), .watchOS(.v6),
     ],
     products: [
         .executable(name: "swift-openapi-generator", targets: ["swift-openapi-generator"]),
--- a/Package.swift
+++ b/Package.swift
@@ -160,7 +160,7 @@ let package = Package(
                 permissions: [.writeToPackageDirectory(reason: "To write the generated Swift files back into the source directory of the package.")]
             ),
             dependencies: [
-                "swift-openapi-generator",
+                "swift-openapi-generator"
             ]
         )
     ]
--- a/Package.swift
+++ b/Package.swift
@@ -159,7 +159,7 @@ let package = Package(
             capability: .command(
                 intent: .custom(
                     verb: "generate-openapi-code",
-                    description: "Generates Swift code from an OpenAPI document."
+                    description: "Generate Swift code from an OpenAPI document."
                 ),
                 permissions: [.writeToPackageDirectory(reason: "To write the generated Swift files back into the source directory of the package.")]
             ),
--- a/Package.swift
+++ b/Package.swift
@@ -15,11 +15,15 @@
 import PackageDescription
 
 // General Swift-settings for all targets.
-let swiftSettings: [SwiftSetting] = [
+var swiftSettings: [SwiftSetting] = []
+
+#if swift(>=5.9)
+swiftSettings.append(
     // https://github.com/apple/swift-evolution/blob/main/proposals/0335-existential-any.md
     // Require `any` for existential types.
     .enableUpcomingFeature("ExistentialAny")
-]
+)
+#endif
 
 let package = Package(
     name: "swift-openapi-generator",
--- a/Package.swift
+++ b/Package.swift
@@ -161,11 +161,15 @@ let package = Package(
                     verb: "generate-code-from-openapi",
                     description: "Generate Swift code from an OpenAPI document."
                 ),
-                permissions: [.writeToPackageDirectory(reason: "To write the generated Swift files back into the source directory of the package.")]
+                permissions: [
+                    .writeToPackageDirectory(
+                        reason: "To write the generated Swift files back into the source directory of the package."
+                    )
+                ]
             ),
             dependencies: [
                 "swift-openapi-generator"
             ]
-        )
+        ),
     ]
 )
--- a/Plugins/OpenAPIGeneratorCommand/plugin.swift
+++ b/Plugins/OpenAPIGeneratorCommand/plugin.swift
@@ -45,77 +45,63 @@ extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
         context: PluginContext,
         arguments: [String]
     ) async throws {
-        let targetName = try parseTargetName(from: arguments)
+        guard arguments.count == 2,
+              arguments[0] == "--target"
+        else {
+            throw PluginError.badArguments(arguments)
+        }
+        let targetName = arguments[1]
+
         let matchingTargets = try context.package.targets(named: [targetName])
 
         switch matchingTargets.count {
         case 0:
             throw PluginError.noTargetsMatchingTargetName(targetName: targetName)
         case 1:
-            let target = matchingTargets[0]
-            guard let swiftTarget = target as? SwiftSourceModuleTarget else {
-                throw PluginError.incompatibleTarget(targetName: target.name)
+            let mainTarget = matchingTargets[0]
+            guard mainTarget is SwiftSourceModuleTarget else {
+                throw PluginError.incompatibleTarget(targetName: mainTarget.name)
             }
-            try runCommand(
-                targetWorkingDirectory: target.directory,
-                tool: context.tool,
-                sourceFiles: swiftTarget.sourceFiles,
-                targetName: target.name
-            )
+            let allDependencies = mainTarget.recursiveTargetDependencies
+            let packageTargets = Set(context.package.targets.map(\.id))
+            let dependenciesInPackage = allDependencies.filter { packageTargets.contains($0.id) }
+
+            var hadASuccessfulRun = false
+
+            for target in [mainTarget] + dependenciesInPackage {
+                guard let swiftTarget = target as? SwiftSourceModuleTarget else {
+                    continue
+                }
+                do {
+                    try runCommand(
+                        targetWorkingDirectory: target.directory,
+                        tool: context.tool,
+                        sourceFiles: swiftTarget.sourceFiles,
+                        targetName: target.name
+                    )
+                    hadASuccessfulRun = true
+                } catch let error as PluginError {
+                    if error.isDefiniteMisconfigurationError {
+                        throw error
+                    }
+                }
+
+                guard hadASuccessfulRun else {
+                    // None of the targets had a successful run and none had any definite
+                    // indication of trying to be OpenAPI-generator compatible.
+                    // To follow how the build command works, we throw file errors to let users
+                    // know they need to add these files to their target.
+                    throw PluginError.fileErrors(
+                        FileError.notFoundFileErrors(forTarget: mainTarget.name),
+                        targetName: mainTarget.name
+                    )
+                }
+            }
         default:
             throw PluginError.tooManyTargetsMatchingTargetName(
                 targetName: targetName,
                 matchingTargetNames: matchingTargets.map(\.name)
             )
         }
-    }
-}
-
-#if canImport(XcodeProjectPlugin)
-import XcodeProjectPlugin
-
-extension SwiftOpenAPIGeneratorPlugin: XcodeCommandPlugin {
-    func performCommand(
-        context: XcodePluginContext,
-        arguments: [String]
-    ) throws {
-        let targetName = try parseTargetName(from: arguments)
-        let matchingTargets = context.xcodeProject.targets.filter {
-            $0.displayName == targetName
-        }
-
-        switch matchingTargets.count {
-        case 0:
-            throw PluginError.noTargetsMatchingTargetName(targetName: targetName)
-        case 1:
-            let xcodeTarget = matchingTargets[0]
-            guard let target = xcodeTarget as? SourceModuleTarget else {
-                throw PluginError.incompatibleTarget(targetName: xcodeTarget.displayName)
-            }
-            try runCommand(
-                targetWorkingDirectory: target.directory,
-                tool: context.tool,
-                sourceFiles: xcodeTarget.inputFiles,
-                targetName: xcodeTarget.displayName
-            )
-        default:
-            throw PluginError.tooManyTargetsMatchingTargetName(
-                targetName: targetName,
-                matchingTargetNames: matchingTargets.map(\.displayName)
-            )
-        }
-    }
-}
-#endif
-
-extension SwiftOpenAPIGeneratorPlugin {
-    /// Parses the target name from the arguments.
-    func parseTargetName(from arguments: [String]) throws -> String {
-        guard arguments.count == 2,
-              arguments[0] == "--target"
-        else {
-            throw PluginError.badArguments(arguments)
-        }
-        return arguments[1]
     }
 }
--- a/Plugins/OpenAPIGeneratorCommand/plugin.swift
+++ b/Plugins/OpenAPIGeneratorCommand/plugin.swift
@@ -60,9 +60,7 @@ extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
             let enumeratedKeyValues = withLocalDependencies.map(\.id).enumerated().map { (key: $0.element, value: $0.offset) }
             let indexLookupTable = Dictionary(enumeratedKeyValues, uniquingKeysWith: { l, _ in l })
             let groupedByID = Dictionary(grouping: withLocalDependencies, by: \.id)
-            let sortedUniqueTargets = groupedByID.map(\.value[0]).sorted {
-                indexLookupTable[$0.id, default: 0] < indexLookupTable[$1.id, default: 0]
-            }
+            let sortedUniqueTargets = groupedByID.map(\.value[0]).sorted { indexLookupTable[$0.id, default: 0] < indexLookupTable[$1.id, default: 0] }
             targets = sortedUniqueTargets
         }
 
--- a/Plugins/OpenAPIGeneratorCommand/plugin.swift
+++ b/Plugins/OpenAPIGeneratorCommand/plugin.swift
@@ -63,10 +63,10 @@ extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
                     errors.append((error, target.name))
                 }
             }
+            try throwErrorsIfNecessary(errors)
             if !hasHadASuccessfulRun {
                 throw PluginError.noTargetsFoundForCommandPlugin
             }
-            try throwErrorsIfNecessary(errors)
         case .target(let targetName):
             let matchingTargets = try context.package.targets(named: [targetName])
             // `matchingTargets.count` can't be 0 because
@@ -116,10 +116,10 @@ extension SwiftOpenAPIGeneratorPlugin: XcodeCommandPlugin {
                     errors.append((error, target.name))
                 }
             }
+            try throwErrorsIfNecessary(errors)
             if !hasHadASuccessfulRun {
                 throw PluginError.noTargetsFoundForCommandPlugin
             }
-            try throwErrorsIfNecessary(errors)
         case .target(let targetName):
             guard let xcodeTarget = context.xcodeProject.targets.first(
                 where: { $0.displayName == targetName }
--- a/Plugins/OpenAPIGeneratorCommand/plugin.swift
+++ b/Plugins/OpenAPIGeneratorCommand/plugin.swift
@@ -52,9 +52,6 @@ extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
             let targetNames = try parseTargetNames(arguments: arguments)
             print("Running OpenAPI generator CommandPlugin on targets: \(targetNames)")
             targets = try context.package.targets(named: Array(targetNames))
-            guard !targets.isEmpty else {
-                throw PluginError.noTargetsMatchingTargetNames(targetNames: Array(targetNames))
-            }
         }
 
         var hasHadASuccessfulRun = false
@@ -75,7 +72,7 @@ extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
                 errors.append((error, target.name))
             }
         }
-        try throwErrorsIfNecessary(errors, targetCount: targets.count)
+        try throwErrorsIfNecessary(errors)
         if !hasHadASuccessfulRun {
             throw PluginError.noTargetsFoundForCommandPlugin
         }
@@ -93,28 +90,26 @@ extension SwiftOpenAPIGeneratorPlugin: XcodeCommandPlugin {
         // On Xcode, it automatically includes all targets when you run the plugin.
         let targetNames = try parseTargetNames(arguments: arguments)
         print("Running OpenAPI generator CommandPlugin on targets: \(targetNames)")
-        let targets = context.xcodeProject.targets
-            .compactMap { $0 as? SourceModuleTarget }
-            .filter { targetNames.contains($0.name) }
-        guard !targets.isEmpty else {
-            throw PluginError.noTargetsMatchingTargetNames(targetNames: Array(targetNames))
-        }
+        let targets = context.xcodeProject.targets.filter { targetNames.contains($0.displayName) }
         var hasHadASuccessfulRun = false
         var errors = [(error: any Error, targetName: String)]()
-        for target in targets {
+        for xcodeTarget in targets {
+            guard let target = xcodeTarget as? SourceModuleTarget else {
+                continue
+            }
             do {
                 try runCommand(
                     targetWorkingDirectory: target.directory,
                     tool: context.tool,
-                    sourceFiles: target.sourceFiles,
-                    targetName: target.name
+                    sourceFiles: xcodeTarget.inputFiles,
+                    targetName: xcodeTarget.displayName
                 )
                 hasHadASuccessfulRun = true
             } catch {
                 errors.append((error, target.name))
             }
         }
-        try throwErrorsIfNecessary(errors, targetCount: targets.count)
+        try throwErrorsIfNecessary(errors)
         if !hasHadASuccessfulRun {
             throw PluginError.noTargetsFoundForCommandPlugin
         }
@@ -140,21 +135,13 @@ extension SwiftOpenAPIGeneratorPlugin {
         return targets
     }
     
-    func throwErrorsIfNecessary(
-        _ errors: [(error: any Error, targetName: String)],
-        targetCount: Int
-    ) throws {
-        // Always throw any existing error if only one target.
-        if targetCount == 1, errors.count == 1 {
-            throw errors[0].error
-        }
-
+    func throwErrorsIfNecessary(_ errors: [(error: any Error, targetName: String)]) throws {
         let errorsToBeReported = errors.compactMap {
             (error, targetName) -> PluginError? in
             if let error = error as? PluginError {
                 switch error {
-                case .fileErrors(let fileErrors, _):
-                    if fileErrors.count != FileError.Kind.allCases.count {
+                case .fileErrors(let errors, _):
+                    if errors.count != FileError.Kind.allCases.count {
                         // There are some file-finding errors but at least 1 file is available.
                         // This means the user means to use the target with the generator, just
                         // hasn't configured their target properly.
--- a/Plugins/OpenAPIGeneratorCommand/plugin.swift
+++ b/Plugins/OpenAPIGeneratorCommand/plugin.swift
@@ -50,7 +50,7 @@ extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
             print("Running OpenAPI generator CommandPlugin all targets")
         } else {
             let targetNames = try parseTargetNames(arguments: arguments)
-            print("Running OpenAPI generator CommandPlugin on select targets: \(targetNames)")
+            print("Running OpenAPI generator CommandPlugin on targets: \(targetNames)")
             targets = try context.package.targets(named: Array(targetNames))
             guard !targets.isEmpty else {
                 throw PluginError.noTargetsMatchingTargetNames(targetNames: Array(targetNames))
@@ -97,7 +97,7 @@ extension SwiftOpenAPIGeneratorPlugin: XcodeCommandPlugin {
         let targets = context.xcodeProject.targets
             .compactMap { $0 as? SourceModuleTarget }
             .filter { targetNames.contains($0.name) }
-            .flatMap { $0.dependencies }
+            .flatMap { $0.recursiveTargetDependencies }
             .compactMap { $0 as? SourceModuleTarget }
         print("Running OpenAPI generator CommandPlugin on targets: \(targets.map(\.name))")
         guard !targets.isEmpty else {
@@ -128,7 +128,6 @@ extension SwiftOpenAPIGeneratorPlugin: XcodeCommandPlugin {
 
 extension SwiftOpenAPIGeneratorPlugin {
     func parseTargetNames(arguments: [String]) throws -> Set<String> {
-
         guard arguments.count % 2 == 0 else {
             throw PluginError.badArguments(arguments: arguments)
         }
--- a/Plugins/OpenAPIGeneratorCommand/plugin.swift
+++ b/Plugins/OpenAPIGeneratorCommand/plugin.swift
@@ -64,11 +64,11 @@ extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
             }
             let allDependencies = mainTarget.recursiveTargetDependencies
             let packageTargets = Set(context.package.targets.map(\.id))
-            let dependenciesInPackage = allDependencies.filter { packageTargets.contains($0.id) }
+            let localDependencies = allDependencies.filter { packageTargets.contains($0.id) }
 
             var hadASuccessfulRun = false
 
-            for target in [mainTarget] + dependenciesInPackage {
+            for target in [mainTarget] + localDependencies {
                 guard let swiftTarget = target as? SwiftSourceModuleTarget else {
                     continue
                 }
@@ -90,7 +90,7 @@ extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
             guard hadASuccessfulRun else {
                 throw PluginError.noTargetOrDependenciesWithExpectedFiles(
                     targetName: mainTarget.name,
-                    dependencyNames: dependenciesInPackage.map(\.name)
+                    dependencyNames: localDependencies.map(\.name)
                 )
             }
         default:
--- a/Plugins/OpenAPIGeneratorCommand/plugin.swift
+++ b/Plugins/OpenAPIGeneratorCommand/plugin.swift
@@ -89,7 +89,8 @@ extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
 
             guard hadASuccessfulRun else {
                 throw PluginError.noTargetOrDependenciesWithExpectedFiles(
-                    targetName: mainTarget.name
+                    targetName: mainTarget.name,
+                    dependencyNames: dependenciesInPackage.map(\.name)
                 )
             }
         default:
--- a/Plugins/OpenAPIGeneratorCommand/plugin.swift
+++ b/Plugins/OpenAPIGeneratorCommand/plugin.swift
@@ -37,6 +37,9 @@ struct SwiftOpenAPIGeneratorPlugin {
         process.environment = [:]
         try process.run()
         process.waitUntilExit()
+        guard process.terminationStatus == 0 else {
+            throw PluginError.generatorFailure(targetName: targetName)
+        }
     }
 }
 
@@ -56,11 +59,11 @@ extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
 
         switch matchingTargets.count {
         case 0:
-            throw PluginError.noTargetsMatchingTargetName(targetName: targetName)
+            throw PluginError.noTargetsMatchingTargetName(targetName)
         case 1:
             let mainTarget = matchingTargets[0]
             guard mainTarget is SwiftSourceModuleTarget else {
-                throw PluginError.incompatibleTarget(targetName: mainTarget.name)
+                throw PluginError.incompatibleTarget(name: mainTarget.name)
             }
             let allDependencies = mainTarget.recursiveTargetDependencies
             let packageTargets = Set(context.package.targets.map(\.id))
@@ -81,7 +84,7 @@ extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
                     )
                     hadASuccessfulRun = true
                 } catch let error as PluginError {
-                    if error.isDefiniteMisconfigurationError {
+                    if error.isMisconfigurationError {
                         throw error
                     }
                 }
@@ -95,7 +98,7 @@ extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
             }
         default:
             throw PluginError.tooManyTargetsMatchingTargetName(
-                targetName: targetName,
+                targetName,
                 matchingTargetNames: matchingTargets.map(\.name)
             )
         }
--- a/Plugins/OpenAPIGeneratorCommand/plugin.swift
+++ b/Plugins/OpenAPIGeneratorCommand/plugin.swift
@@ -90,16 +90,12 @@ extension SwiftOpenAPIGeneratorPlugin: XcodeCommandPlugin {
         context: XcodePluginContext,
         arguments: [String]
     ) throws {
-        // On Xcode, it automatically includes all targets of the scheme when you run the plugin.
-        // We search for recursive dependencies to make sure we don't miss a target just because of
-        // scheme settings.
+        // On Xcode, it automatically includes all targets when you run the plugin.
         let targetNames = try parseTargetNames(arguments: arguments)
+        print("Running OpenAPI generator CommandPlugin on targets: \(targetNames)")
         let targets = context.xcodeProject.targets
             .compactMap { $0 as? SourceModuleTarget }
             .filter { targetNames.contains($0.name) }
-            .flatMap { $0.recursiveTargetDependencies }
-            .compactMap { $0 as? SourceModuleTarget }
-        print("Running OpenAPI generator CommandPlugin on targets: \(targets.map(\.name))")
         guard !targets.isEmpty else {
             throw PluginError.noTargetsMatchingTargetNames(targetNames: Array(targetNames))
         }
--- a/Plugins/OpenAPIGeneratorCommand/plugin.swift
+++ b/Plugins/OpenAPIGeneratorCommand/plugin.swift
@@ -52,6 +52,9 @@ extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
             let targetNames = try parseTargetNames(arguments: arguments)
             print("Running OpenAPI generator CommandPlugin on targets: \(targetNames)")
             targets = try context.package.targets(named: Array(targetNames))
+            guard !targets.isEmpty else {
+                throw PluginError.noTargetsMatchingTargetNames(targetNames: Array(targetNames))
+            }
         }
 
         var hasHadASuccessfulRun = false
@@ -72,7 +75,7 @@ extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
                 errors.append((error, target.name))
             }
         }
-        try throwErrorsIfNecessary(errors)
+        try throwErrorsIfNecessary(errors, targetCount: targets.count)
         if !hasHadASuccessfulRun {
             throw PluginError.noTargetsFoundForCommandPlugin
         }
@@ -90,26 +93,28 @@ extension SwiftOpenAPIGeneratorPlugin: XcodeCommandPlugin {
         // On Xcode, it automatically includes all targets when you run the plugin.
         let targetNames = try parseTargetNames(arguments: arguments)
         print("Running OpenAPI generator CommandPlugin on targets: \(targetNames)")
-        let targets = context.xcodeProject.targets.filter { targetNames.contains($0.displayName) }
+        let targets = context.xcodeProject.targets
+            .compactMap { $0 as? SourceModuleTarget }
+            .filter { targetNames.contains($0.name) }
+        guard !targets.isEmpty else {
+            throw PluginError.noTargetsMatchingTargetNames(targetNames: Array(targetNames))
+        }
         var hasHadASuccessfulRun = false
         var errors = [(error: any Error, targetName: String)]()
-        for xcodeTarget in targets {
-            guard let target = xcodeTarget as? SourceModuleTarget else {
-                continue
-            }
+        for target in targets {
             do {
                 try runCommand(
                     targetWorkingDirectory: target.directory,
                     tool: context.tool,
-                    sourceFiles: xcodeTarget.inputFiles,
-                    targetName: xcodeTarget.displayName
+                    sourceFiles: target.sourceFiles,
+                    targetName: target.name
                 )
                 hasHadASuccessfulRun = true
             } catch {
                 errors.append((error, target.name))
             }
         }
-        try throwErrorsIfNecessary(errors)
+        try throwErrorsIfNecessary(errors, targetCount: targets.count)
         if !hasHadASuccessfulRun {
             throw PluginError.noTargetsFoundForCommandPlugin
         }
@@ -135,13 +140,21 @@ extension SwiftOpenAPIGeneratorPlugin {
         return targets
     }
     
-    func throwErrorsIfNecessary(_ errors: [(error: any Error, targetName: String)]) throws {
+    func throwErrorsIfNecessary(
+        _ errors: [(error: any Error, targetName: String)],
+        targetCount: Int
+    ) throws {
+        // Always throw any existing error if only one target.
+        if targetCount == 1, errors.count == 1 {
+            throw errors[0].error
+        }
+
         let errorsToBeReported = errors.compactMap {
             (error, targetName) -> PluginError? in
             if let error = error as? PluginError {
                 switch error {
-                case .fileErrors(let errors, _):
-                    if errors.count != FileError.Kind.allCases.count {
+                case .fileErrors(let fileErrors, _):
+                    if fileErrors.count != FileError.Kind.allCases.count {
                         // There are some file-finding errors but at least 1 file is available.
                         // This means the user means to use the target with the generator, just
                         // hasn't configured their target properly.
--- a/Plugins/OpenAPIGeneratorCommand/plugin.swift
+++ b/Plugins/OpenAPIGeneratorCommand/plugin.swift
@@ -83,7 +83,7 @@ extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
                     sourceFiles: swiftTarget.sourceFiles,
                     targetName: target.name
                 )
-                print("- OpenAPI code generation successfully completed.")
+                print("- ✅ OpenAPI code generation successfully completed.")
                 hadASuccessfulRun = true
             } catch let error as PluginError {
                 if error.isMisconfigurationError {
--- a/Plugins/OpenAPIGeneratorCommand/plugin.swift
+++ b/Plugins/OpenAPIGeneratorCommand/plugin.swift
@@ -50,7 +50,7 @@ extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
             print("Running OpenAPI generator CommandPlugin all targets")
         } else {
             let targetNames = try parseTargetNames(arguments: arguments)
-            print("Running OpenAPI generator CommandPlugin on targets: \(targetNames)")
+            print("Running OpenAPI generator CommandPlugin on select targets: \(targetNames)")
             targets = try context.package.targets(named: Array(targetNames))
             guard !targets.isEmpty else {
                 throw PluginError.noTargetsMatchingTargetNames(targetNames: Array(targetNames))
@@ -97,7 +97,7 @@ extension SwiftOpenAPIGeneratorPlugin: XcodeCommandPlugin {
         let targets = context.xcodeProject.targets
             .compactMap { $0 as? SourceModuleTarget }
             .filter { targetNames.contains($0.name) }
-            .flatMap { $0.recursiveTargetDependencies }
+            .flatMap { $0.dependencies }
             .compactMap { $0 as? SourceModuleTarget }
         print("Running OpenAPI generator CommandPlugin on targets: \(targets.map(\.name))")
         guard !targets.isEmpty else {
@@ -128,6 +128,7 @@ extension SwiftOpenAPIGeneratorPlugin: XcodeCommandPlugin {
 
 extension SwiftOpenAPIGeneratorPlugin {
     func parseTargetNames(arguments: [String]) throws -> Set<String> {
+
         guard arguments.count % 2 == 0 else {
             throw PluginError.badArguments(arguments: arguments)
         }
--- a/Plugins/OpenAPIGeneratorCommand/plugin.swift
+++ b/Plugins/OpenAPIGeneratorCommand/plugin.swift
@@ -27,14 +27,16 @@ struct SwiftOpenAPIGeneratorPlugin {
             tool: tool,
             sourceFiles: sourceFiles,
             targetName: targetName,
-            invocationSource: .CommandPlugin
+            pluginSource: .command
         )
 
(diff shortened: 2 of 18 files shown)


Plugins/OpenAPIGeneratorCommandPlugin/plugin.swift
```diff
@@ -0,0 +1,161 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import PackagePlugin
+import Foundation
+
+@main
+struct SwiftOpenAPIGeneratorPlugin {
+    enum Error: Swift.Error, CustomStringConvertible, LocalizedError {
```
Kyle-Ye, line 19: This logic is duplicated. Maybe we could move such code to a single file. And make soft link to it so that both plugins can depend on it.

MahdiBM: Maybe for the `error` enum ... but the rest of the logic isn't duplicated enough for me to want to make it reusable. It has some subtle differences.

Kyle-Ye: Yes. I mean the `Error` enum.

MahdiBM: i guess i could make half of the `runCommand` function + the errors, reusable ... it will have the benefit of not needing to change 2 different places if someone wants to modify the code, but the logic is small enough that i'm not sure if it's worth going through the trouble.

Kyle-Ye: Both is fine for me. I think if later we have some diff on the implementation. I can add that part back to the 2 plugin target at any time. So I prefer to "make half of the runCommand function + the errors, reusable"

MahdiBM: Github doesn't automatically refresh the comments when someone responds, so i missed your second comment ....
I have no problems doing that ... but again not sure if it's worth it ... I'll think more about it, or if maintainers think it's a good idea, then i'll just do it.

MahdiBM: Ahh this is getting annoying by Github 😅 now i missed your third comment.
But thanks for your input :\)

MahdiBM: I applied your suggestion and i think it looks clean 🙂

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,45 @@
...
+    case incompatibleTarget(targetName: String)
+    case badArguments(arguments: [String])
+    // The description is only suitable for Xcode, as it's only thrown in Xcode plugins.
+    case noTargetsMatchingTargetName(targetName: String)
+    // The description is not suitable for Xcode, as it's not thrown in Xcode plugins.
+    case tooManyTargetsMatchingTargetName(targetNames: [String])
+    case noConfigFound(targetName: String)
+    case noDocumentFound(targetName: String)
+    case multiConfigFound(targetName: String, files: [Path])
+    case multiDocumentFound(targetName: String, files: [Path])
```
denil-ct, line 14:
```suggestion
    case multipleConfigsFound(targetName: String, files: [Path])
    case multipleDocumentsFound(targetName: String, files: [Path])
```
A bit clearer on the meaning.

MahdiBM: I agree but I'm reluctant to change that without an approval from the maintainers, because that's their code. I only moved it to another file.

denil-ct: No worries, just pointed it out.

simonjbeaumont (commented):
@MahdiBM thanks for starting this thread of work!

> Feel free to let me know what you think although the PR is in the Draft mode.

I'll honour the draft status of the PR and refrain from code review, but I can offer some high-level thoughts at this point.

First, I think this change is probably valuable. We already acknowledge in our docs that some folks really do need ahead-of-time generation as part of their workflow, and document how to compile and run the CLI directly. We haven't offered the command plugin to date because we wanted to strongly encourage the use of build-time generation, but we probably shouldn't let our idealism stand in the way of good ergonomics for those that have a genuine need.

In terms of implementation, I'm wondering if we should prefer to factor out the common bits into an internal target, rather than symlinking the common code. I see we already have a symlink for the `GeneratorMode.swift` so maybe this is the only way for these to coexist, or maybe it was a shortcut because we hadn't reached the threshold for creating another target.

Additionally, it appears some parts of #99 have crept into this PR, probably because you're working on them concurrently. When you're ready for a review, please take a moment to keep the PR to one change.

MahdiBM:
> I'm wondering if we should prefer to factor out the common bits into an internal target

That's a known limitation of plugins. They can't depend on non-executable targets, and i tried again too:

```
plugin 'OpenAPIGeneratorCommandPlugin' cannot depend on 'PluginsShared' of type 'library'; this dependency is unsupported
```

> it appears some parts of https://github.com/apple/swift-openapi-generator/pull/99 have crept into this PR

I do try to not mix PRs together, but that was intentional. Notice how i didn't mix #96 into this, as i think that PR might be up for debate. I can of course still revert the #99 changes if it is needed by the time this is ready for review.

The reason was that i thought there is a low chance for #99 for rejection (as you've now approved it 🙂), and i also thought i can save myself some trouble by just merging that PR into this, since we're using this PR in production.

simonjbeaumont:
> > I'm wondering if we should prefer to factor out the common bits into an internal target
> 
> That's a known limitation of plugins. They can't depend on non-executable targets, and i tried again too:
> 
> ```
> plugin 'OpenAPIGeneratorCommandPlugin' cannot depend on 'PluginsShared' of type 'library'; this dependency is unsupported
> ```

Of course—it's all coming back to me now 😅 

> > it appears some parts of #99 have crept into this PR
> 
> I do try to not mix PRs together, but that was intentional. Notice how i didn't mix #96 into this, as i think that PR might be up for debate. I can of course still revert the #99 changes if it is needed by the time this is ready for review.
> 
> The reason was that i thought there is a low chance for #99 for rejection (as you've now approved it 🙂), and i also thought i can save myself some trouble by just merging that PR into this, since we're using this PR in production.

Makes total sense. And yes, I expect #99 will land before this PR so will likely become a moot point.

MahdiBM:
> I'm just getting back from vacation so apologies for the delay in taking a look at this. When you're ready for a review proper, let us know.

No worries. Hopefully i didn't interrupt your vacation 😅

> Could you elaborate here? I assume the original intention is that it would run once per target for which it found an OpenAPI document and config file. In this case, was your concern it exiting on the first error would lead to a game of whack-an-error for users?

Basically yes ... i haven't actually tried implementing it since it didn't work out in my mind. I can try if you think it might be worth it. It was kind of a hard decision that i'm still not sure about.
I can't even properly remember what situations i was thinking about when i decided it might not be worth it... I guess i'll give it an actual try and see what happens ...

Package.swift
```diff
@@ -27,6 +27,7 @@ let package = Package(
     products: [
         .executable(name: "swift-openapi-generator", targets: ["swift-openapi-generator"]),
         .plugin(name: "OpenAPIGenerator", targets: ["OpenAPIGenerator"]),
+        .plugin(name: "OpenAPIGeneratorCommandPlugin", targets: ["OpenAPIGeneratorCommandPlugin"]),
```
czechboy0, line 30:
```suggestion
        .plugin(name: "OpenAPIGeneratorCommand", targets: ["OpenAPIGeneratorCommand"]),
```

I think `OpenAPIGeneratorCommand` is a good name. `Plugin` is redundant, as the build system ensures that this product type is used in the correct context. It also reads better when it shows up in Xcode, as you're "invoking the OpenAPI generator command", and a command is a type of a plugin, so no need to use the phrase `...CommandPlugin`.

Package.swift
```diff
@@ -134,5 +135,19 @@ let package = Package(
                 "swift-openapi-generator",
             ]
         ),
+
+        .plugin(
+            name: "OpenAPIGeneratorCommandPlugin",
+            capability: Target.PluginCapability.command(
+                intent: .custom(
+                    verb: "generate-openapi-code",
+                    description: "Generates OpenAPI code"
```
czechboy0, line 144:
```suggestion
                    description: "Generates Swift code from an OpenAPI document."
```

Package.swift
```diff
@@ -134,5 +135,19 @@ let package = Package(
...
+                ),
+                permissions: [.writeToPackageDirectory(reason: "To generate OpenAPI code")]
```
czechboy0, line 146:
```suggestion
                permissions: [.writeToPackageDirectory(reason: "To write the generated Swift files back into the source directory of the package.")]
```

Package.swift
MahdiBM: I agree, but FYI, ironically there is a diagnosis bug that makes it so you can still declare an extension in the "dependencies" array of targets: https://github.com/apple/swift-package-manager/issues/6676
Luckily for command plugins, you don't even need to explicitly declare the plugin target as a dependency.

czechboy0:
Hi @MahdiBM – I'm happy to see this change, added a few high level comments, but like @simonjbeaumont, I'll respect your the Draft status of the PR and will not dive into detailed code review. I'll say, #99 _might_ need a bit more work, so I think it's best if you remove those changes from this PR.

I do think we just need to use symlinks to share code between the two plugins, and for the code that doesn't need to also be used by the CLI, you can just add it as a separate file in one of the plugins, and symlink it to the other.

So, like:

```swift
Plugins
    OpenAPIGenerator
        plugin.swift
        GeneratorMode.swift -> ../../Sources/_OpenAPIGeneratorCore/GeneratorMode.swift
        PluginUtils.swift
    OpenAPIGeneratorCommand
        plugin.swift
        GeneratorMode.swift -> ../../Sources/_OpenAPIGeneratorCore/GeneratorMode.swift
        PluginUtils.swift -> ../OpenAPIGenerator/PluginUtils.swift
Sources
    _OpenAPIGeneratorCore
        GeneratorMode.swift
```

Package.swift
czechboy0: Understood. I think it's okay to skate to where the puck will be once known-and-reported issues in the tools are fixed 🙂

MahdiBM:
So i finally managed to settle on the general form of how this is going to work (comments, docs etc... are not done yet)
Some notes:
* Removed the ability to ask the generator to only run on 1 target. Because:
  * In short it'll be buggy.
  * To explain:
    * Xcode always passes arguments of form ["--target", "TARGET1", "--target", "TARGET2"] to the command plugin.
    * For some reason, this triggers the `CommandPlugin` instead of the `XcodeCommandPlugin`. Maybe my expectations are off.
    * The passed targets are not all targets. Xcode won't pass some of the targets sometimes, maybe due to scheme settings, which means if you rely on the arguments, you might miss the actual target that needs OpenAPI generation, as was the case with my tests on [Penny](https://github.com/vapor/penny-bot).
    * So this means we need to check dependencies of targets too, but at that point we're not trusting the arguments anyway.
    * Before, the code was to act on 1 specified target if and only if arguments include 1 target, but that can also be buggy. Just imagine if there are 5 targets and Xcode decides to only pass 1 of the targets.
* Changed the error handling part to first check for existence of both config and document files, and throw only after that.
  * The benefit is that the command plugin can better identify if a target does not have any OpenAPI files and so it might not be supposed to have generated OpenAPI stuff at all, or if the target is supposed to have OpenAPI generation and but one of the files isn't available.
  * Generally so far i've handled the errors in a complicated way, and that's only because i want the thrown user-facing errors to be as accurate as possible.

MahdiBM:
@simonjbeaumont @czechboy0 

The PR is now ready for review. If it seems acceptable, i can work on the docs part too.

From @simonjbeaumont:
> For what we can unit test, great. Otherwise, we could probably add a CI that runs the ahead-of-time generation workflow and diffs against the reference code. We'll have to think about whether it's worth it. WDYT, @czechboy0?

I think we'll need that?

From @czechboy0:
> I do think we just need to use symlinks to share code between the two plugins, and for the code that doesn't need to also be used by the CLI, you can just add it as a separate file in one of the plugins, and symlink it to the other.

I'll do that if you insist, but it looks cleaner to me this way because there are 4 files (`GeneratorMode`, `InvocationSource`, `PluginError`, `PluginUtils`) and 3 of them are required in both plugins. `GeneratorMode` is only required for the BuildTool plugin, but i still moved it to the `PluginsShared` directory to make it look cleaner.
The `PluginsShared` directory is only used by plugins.


Also i renamed the directory from `GeneratedSources` to `OpenAPISources` to basically eliminate chances of name collisions with another plugin in the command-plugin considering both `GeneratedSources` and names like `Types`/`Client`/`Server` are common.

Please feel free to make naming suggestions, comment corrections etc...

MahdiBM:
I also have paid attention to the soundness CI. I'm just not sure what-to-do/how-to-do with the remainder of the warnings.
* It emits some warnings for code that wasn't even edited by me. For example it expects me to not use long lines for error descriptions, which is weird. I can use multiline strings, but that doesn't sound right.
* It says `/code/Package.swift:169:10: warning: [TrailingComma] add trailing comma to the last element in multiline collection literal` while the whole `Package.swift` file is only 168 lines.

Plugins/OpenAPIGeneratorCommand/plugin.swift
```diff
@@ -0,0 +1,137 @@
...
+            tool: tool,
+            sourceFiles: sourceFiles,
+            targetName: targetName,
+            invocationSource: .CommandPlugin
+        )
+
+        let toolUrl = URL(fileURLWithPath: inputs.tool.path.string)
+        let process = Process()
+        process.executableURL = toolUrl
+        process.arguments = inputs.arguments
```
czechboy0, line 36: Can you also clean the environment, to ensure that nothing leaks through. If that breaks the plugin, only add the exact env vars required.
```suggestion
        process.arguments = inputs.arguments
        process.environment = [:]
```

Plugins/OpenAPIGeneratorCommand/plugin.swift
```diff
@@ -0,0 +1,137 @@
...
+        try process.run()
```
czechboy0, line 37: Should we have a `process.waitUntilExit()` here? Otherwise the command plugin might exit before the spawned process does.

Plugins/OpenAPIGeneratorCommand/plugin.swift
```diff
@@ -0,0 +1,137 @@
...
+}
+
+extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
+    func performCommand(
+        context: PluginContext,
+        arguments: [String]
+    ) async throws {
+        var hasHadASuccessfulRun = false
+        var errors = [(error: any Error, targetName: String)]()
+        for target in context.package.targets {
```
czechboy0, line 48: I think we can simplify the whole plugin by _always_ requiring exactly one target name. That way, we don't have to search for dependencies or worry about which targets are included in a scheme.

It also means we can emit good errors in all scenarios, instead of guessing which target the adopter meant to generate code into.

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,82 @@
...
+
+enum PluginError: Swift.Error, CustomStringConvertible, LocalizedError {
+    case incompatibleTarget(targetName: String)
+    case noTargetsFoundForCommandPlugin
+    case fileErrors([FileError], targetName: String)
+
+    var description: String {
+        switch self {
+        case .incompatibleTarget(let targetName):
+            return "Incompatible target called '\(targetName)'. Only Swift source targets can be used with the Swift OpenAPI generator plugin."
```
czechboy0, line 12:
```suggestion
            return "Incompatible target called '\(targetName)'. Only Swift source targets can be used with the Swift OpenAPI Generator plugin."
```

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,82 @@
...
+        case .noTargetsFoundForCommandPlugin:
+            return "None of the targets include valid OpenAPI spec files. Please make sure at least one of your targets has any valid OpenAPI spec files before triggering this command plugin. See documentation for details."
```
czechboy0, line 14: I guess this case will go away if we simplify to always require the specific target to be passed in.

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,82 @@
...
+struct FileError: Swift.Error, CustomStringConvertible, LocalizedError {
+
+    enum Kind: CaseIterable {
+        case config
+        case document
+    }
+
+    enum Issue {
+        case notFound
+        case multiFound(files: [Path])
```
czechboy0, line 40:
```suggestion
        case multipleFilesFound(files: [Path])
```

While we're changing this, we can make it clearer.

Sources/swift-openapi-generator/GenerateCommand.swift
```diff
@@ -42,16 +42,19 @@ struct _GenerateCommand: AsyncParsableCommand {
     )
     var outputDirectory: URL = URL(fileURLWithPath: FileManager.default.currentDirectoryPath)
 
-    @Flag(
+    @Option(
         help:
             "Whether this invocation is from the SwiftPM plugin. We always need to produce all files when invoked from the plugin. Non-requested modes produce empty files."
     )
-    var isPluginInvocation: Bool = false
+    var invokedFrom: InvocationSource = .CLI
```
czechboy0, line 49: Allow me to bikeshed for a moment here: I think the name `pluginInvocationSource` would work well, as an option (so `PluginInvocationSource?`).

Then, the `PluginInvocationSource` would only have the cases `.command` and `.build`.

That way, by default, and when invoked manually, this option is nil. Otherwise, each plugin provides its own type.

Sources/swift-openapi-generator/InvocationSource.swift
```diff
@@ -0,0 +1,6 @@
+/// The source of the generator invocation.
+enum InvocationSource: String, Codable {
+    case BuildToolPlugin
+    case CommandPlugin
+    case CLI
```
czechboy0, line 5:
```suggestion
    case build
    case command
```

Plugins/PluginsShared/PluginUtils.swift
```diff
@@ -0,0 +1,118 @@
...
+    /// Validates the inputs and returns the necessary values to run a plugin.
+    static func validateInputs(
+        workingDirectory: Path,
+        tool: (String) throws -> PluginContext.Tool,
+        sourceFiles: FileList,
+        targetName: String,
+        invocationSource: InvocationSource
+    ) throws -> ValidatedInputs {
+        let (config, doc) = try findFiles(inputFiles: sourceFiles, targetName: targetName)
+        let genSourcesDir = workingDirectory.appending("OpenAPISources")
```
czechboy0, line 25: I'd keep the name to the original name, that way when folks upgrade to the newer generator, they won't immediately get a build error as we'd generate duplicate files to the new directory name, and not clean up the old one. `GeneratedSources` is explicit enough, IMO.

czechboy0 (changes requested):
Thanks @MahdiBM for moving the ball forward, added some suggestions on how to simplify the plugin, I don't think we should try to be too smart – let's require that adopters provide the exact target name during invocation of the command.

Plugins/PluginsShared/PluginUtils.swift
MahdiBM: > that way when folks upgrade to the newer generator, they won't immediately get a build error

Right. What i was thinking was that if it might actually cause problems in the future, it's better to pay the price now rather than later.
I can't say i really disagree with you on "`GeneratedSources` is explicit enough, IMO.". It's your responsibility so if you think it's good enough, i'll change it back :)

Sources/swift-openapi-generator/GenerateCommand.swift
MahdiBM: Changed the name to `PluginSource` as i think considering the requested changes, "invocation" was a little bit extra.

Plugins/PluginsShared/PluginError.swift
MahdiBM: yeah it should

MahdiBM:
@czechboy0  @simonjbeaumont   Take 3!

As "bullet points" since that's easier to write for me:
* I tried to make the Command plugin work with only one target, but who knew there would be even more problems!
* The main problem this time around was that i noticed Xcode's command plugin run UI doesn't include all targets to pass to a command plugin. 
  * It only includes products.
* So this means if we have a OpenAPI-generated target which is not e.g. an executable target and is only used by some other executable targets, we will not be able to choose the OpenAPI-generated target from Xcode's UI.
* To solve this problem, what i did was to find local dependencies of a package and run the generator on both the main target and its local dependencies.
  * Luckily the normal `CommandPlugin` protocol does provide enough context to be able to accurately find a target's local dependencies.
  * But the `XcodeCommandPlugin` protocol does not provide that info, so i had to get rid of it.
    * To be honest, as i mentioned before, i'm not sure if it used to do anything in the first place.
    * Even Xcode seemed to run the `CommandPlugin`'s `performCommand()` function instead of `XcodeCommandPlugin`, not 100% sure why.
* The problem in all this, is that similar to what i tried before with running the command plugin automatically on all targets, it makes the error handling worse than if we could get a singular correct/exact target name from users.
* Though as always, i've tried to throw proper useful errors to users so i don't think this is a deal breaker.

If you wanted my opinion, i'd say yeah it's suboptimal that Xcode UI doesn't allow passing of any target names you want through the UI itself, but i still think the command plugin is still good enough to be worth it.

You may also ask why not just go back to doing it for all targets of the package at least?
And to answer, i'm not 100% sure but i'm fine with this small level of explicitly choosing a product from Xcode's UI although it's one more click, considering allows us to throw slightly more meaningful/accurate errors. Again, not sure, i can set it back to just run on all of a package's targets at once if you think that's better.

czechboy0:
> The main problem this time around was that i noticed Xcode's command plugin run UI doesn't include all targets to pass to a command plugin.

Well, that's unfortunate. We have at least these two options:
- live with that limitation, and ask adopters to wrap any target with OpenAPI documents in a product
- do what you did, and crawl the local dependency graph and try to run it on all of the targets with the right files

I'm on the fence, but at this point leaning towards what you did, and please file a FB on Xcode to support any targets, not just products, and we'll link it here to ensure that we can clean up the behavior in the future once Xcode improves that behavior.

Also curious which direction @simonjbeaumont is leaning.

Ok, let me re-review the code now 🙂

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,125 @@
...
+    case noTargetsMatchingTargetName(targetName: String)
+    case tooManyTargetsMatchingTargetName(targetName: String, matchingTargetNames: [String])
+    case fileErrors([FileError])
+
+    var description: String {
+        switch self {
+        case .incompatibleTarget(let targetName):
+            return "Incompatible target called '\(targetName)'. Only Swift source targets can be used with the Swift OpenAPI Generator plugin."
+        case .noTargetOrDependenciesWithExpectedFiles(let targetName, let dependencyNames):
+            let introduction = dependencyNames.isEmpty ? "Target called '\(targetName)' doesn't contain" : "Target called '\(targetName)' or its local dependencies \(dependencyNames) don't contain"
```
czechboy0, line 17:
```suggestion
            let introduction = dependencyNames.isEmpty ? "Target called '\(targetName)' doesn't contain" : "Target called '\(targetName)' or its local dependencies \(dependencyNames.joined(separator: ", ") don't contain"
```

Should we format the names somehow?

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,125 @@
+import PackagePlugin
+import Foundation
+
+enum PluginError: Swift.Error, CustomStringConvertible, LocalizedError {
+    case incompatibleTarget(targetName: String)
```
czechboy0, line 5:
```suggestion
    case incompatibleTarget(name: String)
```

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,125 @@
...
+    case noTargetOrDependenciesWithExpectedFiles(targetName: String, dependencyNames: [String])
+    case badArguments([String])
+    case noTargetsMatchingTargetName(targetName: String)
```
czechboy0, line 8:
```suggestion
    case noTargetsMatchingTargetName(String)
```

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,125 @@
...
+    case tooManyTargetsMatchingTargetName(targetName: String, matchingTargetNames: [String])
+    case fileErrors([FileError])
+
+    var description: String {
+        switch self {
+        case .incompatibleTarget(let targetName):
+            return "Incompatible target called '\(targetName)'. Only Swift source targets can be used with the Swift OpenAPI Generator plugin."
+        case .noTargetOrDependenciesWithExpectedFiles(let targetName, let dependencyNames):
+            let introduction = dependencyNames.isEmpty ? "Target called '\(targetName)' doesn't contain" : "Target called '\(targetName)' or its local dependencies \(dependencyNames) don't contain"
+            return "\(introduction) any config or document files with expected names. For OpenAPI code generation, a target needs to contain a config file named 'openapi-generator-config.yaml' or 'openapi-generator-config.yml', as well as an OpenAPI document named 'openapi.yaml', 'openapi.yml' or 'openapi.json' under target's source directory. See documentation for details."
```
czechboy0, line 18:
```suggestion
            return "\(introduction) any config or OpenAPI document files with expected names. See documentation for details."
```

Let's add "OpenAPI" to the sentence and drop the summary of the docs, we don't want to have to keep it in sync here with the other places. Sending people to docs is good enough.

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,125 @@
...
+        case .badArguments(let arguments):
+            return "On Xcode, use Xcode's command plugin UI to choose one specific target before hitting 'Run'. On CLI make sure arguments are exactly of form '--target <target-name>'. The reason for this error is unexpected arguments: \(arguments)"
```
czechboy0, line 20: Adopters might be using this from other IDEs, so I'd drop the reference to Xcode specifically.
```suggestion
            return "Unexpected arguments: '\(arguments.joined(separator: " "))', expected: '--target MyTarget'."
```

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,125 @@
...
+        case .noTargetsMatchingTargetName(let targetName):
+            return "Found no targets matching target name '\(targetName)'. Please make sure the target name argument leads to one and only one target."
```
czechboy0, line 22:
```suggestion
            return "Found no targets with the name '\(targetName)'."
```

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,125 @@
...
+        case .tooManyTargetsMatchingTargetName(let targetName, let matchingTargetNames):
+            return "Found too many targets matching target name '\(targetName)': \(matchingTargetNames). Please make sure the target name argument leads to a unique target."
```
czechboy0, line 24:
```suggestion
            return "Found too many targets with the name '\(targetName)': \(matchingTargetNames.joined(separator: ", ")). Select a target name with a unique name."
```

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,125 @@
...
+        case .fileErrors(let errors):
+            return "Found file errors: \(errors)."
```
czechboy0, line 26:
```suggestion
            return "Issues with required files: \(errors.map(\.localizedDescription).joined(separator: ", and"))."
```

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,125 @@
...
+        switch self {
+        case .incompatibleTarget:
+            return false
+        case .noTargetOrDependenciesWithExpectedFiles:
+            return false
+        case .badArguments:
+            return false
+        case .noTargetsMatchingTargetName:
+            return false
+        case .tooManyTargetsMatchingTargetName:
```
czechboy0, line 45: Nit: You could put these first 5 on the same line, to more clearly communicate that they are handled the same way.

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,125 @@
...
+            return "Found file errors: \(errors)."
+        }
+    }
+
+    var errorDescription: String? {
+        description
+    }
+
+    /// The error is definitely due to misconfiguration of a target.
+    var isDefiniteMisconfigurationError: Bool {
```
czechboy0, line 35:
```suggestion
    var isMisconfigurationError: Bool {
```

You can drop the "Definite", as `isMisconfigurationError` already communicates what you need.

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,125 @@
...
+        case config
+        case document
+    }
+
+    enum Issue {
+        case noFilesFound
+        case multipleFilesFound(files: [Path])
+
+        /// The error is definitely due to misconfiguration of a target.
+        var isDefiniteMisconfigurationError: Bool {
```
czechboy0, line 65:
```suggestion
        var isMisconfigurationError: Bool {
```

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,125 @@
...
+            }
+        }
+    }
+
+    let targetName: String
+    let fileKind: Kind
+    let issue: Issue
+
+    var description: String {
+        "FileError { \(helpAnchor!) }"
```
czechboy0, line 80: Let's reshuffle this a bit to get rid of the force unwrap.

1. Move the contents of `helpAnchor` to be the implementation of `description`.
2. Have `helpAnchor` call `description`.
3. Drop the `FileError { ... }` part completely, unless it's specifically parsed some other tool?

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,125 @@
...
+                return "Multiple OpenAPI documents found in the target named '\(targetName)', but exactly one is expected. Found \(files.map(\.description).joined(separator: " "))."
+            }
+        }
+    }
+
+    var errorDescription: String? {
+        description
+    }
+
+    static func notFoundFileErrors(forTarget targetName: String) -> [FileError] {
```
czechboy0, line 106: Is anything calling this helper function?

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,125 @@
...
+        FileError.Kind.allCases.map { kind in
+            FileError(targetName: targetName, fileKind: kind, issue: .noFilesFound)
+        }
+    }
+}
+
+private extension [FileError] {
```
czechboy0, line 113:
```suggestion
private extension Array where Element == FileError {
```

Nit: This is the spelling we use in this project.

Plugins/PluginsShared/PluginError.swift
```diff
@@ -0,0 +1,125 @@
...
+    /// The error is definitely due to misconfiguration of a target.
+    var isDefiniteMisconfigurationError: Bool {
```
czechboy0, line 115:
```suggestion
    var isMisconfigurationError: Bool {
```

Sources/swift-openapi-generator/PluginSource.swift
```diff
@@ -0,0 +1,5 @@
+/// The source of a plugin generator invocation.
+enum PluginSource: String, Codable {
+    case build
+    case command
```
czechboy0, line 4: Please add doc comments for the cases.

Sources/swift-openapi-generator/runGenerator.swift
```diff
@@ -94,18 +105,31 @@ extension _Tool {
     /// - Throws: When writing to disk fails.
     /// - Returns: `true` if the generated contents changed, otherwise `false`.
     @discardableResult
-    static func replaceFileContents(at path: URL, with contents: () throws -> Data) throws -> Bool {
+    static func replaceFileContents(
+        inDirectory outputDirectory: URL,
+        fileName: String,
+        with contents: () throws -> Data
+    ) throws -> Bool {
+        let fm = FileManager.default
```
czechboy0, line 113:
```suggestion
        let fileManager = FileManager()
```

Variable naming, and it's better to create a new FileManager, as otherwise someone might have set the delegate on the shared file manager, which can lead to unexpected results.

Sources/swift-openapi-generator/GenerateCommand.swift
```diff
@@ -42,16 +42,19 @@ struct _GenerateCommand: AsyncParsableCommand {
...
     var outputDirectory: URL = URL(fileURLWithPath: FileManager.default.currentDirectoryPath)
 
-    @Flag(
+    @Option(
         help:
-            "Whether this invocation is from the SwiftPM plugin. We always need to produce all files when invoked from the plugin. Non-requested modes produce empty files."
+            "Source of invocation if by a plugin. The generator needs to produce all files when invoked as a build plugin, so non-requested modes produce empty files."
     )
-    var isPluginInvocation: Bool = false
+    var invokedFrom: PluginSource?
```
czechboy0, line 49: Just bumping the first half of this comment: https://github.com/apple/swift-openapi-generator/pull/98#discussion_r1265064137

Let's also rename the option (will require a change in the plugins as well, and in the docs) to `--plugin-source`, I'm not a fan of options ending with a preposition.

```suggestion
    var pluginSource: PluginSource?
```

Plugins/PluginsShared/PluginUtils.swift
```diff
@@ -0,0 +1,118 @@
...
+        pluginSource: PluginSource
+    ) throws -> ValidatedInputs {
+        let (config, doc) = try findFiles(inputFiles: sourceFiles, targetName: targetName)
+        let genSourcesDir = workingDirectory.appending("GeneratedSources")
+
+        let arguments = [
+            "generate", "\(doc)",
+            "--config", "\(config)",
+            "--output-directory", "\(genSourcesDir)",
+            "--invoked-from", "\(pluginSource.rawValue)",
```
czechboy0, line 31:
```suggestion
            "--plugin-source", "\(pluginSource.rawValue)",
```

Plugins/OpenAPIGeneratorCommand/plugin.swift
```diff
@@ -0,0 +1,103 @@
...
+            pluginSource: .command
+        )
+
+        let toolUrl = URL(fileURLWithPath: inputs.tool.path.string)
+        let process = Process()
+        process.executableURL = toolUrl
+        process.arguments = inputs.arguments
+        process.environment = [:]
+        try process.run()
+        process.waitUntilExit()
```
czechboy0, line 39: Also, we should inspect the `process.terminationStatus` and throw an error if it's not 0.

MahdiBM:
> live with that limitation, and ask adopters to wrap any target with OpenAPI documents in a product

Honestly that idea did not occur to me. It does sound like a good idea.
I just tried adding the generator-related target to the `products` array of the package and it worked pretty fine.

Plugins/OpenAPIGeneratorCommand/plugin.swift
```diff
@@ -0,0 +1,103 @@
...
+                        tool: context.tool,
+                        sourceFiles: swiftTarget.sourceFiles,
+                        targetName: target.name
+                    )
+                    hadASuccessfulRun = true
+                } catch let error as PluginError {
+                    if error.isDefiniteMisconfigurationError {
+                        throw error
+                    }
+                }
```
czechboy0, line 87: Please add a simple `print(...)` describing which target is being inspected, and whether we tried to run the command on it, and if it succeeded or not. Just for easier debugging.

czechboy0 (changes requested):
A bunch of nits, but overall I think this direction is the right one. Thanks! 🙏

czechboy0:
> > live with that limitation, and ask adopters to wrap any target with OpenAPI documents in a product
> 
> Honestly that idea did not occur to me. It does sound like a good idea. I just tried adding the generator-related target to the `products` array of the package. It worked seamlessly and made Xcode show the target to choose for the command plugin.

Right, the question is whether this is a reasonable thing to ask people, to basically publicize any target where they use the plugin. Maybe that goes against our [documented recommendations](https://swiftpackageindex.com/apple/swift-openapi-generator/0.1.5/documentation/swift-openapi-generator/api-stability-of-generated-code#Create-a-curated-client-library-package). Idk, @simonjbeaumont what do you think?

Package.swift
```diff
@@ -151,5 +152,20 @@ let package = Package(
...
             ]
         ),
+
+        // Command Plugin
+        .plugin(
+            name: "OpenAPIGeneratorCommand",
+            capability: Target.PluginCapability.command(
+                intent: .custom(
+                    verb: "generate-openapi-code",
+                    description: "Generates Swift code from an OpenAPI document."
```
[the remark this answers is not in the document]
MahdiBM: I see swift format doing the same so i guess we can keep the theme although i prefer the existing sentence.

Plugins/PluginsShared/PluginError.swift
MahdiBM: I did want to format dependencies, but preferred to skip it since it would have been a tiny bit complicated.
I'll see what i can do, probably just the same thing as `.joined(separator: ", ")` but with a "or" as the last separator to make it sound more human.

Plugins/PluginsShared/PluginError.swift
MahdiBM: Generally agree with all these enum case associated value renames... I didn't do them myself just to keep the looks of what the project already had :)

Plugins/PluginsShared/PluginError.swift
MahdiBM: The soundness CI wasn't happy when i did that 😅

Plugins/PluginsShared/PluginError.swift
MahdiBM: not anymore!

Plugins/OpenAPIGeneratorCommand/plugin.swift
MahdiBM: Right i was thinking about that ... i wonder if we can report the error too somehow ...

Plugins/OpenAPIGeneratorCommand/plugin.swift
```diff
@@ -0,0 +1,103 @@
...
+    }
+}
+
+extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
+    func performCommand(
+        context: PluginContext,
+        arguments: [String]
+    ) async throws {
+        guard arguments.count == 2,
+              arguments[0] == "--target"
```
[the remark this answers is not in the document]
MahdiBM: Won't work with Xcode unfortunately

Plugins/OpenAPIGeneratorCommand/plugin.swift
MahdiBM: Sorry i'm not sure i understand the point 🤔
If you want the command plugin to mention the targets that the command plugin will try to run on, should i add it higher in the function where we calculate the local dependencies?
If you mean we should print the error when we don't throw it, i'm not sure if that's a good idea since there will a lot of false positives... I could add a print expression in at the end of the function to mention it tried to run the plugin on <> targets and <> targets seemed to have generator-related files.

MahdiBM:
One thing that i forgot ... should i make the command-plugin-invocation delete the build-plugin-invocation generated files?
So users don't need to manually clean-build or modify DerivedData.

The build plugin won't be able to do the same though, considering it doesn't have write access to the source code.

Plugins/PluginsShared/PluginError.swift
czechboy0: Nice, yeet that code then 🙂

Plugins/PluginsShared/PluginError.swift
MahdiBM: Considering the errors generally don't have `helpAnchor` at all, should i just remove `helpAnchor`?
It doesn't really do anything. I just put the error description in there because it didn't hurt to have `helpAnchor` i guess.

Plugins/OpenAPIGeneratorCommand/plugin.swift
czechboy0: I meant that we'd see something like this in the log:

```
- Target Foo1 has no OpenAPI generator files
- Target Foo2 has no OpenAPI generator files
- Target Foo3 has OpenAPI files, successfully generated code
- Target Foo4 has OpenAPI files, fail to run with error: ...
```

Plugins/OpenAPIGeneratorCommand/plugin.swift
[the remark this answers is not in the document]
czechboy0: I meant that the plugin would accept _either_ syntax:

Xcode would provide:
```
--target Foo
```

CLI could provide:
```
Foo
```

And the plugin would be able to handle either.

Sources/swift-openapi-generator/runGenerator.swift
MahdiBM: This would be inconsistent with the rest of the project.
Should i still go for it?
I'll rename the variable for now.

czechboy0:
> One thing that i forgot ... should i make the command-plugin-invocation delete the build-plugin-invocation generated files? So users don't need to manually clean-build or modify DerivedData.
> 
> The build plugin won't be able to do the same though, considering it doesn't have write access to the source code.

I don't think so, they each produce files into different locations. The build plugin produces them in GeneratedSources in Derived Data, the command plugin writes them back to the package source directory. I wouldn't want them to try to delete each other's files, seems it could go wrong very badly very quickly, even if the sandbox allowed it (I don't think it would).

Plugins/PluginsShared/PluginError.swift
czechboy0: Yeah remove it. Make sure the error has `description` and `errorDescription` (the latter calling the former), but no need for `helpAnchor`, I've also never used it.

simonjbeaumont:
Catching up a bit here...

> > > live with that limitation, and ask adopters to wrap any target with OpenAPI documents in a product
> > 
> > 
> > Honestly that idea did not occur to me. It does sound like a good idea. I just tried adding the generator-related target to the `products` array of the package. It worked seamlessly and made Xcode show the target to choose for the command plugin.
> 
> Right, the question is whether this is a reasonable thing to ask people, to basically publicize any target where they use the plugin. Maybe that goes against our [documented recommendations](https://swiftpackageindex.com/apple/swift-openapi-generator/0.1.5/documentation/swift-openapi-generator/api-stability-of-generated-code#Create-a-curated-client-library-package). Idk, @simonjbeaumont what do you think?

Just confirming my understanding here: are we saying that it works as expected with `swift package ...` but with Xcode there's some issues, that require this workaround?

In that event I don't think we should hold back the utility of having a supported workflow for Swift PM users.

We then are free to choose how to approach the Xcode flow, in light of any limitations it might have:
1. Actively disable it in code.
2. Let people use it but call out it's not supported in docs.
3. Actively support it and call out how to use it in docs, with all the limitations, and consequences of any workarounds.

> One thing that i forgot ... should i make the command-plugin-invocation delete the build-plugin-invocation generated files? So users don't need to manually clean-build or modify DerivedData.
> 
> The build plugin won't be able to do the same though, considering it doesn't have write access to the source code.

I think I'd be OK with it removing things from DerivedData. But I don't feel strongly about it. What are the other options?
- Let the build fail?
- Emit warnings?
- Abort and tell the user to clean?

Aside: Does the build plugin have _read_ access to the source tree? If it did, it might be able to provide sensible errors/warnings, even if it cannot delete things automatically.

Plugins/OpenAPIGeneratorCommand/plugin.swift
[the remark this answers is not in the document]
MahdiBM: ah right

MahdiBM:
@czechboy0 one more thing i remembered:
If a project has been using the build plugin and switches to the command plugin, Xcode/Swift would throw the error that says "there are files with same names". Though i think it'll still be fine, because even if we change file names, then Xcode will just throw "duplicate type name" errors instead. So i think the solution is to just point this out in the docs. Possibly have a "Switching between plugins" tutorial.

czechboy0:
> @czechboy0 one more thing i remembered:
> If a project has been using the build plugin and switches to the command plugin, Xcode/Swift would throw the error that says "there are files with same names". Though i think it'll still be fine, because even if we change file names, then Xcode will just throw "duplicate type name" errors instead. So i think the solution is to just point this out in the docs. Possibly have a "Switching between plugins" tutorial.

We'll see if people actually hit it and struggle to figure it out, but I'm fine with us not adding complexity just to catch this one case and try to provide a better error. It seems it'd be too fragile anyway, as it'd rely on a bunch of implementation details of the build system and Xcode, so I'd like to avoid it.

Plugins/PluginsShared/PluginError.swift
MahdiBM: > Xcode UI - unless they explicitly add extra arguments, they'll never see this error, as the Xcode UI will force them to select one product, and provide the right flags.

Not quite right. If they don't choose any target, Xcode will pass _all targets_ automatically in form of ["--target", "TARGET1", "--target", "TARGET2"]. If users press the "Run" button without choosing a product, this error will help them know what to do.
It's also possible for users to multi-select multiple products.

Knowing that, do you still want me to remove the reference to Xcode?

czechboy0: Yes, let's remove the reference.

But the multiple --target case is interesting. Would it be difficult to just iterate them all, now that we already crawl local dependencies anyway? That way, if you don't provide an explicit target, we run on them all?

MahdiBM: Not really, shouldn't be difficult maybe very slightly worse error handling, but not sure, might be worth it. Should i go for it?

czechboy0: Yeah I guess so. If we can't have full script mode, where the user provides exactly one target that we run it on, with great error handling, we might as well go the other direction, and allow running on all targets.

Once we added the dependency crawling, we lost most of the great error handling anyway, so this change doesn't seem as that big of a leap, and the experience for new adopters should be nicer too.

MahdiBM:
@czechboy0 
> please file a FB on Xcode to support any targets, not just products, and we'll link it here to ensure that we can clean up the behavior in the future once Xcode improves that behavior.

FB12674830

Plugins/PluginsShared/PluginError.swift
MahdiBM: I'll give it another try, with all the errors in one statement but in different lines

Plugins/PluginsShared/PluginError.swift
MahdiBM: @czechboy0 himself asked me to change this to "multiple" in a later comment, so i guess i can resolve this conversation now.

Plugins/OpenAPIGeneratorCommand/plugin.swift
MahdiBM: It does show, as it shows the "Swift OpenAPI Generator is running with the following configuration:" logs too.

MahdiBM:
@czechboy0 one more try :) :

#### Not choosing any product from Xcode UI:

Screenshot 2023-07-20 at 10 47 45 AM

#### Choosing 'AutoPingsLambda' which doesn't have OpenAPI files, and its local dependencies too:

Screenshot 2023-07-20 at 10 49 54 AM

#### Choosing the exact product:

Screenshot 2023-07-20 at 10 50 43 AM

simonjbeaumont:
This seems good to me. To summarise:

1. If you don't select any target, it tries them all and "skips" targets that aren't deemed appropriate.
2. If you explicitly select an appropriate target, it succeeds.
3. If you explicitly select an inappropriate target, it fails.

In scenario (1), I can see in the screenshot two things that would be nice to improve:
- The order of the logs: would be nice to log the target being operated on before the rest of the logging. This reads clearer IMO.
- Would be nice for these "skipped" targets to appear as "skipped" or "warnings" rather than "failed" in the logs.

I don't mind deferring that to another PR though. WDYT @czechboy0 @MahdiBM?

MahdiBM:
@simonjbeaumont 
I know this PR has been going for long enough that you probably don't want to bother me anymore, but those changes are small enough, and even if they weren't, i still wouldn't mind them 😅

> The order of the logs: would be nice to log the target being operated on before the rest of the logging. This reads clearer IMO.

Good idea but will basically duplicate the logs because we'll need to say `running on target '----'` then `run failed on target '-----'`. (Unless the target is just straight up not a swift source module in which case no need for duplication.)

Should i go for it? or did you mean something else?

> Would be nice for these "skipped" targets to appear as "skipped" or "warnings" rather than "failed" in the logs.

No objections, but can you mention some examples so i know what you exactly have in mind?
Like just changing the words, or kind of flagging the print statement by starting them with e.g. `Skipped target '----':`, `Warnings for target '-----':`?

MahdiBM:
What i do prefer though, is for the Docs part to be deferred to another PR considering this has taken enough time/length 😅 would prefer this to be merged sooner than later, when possible, so i can also start using the main Apple repo in my `Package.swift`s 🙂

simonjbeaumont:
> @simonjbeaumont I know this PR has been going for long enough that you probably don't want to bother me anymore, but those changes are small enough, and even if they weren't, i still wouldn't mind them 😅

OK, great!

Both of these comments pertain to this logging from your screenshot above, so I'll paste the textual version here before continuing to discuss your specific questions:

```
✅ Invoking plug-in command "OpenAPIGeneratorCommand" 18.5 seconds
- Target 'PennyTests': OpenAPI code generation failed. No expected config or OpenAPI document files.
- Target 'Models': OpenAPI code generation failed. No expected config or OpenAPI document files.
- Target 'Penny': OpenAPI code generation failed. No expected config or OpenAPI document files.
- OpenAPI document path: /Users/mahdibm/Github/penny-bot/Sources/GitHubAP|/openapi.yaml
- Configuration path: /Users/mahdibm/Github/penny-bot/Sources/GitHubAP|/openapi-generator-config.yml
- Output directory: /Users/mahdibm/Github/penny-bot/Sources/GitHubAP|/GeneratedSources
- Current directory: /Users/mahdibm/Github/penny-bot
- Plugin source: command
- Additional imports: <none>
File Types.swift: unchanged
File Client.swift: unchanged
- Target 'GitHubAPI': OpenAPI code generation successfully completed.
- Target 'Extensions': OpenAPI code generation failed. No expected config or OpenAPI document files.
- Target 'GHHooksLambda': OpenAPI code generation failed. No expected config or OpenAPI document files.
- Target 'Fake': OpenAPI code generation failed. No expected config or OpenAPI document files.
- Target 'SponsorsLambda': OpenAPI code generation failed. No expected config or OpenAPI document files.
- Target 'SharedServices': OpenAPI code generation failed. No expected config or OpenAPI document files.
- Target 'CoinsLambda': OpenAPI code generation failed. No expected config or OpenAPI document files.
- Target 'FaqsLambda': OpenAPI code generation failed. No expected config or OpenAPI document files.
- Target 'AutoPingsLambda': OpenAPI code generation failed. No expected config or OpenAPI document files.
✅ Activity Log Complete 7/20/23, 10:50 AM 17.8 seconds
No issues
```

> > The order of the logs: would be nice to log the target being operated on before the rest of the logging. This reads clearer IMO.
> 
> Good idea but will basically duplicate the logs because we'll need to say `running on target '----'` then `run failed on target '-----'`. (Unless the target is just straight up not a swift source module in which case no need for duplication.)
> 
> Should i go for it? or did you mean something else?

Personally, I prefer logging to be before an operation, and, looking through the build logs in Xcode for a target I happen to have open I see this pattern. Examples include `Compiling ...`, and `Discovering Swift tasks after ...`. When an operation is failed, there is some "duplicate" logging; here is a textual version of what I see in the report navigator when I replace `print` with `pint` (note, the emojis aren't actually in the log—they are a representation of how Xcode renders the icons):

```
...
✅ Planning Swift module swift_openapi_generator (arm64)
ℹ️ Compiling runGenerator.swift 
    ℹ️ Compile runGenerator.swift (arm64)
        ❌ Cannot find 'pint' in scope
❌ Build failed ...
```

When the issue if fixed, I see:

```
✅ Planning Swift module swift_openapi_generator (arm64)
✅ Compiling runGenerator.swift 
    ✅ Compile runGenerator.swift (arm64)
...
✅ Build succeeded ...
```

> > Would be nice for these "skipped" targets to appear as "skipped" or "warnings" rather than "failed" in the logs.
> 
> No objections, but can you mention some examples so i know what you exactly have in mind? Like just changing the words, or kind of flagging the print statement by starting them with e.g. `Skipped target '----':`, `Warnings for target '-----':`?

I think it's odd to see big green check marks and lots of `<some operation> failed` in the logs.

I'm not suggesting we go for emojis, but I'd be happy with us finding some terminology that is clear about what's going on, for example:

```
Considering target for OpenAPI generation: InappropriateTarget
Skipped target (no OpenAPI document): InappropriateTarget
Considering target for OpenAPI generation: AppropriateTarget
Running Swift OpenAPI Generator on target AppropriateTarget with the following configuration:
...
```

I'd be happy if it then printed a line explicitly saying whether it was succeeded or failed.

IMHO this small amount of "duplicate" logging is a small price to pay for clarity when debugging a failed build.

MahdiBM:
@simonjbeaumont This is what i came up with.

Please let me know if i can improve any of the sentences.

I used the ✅ emoji for a successful code generation because i think it makes long logs like this more skimmable.

The `- Stopping because target isn't configured for OpenAPI code generation.` logs don't feel satisfying in terms of being different from the theme/mood of the rest of the logs, but that's the best i could come up with.


Screenshot 2023-07-20 at 4 39 04 PM

simonjbeaumont:
> @simonjbeaumont This is what i came up with.
Thanks for going the extra mile on this one.

> I used the ✅ emoji for a successful code generation because i think it makes long logs like this more skimmable.

While I initially hinted at not doing this, I think I agree; it does make it easier to read at a glance so I'm fine with it 👍 

> Please let me know if i can improve any of the sentences.
> 
> The `- Stopping because target isn't configured for OpenAPI code generation.` logs don't feel satisfying in terms of being different from the theme/mood of the rest of the logs, but that's the best i could come up with.

Aside from tone, this one in particular isn't informative—why isn't it "configured" and what does it mean for it to be configured? I this case we could have something like: `- Skipped target: Target does not contain OpenAPI document` or something similar?

Other than that I'm pleased with how this is looking.

@czechboy0 do you have anything to add before we push this over the line?

MahdiBM:
@simonjbeaumont i made it non-informative because i suspected @czechboy0 would not like the duplication of "there are no files" messages as he requested some changes about them in the past reviews.

Though i think both views are completely valid. It's just a question of if the information is worth the duplication. I'm personally fine with both, maybe slightly leaning towards the current non-informative message.

simonjbeaumont:
> @simonjbeaumont i made it non-informative because i suspected @czechboy0 would not like the duplication of "there are no files" messages as he requested some changes about them in the past reviews.
> 
> Though i think both views are completely valid. It's just a question of if the information is worth the duplication. I'm personally fine with both, maybe slightly leaning towards the current non-informative message.

OK, that's fair enough, and sorry for missing that historical review context.

I'm happy with things the way they are now (subject to the soundness check passing again).

czechboy0 (approved):

MahdiBM:
@simonjbeaumont @czechboy0 what should i do to resolve this last soundness error?
```

** ✅ Found no unacceptable language.
** ERROR: Unsupported file extension for file (exclude or update this script): Plugins/OpenAPIGenerator/PluginsShared
** Running /code/scripts/run-swift-format.sh...
** ✅ Ran swift-format with no errors.
```

MahdiBM:
Had to exclude 2 to directories from `check-license-headers.sh`.
These are the symlinked directories, and the actual files are still checked for headers, as it threw errors for the files that didn't have the headers in the second to last CI run.
```diff
read -ra PATHS_TO_CHECK_FOR_LICENSE <<< "$( \
  git -C "${REPO_ROOT}" ls-files -z \
  ":(exclude).gitignore" \
  ":(exclude).spi.yml" \
  ":(exclude).swift-format" \
  ":(exclude).github/*" \
  ":(exclude)CODE_OF_CONDUCT.md" \
  ":(exclude)CONTRIBUTING.md" \
  ":(exclude)CONTRIBUTORS.txt" \
  ":(exclude)LICENSE.txt" \
  ":(exclude)NOTICE.txt" \
  ":(exclude)Package.swift" \
  ":(exclude)Package.resolved" \
  ":(exclude)README.md" \
  ":(exclude)SECURITY.md" \
  ":(exclude)scripts/unacceptable-language.txt" \
  ":(exclude)Tests/PetstoreConsumerTests/Generated" \
  ":(exclude)Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/*" \
  ":(exclude)docker/*" \
  ":(exclude)**/*.docc/*" \
  ":(exclude)**/.gitignore" \
  ":(exclude)**/Package.swift" \
  ":(exclude)**/Package.resolved" \
  ":(exclude)**/README.md" \
  ":(exclude)**/openapi.yaml" \
  ":(exclude)**/openapi.yml" \
  ":(exclude)**/petstore.yaml" \
  ":(exclude)**/openapi-generator-config.yaml" \
  ":(exclude)**/openapi-generator-config.yml" \
+  ":(exclude)Plugins/OpenAPIGenerator/PluginsShared" \
+  ":(exclude)Plugins/OpenAPIGeneratorCommand/PluginsShared" \
  | xargs -0 \
)"
```

czechboy0:
Thanks @MahdiBM for bearing with us during this marathon of a PR 🙂 Just updated from main and will land once CI passes.

PR #100, opened 2023-07-06
Docs: Change macos platform version

abdulajet:
### Motivation

I was following the tutorial, but it was unable to build due to `OpenAPIVapor` requiring a higher minimum target

Screenshot 2023-07-06 at 09 39 59

### Modifications

I changed the version of macOS in the "Generating server stubs in a Swift package" tutorial.

### Result

You should be able to follow the tutorial all the way through without errors

### Test Plan

Ran the soundness script.

### DIFF
--- a/Sources/swift-openapi-generator/Documentation.docc/Tutorials/_Resources/server.Package.1.swift
+++ b/Sources/swift-openapi-generator/Documentation.docc/Tutorials/_Resources/server.Package.1.swift
@@ -4,7 +4,7 @@ import PackageDescription
 let package = Package(
     name: "GreetingService",
     platforms: [
-        .macOS(.v10_15)
+        .macOS(.v13)
     ],
     targets: [
         .executableTarget(


czechboy0:
We reduced the deployment target in the last release, I wonder if Vapor's transport package could do the same, we could support 10.15 in the tutorial still? cc @0xTim for thoughts?

0xTim:
@czechboy0 @abdulajet done! If you do a package update it should work

simonjbeaumont:
Thanks @0xTim for the quick turnaround.

@abdulajet I'll close this PR, as we deliberately lowered the version. Could you confirm that you are now unblocked once you update to the latest version of swift-openapi-vapor and, if you still have any problem, open an issue and we'll happily take a look.

abdulajet:
@0xTim @simonjbeaumont @czechboy0 Nice one, thanks.

PR #103, opened 2023-07-07
Add finer-grain reference tests for translation phase

simonjbeaumont:
### Motivation

We have a reference test, which we've been making use of to validate end-to-end generation from OpenAPI document matches handwritten Swift files. This started out life as the canonical Petstore example from the OpenAPI specification but has since been augmented in various ways to test out some edge cases. This approach isn't scaling too well as we have an all-or-nothing test.

We also don't have great test coverage over the translation phase of the generator pipeline, which translates between the parsed OpenAPI document to an intermediate representation.

### Modifications

Add test harnesses for finer grain reference tests that exercise smaller sections of the translation phase. It should allow us to test a small OpenAPI document renders to _equivalent_ Swift source (stripping comments and reformatting).

### Result

Hopefully a more modular and maintainable testing strategy as we try to support more of the OpenAPI specification.

### Test Plan

This PR _is_ tests.

### DIFF
--- a/Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
+++ b/Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
@@ -86,7 +86,101 @@ final class SnippetBasedReferenceTests: XCTestCase {
         )
     }
 
-    func testComponentsSchemasObject() throws {
+    func testComponentsSchemasString() throws {
+        try self.assertSchemasTranslation(
+            """
+            schemas:
+              MyString:
+                type: string
+            """,
+            """
+            public enum Schemas {
+                public typealias MyString = Swift.String
+            }
+            """
+        )
+    }
+
+    func testComponentsObjectNoAdditionalProperties() throws {
+        try self.assertSchemasTranslation(
+            """
+            schemas:
+              MyObject:
+                type: object
+                properties: {}
+                additionalProperties: false
+            """,
+            """
+            public enum Schemas {
+                public struct MyObject: Codable, Equatable, Hashable, Sendable {
+                    public init() {}
+                    public init(from decoder: Decoder) throws {
+                        try decoder.ensureNoAdditionalProperties(knownKeys: [])
+                    }
+                }
+            }
+            """
+        )
+    }
+
+    func testComponentsObjectExplicitUntypedAdditionalProperties() throws {
+        try self.assertSchemasTranslation(
+            """
+            schemas:
+              MyObject:
+                type: object
+                properties: {}
+                additionalProperties: true
+            """,
+            """
+            public enum Schemas {
+                public struct MyObject: Codable, Equatable, Hashable, Sendable {
+                    public var additionalProperties: OpenAPIRuntime.OpenAPIObjectContainer
+                    public init(additionalProperties: OpenAPIRuntime.OpenAPIObjectContainer = .init()) {
+                        self.additionalProperties = additionalProperties
+                    }
+                    public init(from decoder: Decoder) throws {
+                        additionalProperties = try decoder.decodeAdditionalProperties(knownKeys: [])
+                    }
+                    public func encode(to encoder: Encoder) throws {
+                        try encoder.encodeAdditionalProperties(additionalProperties)
+                    }
+                }
+            }
+            """
+        )
+    }
+
+    func testComponentsObjectExplicitTypedAdditionalProperties() throws {
+        try self.assertSchemasTranslation(
+            """
+            schemas:
+              MyObject:
+                type: object
+                properties: {}
+                additionalProperties:
+                  type: integer
+            """,
+            """
+            public enum Schemas {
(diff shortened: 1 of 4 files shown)


Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
```diff
@@ -0,0 +1,351 @@
...
+/// Tests that the generator produces Swift code that match a reference.
+final class SnippetBasedReferenceTests: XCTestCase {
+    func testComponentsHeadersInline() throws {
+        try self.assertHeadersTranslation(
+            """
+            openapi: "3.0.3"
+            info:
+              title: "Test"
+              version: "0.0.0"
+            paths: {}
```
czechboy0, line 28: Overall looks great, just one idea - to avoid having to duplicate the openapi version, info, and paths keys in every snippet, could we have them be added by default, unless specified explicitly? That way you can only specify the important part of the OpenAPI doc in each test, without duplicating the unimportant parts.

Might require using Yams to parse, fill in defaults, then serialize again, but could be worthwhile.

czechboy0 (approved):
Looks great, just one enhancement idea.

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
simonjbeaumont: I added a followup commit with one way of doing this: https://github.com/apple/swift-openapi-generator/pull/103/commits/3f377462743080ff376247a15d516d375ff8b22a.

It creates the translator from a YAML for the components only (cf. the whole OpenAPI doc).

I'm on the fence about it. There was something quite nice about knowing we used a proper (albeit) minimal OpenAPI document for each test case—even if that did mean some duplication.

czechboy0: I think good helper function naming can still convey that under the hood, we're producing a full OpenAPI doc, I just prefer to remove the duplicated/unchanging values, to make the tests easier to read, as you know that if it's specified in the test, it's meaningful to the object code being tested. When there's a ton of duplication, it's really hard to know what is being tested and what's just "context" bringup.

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
```diff
@@ -0,0 +1,321 @@
...
+//===----------------------------------------------------------------------===//
+import OpenAPIKit30
+import XCTest
+import Yams
+@testable import _OpenAPIGeneratorCore
+
+/// Tests that the generator produces Swift code that match a reference.
+final class SnippetBasedReferenceTests: XCTestCase {
+    func testComponentsHeadersInline() throws {
+        try self.assertHeadersTranslation(
```
czechboy0, line 22: Yeah these three methods are a great way to test this, IMO. I'd be in favor of fleshing this out. 👍

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
```diff
@@ -0,0 +1,777 @@
...
+    func testComponentsSchemasString() throws {
+        try self.assertSchemasTranslation(
+            """
+            schemas:
+              MyString:
+                type: string
+            """,
+            """
+            public enum Schemas {
+                public typealias MyString = Swift.String
```
czechboy0, line 98: I see here we have 4 spaces and 2 in the previous test, is that intentional?

czechboy0 (changes requested):
Just a question about the whitespace, otherwise lgtm!

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
```diff
@@ -0,0 +1,777 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import OpenAPIKit30
+import XCTest
+import Yams
+@testable import _OpenAPIGeneratorCore
+
+/// Tests that the generator produces Swift code that match a reference.
+final class SnippetBasedReferenceTests: XCTestCase {
```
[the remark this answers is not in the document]
simonjbeaumont: Might be worth it, in time. I don't think that's right now though. I could easily see this test suite covering all the "types" use cases, for example, without being too problematic.

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
simonjbeaumont: Most of them use 4 spaces for the Swift code. It doesn't really matter what they use because we're not testing for that in the helper function. These are fuzzy tests w.r.t. to formatting to make them a little more robust.

czechboy0 (approved):
:shipit:

PR #108, opened 2023-07-10
Simplify swift-format soundness script

simonjbeaumont:
### Motivation

The `run-swift-format.sh` script run as part of soundness explicitly lists the directories to format. This is to avoid running `swift-format` on the contents of `.build/` which contains a checkouts of all the dependencies. This isn't necessary, as we can use Git to tell us what files need to be formatted, which can also make for a more robust script/command as we add new directories.

Additionally, with the previous strategy, `Package.swift` at the root of the repository was ignored. This was deliberate, but in hindsight I don't think the lack of trailing comma enforcement is worth us excluding it.

### Modifications

- Use `git ls-files` to determine which files need to be passed to `swift-format`.

### Result

- Script is simpler (also easier to add a one-shot command as a pre-commit hook).
- Package.swift is also formatted.

### Test Plan

None.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -90,7 +90,7 @@ let package = Package(
         .testTarget(
             name: "OpenAPIGeneratorCoreTests",
             dependencies: [
-                "_OpenAPIGeneratorCore",
+                "_OpenAPIGeneratorCore"
             ]
         ),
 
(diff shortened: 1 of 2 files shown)


czechboy0 (approved):
Net-negative line count, a simplified script, and more consistent code. What are you, a magician? 🙂💯

czechboy0:

PR #109, opened 2023-07-10
Generate @frozen enums for operation outputs and request/response bodies

### LINKED ISSUE #105: Mark operation output and request/response body enums as @frozen
Since it's not expected that the generated code should support library evolution, we could add @frozen to all of the generated enums. This would greatly improve the ergonomics of switching over enums in the generated code. 

Borrowing an example from [another issue](https://github.com/apple/swift-openapi-generator/issues/104):
```swift
switch result {
case let .ok(reponse):
    case let .json(payload):
        print(payload.body, payload.headers)
    @unknown default:
        break
    }
@unknown default:
    break
} 
```

would become 

```swift
switch result {
case let .ok(reponse):
    case let .json(payload):
        print(payload.body, payload.headers)
    }
} 
```

simonjbeaumont:
### Motivation

The generated code contains several enum types. A lot of these are just namespaces (i.e. the enum has no cases), but some are "proper" enums that we expect adopters to switch over. For example the operation output, and the request and response bodies.

To make using these types more ergonomic, we can annotate them as `@frozen`.  This is something we already do for other generated enum types (e.g. OneOf and enum types declared in the OpenAPI document).

### Modifications

- Add desired `@frozen` attributes to reference code
- Generate `@frozen` enum for request bodies
- Generate `@frozen` enum for response bodies
- Generate `@frozen` enum for operation outputs

### Result

Switching over operation outputs and request/response bodies is simpler.

### Resolves

- Resolves #105.

### Test Plan

Reference test updated.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/RequestBody/translateRequestBody.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/RequestBody/translateRequestBody.swift
@@ -150,6 +150,7 @@ extension TypesFileTranslator {
         members: [Declaration]
     ) -> Declaration {
         let bodyEnumDecl: Declaration = .enum(
+            isFrozen: true,
             accessModifier: config.access,
             name: typeName.shortSwiftName,
             conformances: Constants.Operation.Output.conformances,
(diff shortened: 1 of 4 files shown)


simonjbeaumont:
I'm going to move this to draft and mark as blocked for now because there's some more discussion that maybe this is the wrong direction to be moving in for the generated code.

czechboy0 (approved):
lgtm on the implementation, but as you say, this is still being discussed elsewhere.

czechboy0:

czechboy0:
Updated from semver/minor to semver/patch, as this isn't a breaking change.

PR #121, opened 2023-07-17
[Tests] Add an integration test product for easier iteration

czechboy0:
### Motivation

Some IDEs don't allow you to trigger building of targets, only products.

### Modifications

Added an umbrella library product to make it possible to build locally in an IDE, not just using `swift build`.

### Result

Now the product shows up in IDEs, allowing developers to iterate on the code and rebuild.

### Test Plan

Tested that it shows up locally.


simonjbeaumont (approved):
LGTM, once swift-format is happy.

Issue #120, opened 2023-07-17
Disable the ExistentialAny feature on all the repos for 5.8


czechboy0:
Disable the ExistentialAny feature on all the repos temporarily.

See #119 for details, we just need to disable the enforcement temporarily.

czechboy0:
Ok this is done, we now only enable the feature enforcement on 5.9+.

PR #122, opened 2023-07-17
Disable ExistentialAny temporarily

czechboy0:
### Motivation

Unfortunately until we adopt 5.9, adding ExistentialAny on upstream packages has unintended consequences for some downstream packages, so disabling for now. Details in https://github.com/apple/swift-openapi-generator/issues/120

### Modifications

Disabled the feature enforcement, but the code changes are there, so downstream adopters can still use them.

### Result

We won't be seeing the issue described in https://github.com/apple/swift-openapi-generator/issues/120.

### Test Plan

PR CI, which discovered the original issue.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -14,6 +14,17 @@
 //===----------------------------------------------------------------------===//
 import PackageDescription
 
+// General Swift-settings for all targets.
+var swiftSettings: [SwiftSetting] = []
+
+#if swift(>=5.9)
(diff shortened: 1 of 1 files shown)


czechboy0:
@simonjbeaumont Took @FranzBusch's idea and only disabled it on 5.8 in Package.swift, so gets us the same result.

simonjbeaumont (approved):

Issue #27, opened 2023-05-25
Dry run option on the CLI


czechboy0:
Since the CLI overwrites existing files in `--output-path`, we could have a `--dry-run` option that just prints what it would have done.

PR #123, opened 2023-07-17
Add dry run option to CLI

denil-ct:
### Motivation

Resolves https://github.com/apple/swift-openapi-generator/issues/27

### Modifications

Add a dry run option to CLI `--dry-run`

### Result

If passed, the changes would be printed to terminal, without actually affecting the file system.

### Test Plan

Manually tested with various input combinations.


Sources/swift-openapi-generator/runGenerator.swift
```diff
@@ -67,44 +99,69 @@ extension _Tool {
...
+                    diagnostics: diagnostics
+                )
+                return output.contents
+            },
+            isDryRun: isDryRun
+        )
+        if !isDryRun {
+            print("File \(outputFilePath.lastPathComponent): \(didChange ? "changed" : "unchanged")")
         }
-        print("File \(outputFilePath.lastPathComponent): \(didChange ? "changed" : "unchanged")")
```
czechboy0, line 86: We can continue printing this line, it'd be useful to still see if the contents of this file changed or not, even in dry run.

denil-ct: Moved it to `replaceFileContents` method to avoid duplication.

Sources/swift-openapi-generator/runGenerator.swift
```diff
@@ -42,20 +44,25 @@ extension _Tool {
         let filePathForMode: (GeneratorMode) -> URL = { mode in
             outputDirectory.appendingPathComponent(mode.outputFileName)
         }
+        if isDryRun {
+            print("--------------------------------")
+            print("Dry run mode: No files will be created or modified")
+        }
```
czechboy0, line 50: I don't think we need it, as it's already included in the launch summary that it's dry run.

Sources/swift-openapi-generator/runGenerator.swift
```diff
@@ -67,44 +74,66 @@ extension _Tool {
...
+            print("File \(path.lastPathComponent) does not exist.\nCreating new file...")
             didChange = true
         }
         if didChange {
-            try data.write(to: path)
+            if isDryRun {
+                print("Writing data to \(path.lastPathComponent)...")
+            } else {
+                try data.write(to: path)
+            }
```
czechboy0, line 136: This method shouldn't print anything, just return whether the file contents have changed. Let's keep the printing in the caller, though.

czechboy0 (changes requested):
Two more minor requests, thanks.

Sources/swift-openapi-generator/runGenerator.swift
denil-ct: I thought about it, but then I couldn't differentiate if the file exists or not. Should I drop that operation or return a tuple or a state instead of a singular bool?

czechboy0: Sure, expanding the return value makes sense.

czechboy0 (approved):
Looks good, thanks!

Sources/swift-openapi-generator/runGenerator.swift
simonjbeaumont: I think that we've added a bunch of unnecessary complexity between these two functions.

I think that's because we said that this function shouldn't print. I don't agree. IMO a function that can operate in dry run mode should either:
- Do what it was asked (doesn't need to print); or
- Print what it would have done.

If we allowed this function to print the rest of the stack would be unaffected.

simonjbeaumont (changes requested):

Sources/swift-openapi-generator/runGenerator.swift
```diff
@@ -67,45 +70,72 @@ extension _Tool {
...
+                    config: config,
+                    diagnostics: diagnostics
+                )
+                return output.contents
+            },
+            isDryRun: isDryRun
+        )
+        if result.didCreate && isDryRun {
+            print("File \(outputFilePath.lastPathComponent) does not exist.\nCreating new file...")
+        }
```
czechboy0, line 98: I think we can get rid of this print, the summary below is enough to know what the generator did for this file.

denil-ct: Just overwritten does not convey the full meaning imo. It doesn't distinguish between the file existing or not.

Sources/swift-openapi-generator/runGenerator.swift
simonjbeaumont: Yes. Side-effects should be replaced with prints. Rather than hypothetical-side-effects being passed back up through the stack for no other reason and the caller needing to keep track of when how to print them.

denil-ct: In that case, can we remove the return value entirely?

Sources/swift-openapi-generator/runGenerator.swift
czechboy0: Is that important though, whether the file existed before or not? What would it help you achieve?

denil-ct: I thought from the point of filesystem operations. To the end user its not much relevant.
Will drop it then.

Sources/swift-openapi-generator/runGenerator.swift
simonjbeaumont: IMO, yes.

Sources/swift-openapi-generator/runGenerator.swift
```diff
@@ -104,8 +118,12 @@ extension _Tool {
...
-            try data.write(to: path)
+            print("File \(path.lastPathComponent) will be overwritten.")
+            if !isDryRun {
+                try data.write(to: path)
+            }
+        } else {
+            print("File \(path.lastPathComponent) will remain unchanged.")
         }
-        return didChange
     }
```
simonjbeaumont, line 128: At this stage, I'm going to approve the PR because it does what it set out to do.

However, IMO the code has become a bit too branchy. On inspection, I don't think that's really because of this PR.


We could also make use of some `guards` to early-return and make the code more linear.

Finally, this function also only replaces the contents if they are different but isn't clear from its name. This is an important nuance because IIUC this affects the timestamp on the file and, therefore, the rest of the build graph.

This function could therefore be rewritten as:

```swift
    static func replaceFileContentsIfDifferent(
        at path: URL,
        with data: Data,
        isDryRun: Bool
    ) throws {
        if let existingData = try? Data(contentsOf: path), existingData == data {
            print("File \(path.lastPathComponent) already up to date.")
            return
        }
        guard !isDryRun else {
            print("Not writing file \(path.lastPathComponent) (dry run).")
            return
        }
        print("Writing file \(path.lastPathComponent).")
        try data.write(to: path)
    }
```

At this point it is questionable why we had this `@discardableResult` function that took a throwing closure anyway, because at the call site it's basically the only thing that's called by `runGenerator(...)`.

All that said, I'm hitting approve here because removing this complexity isn't on your plate @denil-ct :)

simonjbeaumont (approved):

Sources/swift-openapi-generator/runGenerator.swift
denil-ct: I'll be happy to simplify it another PR post this.

Sources/swift-openapi-generator/GenerateCommand.swift
```diff
@@ -48,10 +48,17 @@ struct _GenerateCommand: AsyncParsableCommand {
     )
     var isPluginInvocation: Bool = false
 
+    @Flag(
+        help:
+            "Simulate the command and print the operations, without actually affecting the file system."
+    )
+    var isDryRun: Bool = false
```
czechboy0, line 55: Oh sorry for the late request, could we just make it `--dry-run`? At the moment, the name would be `--is-dry-run`, which feels like a question rather than an instruction.

I believe there's a parameter on the Option property wrapper customize the CLI name, which means you won't need to change the property name itself.

czechboy0 (changes requested):
One last request, otherwise lgtm.

Sources/swift-openapi-generator/GenerateCommand.swift
denil-ct: I agree, although I just followed the same style as `isPluginInvocation`

czechboy0: Good point, but fortunately that one is getting renamed in #98.

czechboy0:
@denil-ct Unfortunately #98 touched similar code as your PR, so there are a few conflicts you'll need to resolve before landing. Once that's all green, and https://github.com/apple/swift-openapi-generator/pull/123#discussion_r1269421605 is addressed, we can land it.

czechboy0 (approved):

Issue #70, opened 2023-06-16
Support recursive types


scsinke:
Currently the package does not support recursive types. Should this be possible? Maybe I'm doing something wrong. otherwise, I would propose to add this constraint in the docs. I have this use case due to Server Driven UI. This results in components or actions that can be embedded in each other.

Errors:

- `Value type 'x' cannot have a stored property that recursively contains it`
- `Value type 'x' has infinite size`

czechboy0:
Thanks for filing the issue, we'd certainly accept a contribution of support of recursive types.

Could you please add a snippet of OpenAPI that uses it in YAML, and how you imagine the generated code would change? That'll help us discuss the desired solution.

scsinke:
I don't know if this can solve the issue. But maybe we can make use of the `indirect` keyword? https://www.hackingwithswift.com/example-code/language/what-are-indirect-enums

czechboy0:
It seems `indirect` can only be applied to enums, but not structs. Can you paste a snippet of OpenAPI to show what the use case is?

czechboy0:
Seems that indirect can only be used on enums, not structs.

Before we consider the explicit Box approach, I'd like to see if there's another recommended way in Swift to solve this.

scsinke:
Hereby a simplified spec we use. For the full openAPI spec we use. check [this](https://swagger.customerapp.experience-dev.hema.digital/) link.

```json
{
  "openapi": "3.0.2",
  "components": {
    "schemas": {
      "Action": {
        "anyOf": [
          {
            "$ref": "#/components/schemas/PromptAction"
          }
        ]
      },
      "PromptAction": {
        "type": "object",
        "properties": {
          "type": {
            "type": "string",
            "enum": [
              "PROMPT"
            ]
          },
          "prompt": {
            "$ref": "#/components/schemas/Prompt"
          }
        },
        "required": [
          "type",
          "prompt"
        ]
      },
      "Prompt": {
        "anyOf": [
          {
            "$ref": "#/components/schemas/DialogPrompt"
          }
        ]
      },
      "DialogPrompt": {
        "type": "object",
        "properties": {
          "type": {
            "type": "string",
            "enum": [
              "DIALOG"
            ]
          },
          "title": {
            "type": "string"
          },
          "message": {
            "type": "string"
          },
          "confirmButtonTitle": {
            "type": "string"
          },
          "cancelButonTitle": {
            "type": "string"
          },
          "confirmAction": {
[...]

PR #130, opened 2023-07-19
Validate incoming OpenAPI docs using OpenAPIKit's built-in validation

czechboy0:
### Motivation

When provided with an OpenAPI document that has recursion, the generator either crashes, or produces Swift code that doesn't compile. We should catch recursion earlier in the process, and emit a descriptive error.

### Modifications

This PR adds two validation steps:
- OpenAPIKit's `validate` method catches some structural issues in the OpenAPI doc.
- OpenAPIKit's `locallyDereferenced()` method is a great way to ensure no reference cycles appear in the document.

Catching reference cycles earlier in the process is great, as even if we generate the Swift code, it won't compile until we have some explicit support for recursive types (tracked by #70).

### Result

Now, when an OpenAPI document with recursion is provided, instead of crashing or producing non-compiling Swift code, it prints a descriptive error and returns a non-0 code.

### Test Plan

Tested manually on the CLI with purposefully malformed documents. But since this isn't our code, I don't want to add detailed tests of the validation details.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
+++ b/Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
@@ -131,6 +131,10 @@ func makeGeneratorPipeline(
                     // Validate that the document is dereferenceable, which
                     // catches reference cycles, which we don't yet support.
                     _ = try doc.locallyDereferenced()
+
+                    // Also explicitly dereference the parts of components
+                    // that the generator uses. `locallyDereferenced()` above
+                    // only dereferences paths/operations, but not components.
                     let components = doc.components
                     try components.schemas.forEach { schema in
                         _ = try schema.value.dereferenced(in: components)
--- a/Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
+++ b/Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
@@ -122,7 +122,19 @@ func makeGeneratorPipeline(
                     diagnostics: diagnostics
                 )
             },
-            postTransitionHooks: []
+            postTransitionHooks: [
+                { doc in
+
+                    // Run OpenAPIKit's built-in validation.
+                    try doc.validate()
+
+                    // Validate that the document is dereferenceable, which
+                    // catches reference cycles, which we don't yet support.
+                    _ = try doc.locallyDereferenced()
+
+                    return doc
+                }
+            ]
         ),
         translateOpenAPIToStructuredSwiftStage: .init(
             preTransitionHooks: [],
--- a/Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
+++ b/Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
@@ -131,6 +131,22 @@ func makeGeneratorPipeline(
                     // Validate that the document is dereferenceable, which
                     // catches reference cycles, which we don't yet support.
                     _ = try doc.locallyDereferenced()
+                    let components = doc.components
+                    try components.schemas.forEach { schema in
+                        _ = try schema.value.dereferenced(in: components)
+                    }
+                    try components.parameters.forEach { schema in
+                        _ = try schema.value.dereferenced(in: components)
+                    }
+                    try components.headers.forEach { schema in
+                        _ = try schema.value.dereferenced(in: components)
+                    }
+                    try components.requestBodies.forEach { schema in
+                        _ = try schema.value.dereferenced(in: components)
+                    }
+                    try components.responses.forEach { schema in
+                        _ = try schema.value.dereferenced(in: components)
+                    }
 
                     return doc
                 }
--- a/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeAssigner.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeAssigner.swift
@@ -425,6 +425,9 @@ enum JSONReferenceParsingError: Swift.Error {
     /// An error thrown when parsing a JSON reference that points to
     /// other OpenAPI documents.
     case externalPathsUnsupported(String)
+
+    /// Reference cycle are unsupported.
+    case referenceCycleUnsupported(String)
 }
 
 extension JSONReferenceParsingError: CustomStringConvertible {
@@ -434,6 +437,8 @@ extension JSONReferenceParsingError: CustomStringConvertible {
             return "JSON references outside of #/components are not supported, found: \(string ?? "<nil>")"
         case let .externalPathsUnsupported(string):
             return "External JSON references are not supported, found: \(string)"
+        case let .referenceCycleUnsupported(string):
+            return "Reference cycles are not supported, found: \(string)"
         }
     }
 }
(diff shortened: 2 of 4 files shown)


simonjbeaumont (commented):
Rather than roll our own cycle detection, can we leverage something from OpenAPIKit? It seems to provide `dereferenced(...)` functions that throw on cycles—here's a simple example:

```swift
import XCTest
import OpenAPIKit30
import Yams

final class ReferenceCycleTests: XCTestCase {
    func test() throws {
        let componentsYAML = """
        schemas:
          A:
            type: string
          B:
            $ref: "#/components/schemas/A"
          C:
            $ref: "#/components/schemas/B"
          D:
            $ref: "#/components/schemas/E"
          E:
            $ref: "#/components/schemas/D"
        """
        let components = try YAMLDecoder().decode(OpenAPI.Components.self, from: componentsYAML)
        XCTAssertNotNil(try components.schemas["A"]!.dereferenced(in: components))
        XCTAssertNotNil(try components.schemas["B"]!.dereferenced(in: components))
        XCTAssertNotNil(try components.schemas["C"]!.dereferenced(in: components))
        XCTAssertThrowsError(try components.schemas["D"]!.dereferenced(in: components))
        XCTAssertThrowsError(try components.schemas["D"]!.dereferenced(in: components))
    }
}
```

This test passes, and the error returned for the cycle cases is:
```
"Encountered a JSON Schema $ref cycle that prevents fully dereferencing document at '#/components/schemas/B'. This type of reference cycle is not inherently problematic for JSON Schemas, but it does mean OpenAPIKit cannot fully resolve references because attempting to do so results in an infinite loop over any reference cycles. You should still be able to parse the document, just avoid requesting a `locallyDereferenced()` copy."
```

Additionally, once we land a solution here, could you add something to the reference test suite? e.g. adding this currently results in an infinite loop, as you expect, perhaps something that tests that we throw the right error:

```swift
    func testComponentsSchemasReferenceCycle() throws {
        try self.assertSchemasTranslation(
            """
            schemas:
              A:
                $ref: "#/components/schemas/B"
              B:
                $ref: "#/components/schemas/A"
            """,
            """
            public enum Schemas {
                public typealias A = A
            }
            """
        )
    }
```

czechboy0:
As the OpenAPIKit error says, `You should still be able to parse the document, just avoid requesting a `locallyDereferenced()` copy."` - that's what we do. We never use the dereferencing functions in OpenAPIKit, because our generator retains as much structure as possible, and dereferencing is lossy.

So we _could_ dereference the whole document once at the start, and throw away the result if it doesn't throw an error – would you prefer that?

Regarding the snippet test, since we're testing that we throw an error, I'm not sure what code we'd expect to be generated.

simonjbeaumont:
> As the OpenAPIKit error says, `You should still be able to parse the document, just avoid requesting a locallyDereferenced() copy."` - that's what we do. We never use the dereferencing functions in OpenAPIKit, because our generator retains as much structure as possible, and dereferencing is lossy.
> 
> So we _could_ dereference the whole document once at the start, and throw away the result if it doesn't throw an error – would you prefer that?

That's what I was meaning. We could use this function at the top of `isSchemaSupported` to test for and rethrow if there is a cycle. We'd then discard the result of the function. But it would mean we wouldn't need to carry around the bag of seen references and implement our own cycle detection IIUC.


> Regarding the snippet test, since we're testing that we throw an error, I'm not sure what code we'd expect to be generated.

Right, I said "e.g. adding this currently results in an infinite loop, as you expect, perhaps something that tests that we throw the right error" so I'm not proposing that exact code, just used it as a way of showing the OpenAPI snippet. It would be a test that verifies the error thrown is whatever ends up in this PR.

simonjbeaumont:
> The dereferencing is expensive, so I think we should just let OpenAPIKit dereference the full document at the start, during preflight validation, and then just assume there are no cycles from that point on. We call things like `isSchemaSupported` frequently, so I don't think we should call it every time. But the result will be the same, and will mean I can undo most of this PR 🙂

That works for me. I guess we'll just do this as part of the parsing stage of the pipeline. Might be a good place to use the post-transition hook in that pipeline stage, or just bolt it into the implementation of that stage—up to you.
 
> Re the error, sure we can test for the exact error. I'll change the PR now to use OpenAPIKit dereferencing, will take it back to draft.

SGTM—thanks @czechboy0!

czechboy0:
Ok, ready for another look, @simonjbeaumont 🙏

Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
```diff
@@ -122,7 +122,39 @@ func makeGeneratorPipeline(
...
+                    // Run OpenAPIKit's built-in validation.
+                    try doc.validate()
+
+                    // Validate that the document is dereferenceable, which
+                    // catches reference cycles, which we don't yet support.
+                    _ = try doc.locallyDereferenced()
+
+                    // Also explicitly dereference the parts of components
+                    // that the generator uses. `locallyDereferenced()` above
+                    // only dereferences paths/operations, but not components.
```
simonjbeaumont, line 137: Why not do it for all parts of `components`, rather than just the ones we use _today_?

Otherwise I'd be concerned that we'd forget to update this as we move forward.

If there's a reference cycle in the OpenAPI document provided to the generator then it's a wonky input regardless of whether we try to make use of _that_ part of the document.

Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
```diff
@@ -122,7 +122,39 @@ func makeGeneratorPipeline(
                     diagnostics: diagnostics
                 )
             },
-            postTransitionHooks: []
+            postTransitionHooks: [
+                { doc in
+
+                    // Run OpenAPIKit's built-in validation.
+                    try doc.validate()
```
simonjbeaumont, line 129: Are we happy just surfacing the errors, as-is, from OpenAPIKit?

simonjbeaumont (commented):
Thanks for updating this. I think it's great that we found a way to leverage this in the library we already depend on, so thanks for looking into it!

Just a couple of questions on this one.

Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
czechboy0: Reference cycles are completely legal in OpenAPI, we just don't currently have a way to generate compiling Swift code for them, tracked by #70.

So I wouldn't want to reject OpenAPI documents that have a reference cycle in part of the document we don't read, as it feels overly strict.

In the future, once we support recursive types, we'll be able to remove much of this cycle checking.

Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
czechboy0: Yes, the errors are well written, and I checked all of the error checking, and we do want those validations as well.

simonjbeaumont (approved):

hactar:
I was trying out this generator for the first time and triggered this issue. The problem is that the error message can be misleading:


`Encountered a JSON Schema $ref cycle that prevents fully dereferencing document at '#/components/schemas/KommentarDTO'. This type of reference cycle is not inherently problematic for JSON Schemas, but it does mean OpenAPIKit cannot fully resolve references because attempting to do so results in an infinite loop over any reference cycles. You should still be able to parse the document, just avoid requesting a locallyDereferenced() copy.`

I understood this to mean I have to change some generator config setting to not try to request a locallyDereferenced copy. Spent some time trying to "fix" this first ☺️. Only once I read the source code comment I now understand that there is no workaround at this moment and that the openapi file that I would like to use will not work. I think it would be useful to catch this specific error and state that cyclic jsons are not supported yet?

Issue #136, opened 2023-07-21
Don't generate format: date as Foundation.Date, use a string


czechboy0:
[This](https://github.com/apple/swift-openapi-generator/issues/84#issuecomment-1645674563) comment from @yanniks made me check what type we generate for `format: date` (not `date-time`), and it's `Foundation.Date`. That's wrong, as `Foundation.Date` only represents a full timestamp (aka `date-time` in OpenAPI terms).

This is an oversight, we should just generate a simple string, as there isn't an obvious currency type to deserialize dates (without time) into.

Should be as simple as removing `.date` from [here](https://github.com/apple/swift-openapi-generator/blob/31bb51799533e99e77fda5eef681571b975a39b5/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeMatcher.swift#L218).

yanniks:
Hi Honza,

thanks for opening this issue! I agree that changing `date` values to a string is definitely a better solution than the current implementation, especially since the default `ISO8601DateTranscoder` is not able to decode OpenAPI `date` formats at all.

However, I would prefer some kind of predefined date object as the format (`yyyy-MM-dd`) is fixed. Otherwise, all developers will have to implement their own (de-) serialization, although the OpenAPI specification defines a concrete format.

czechboy0:
@yanniks Our approach has been to use platform-specific types when they exist, and otherwise let developers solve it on their side. Unfortunately, I don't think there's a "date" (no time) type in the Swift standard library or Foundation, and I'd prefer we don't try to invent one, so keeping it as the string that arrives in JSON is the right call for now.

Parsing it as Date is a bug, it was never meant to be this way. A contribution with the fix is always welcome 🙂

yanniks:
@czechboy0 Okay, got it! I'll provide a PR that changes this behavior.

By the way: when talking about platform-specific types, is there a specific reason why UUIDs are not represented as `Foundation.UUID` objects?

czechboy0:
Technically, `format: uuid` isn't part of the OpenAPI 3.0.3 specification, it's just a convention people use, like `format: email`. We haven't felt we needed to support it yet, for that reason, but I also believe that UUID is defined well enough that it'd be a reasonable thing to do.

PR #137, opened 2023-07-21
fix: Correct data type of date-formatted strings

yanniks:
### Motivation

While OpenAPI date-formatted strings do not contain information about a specific time, `Foundation.Date` only supports full timestamps. Also, `ISO8601DateTranscoder` does not support parsing ISO-8601 dates without time information. This commit introduces the expected behavior that date-only strings are not automatically decoded to `Foundation.Date` objects. This PR fixes issue #136.

### Modifications

`.date` was removed from the case-condition that defined `Foundation.Date` as the applicable type name.

### Result

`date`-formatted strings are no longer represented as `Foundation.Date` but as `Swift.String` objects.

### Test Plan

Tests were adapted so that the correct Swift type for `date`-formatted strings is ensured.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeMatcher.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeMatcher.swift
@@ -215,7 +215,7 @@ struct TypeMatcher {
                 typeName = .swift("String")
             case .binary:
                 typeName = .foundation("Data")
-            case .date, .dateTime:
+            case .dateTime:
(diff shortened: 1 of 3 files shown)


czechboy0 (approved):
Thank you!

czechboy0:
Technically, this will change which type we generate and could be source breaking, but the previously generated code never actually worked, so I'm leaning towards not requiring a minor here, WDYT @simonjbeaumont?

simonjbeaumont (approved):

czechboy0:

PR #140, opened 2023-07-25
Enable the generated server-side code to validate the content-type.

### LINKED ISSUE #16: Server should validate Content-Type header
Currently, a generated server does not validate the content-type header before it tries to parse the request body contents, which means that if incorrect request body data is sent, the error message will be thrown by e.g. the JSONDecoder, and might be a bit cryptic, versus a clear error thrown by a method like [`validateContentTypeIfPresent`](https://github.com/apple/swift-openapi-generator/blob/main/Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Client.swift#L79) does for responses bodies.

takeshi-1000:
### Motivation
I am interested this tool and have selected something that I can work on.
Resolve #16 

### Modifications

To validate content-type header string in server program, I modify `traslateServerDesirializer` method . 

### Result

For example you will define below opeapi yaml 

```
  /greet:
    post:
      operationId: postGreeting
      requestBody:
        required: true
        content:
          text/plain:
            schema:
              type: string
```

when you try the request, runtime error happen.

```
$ curl localhost:8080/api/greet -X POST \             
> -H "Content-Type: application/json" \
> -d '{ "name": "foo" }'

{"error":true,"reason":"Server error - operationID: postGreeting, request: path: \/api\/greet, query: <nil>, method: HTTPMethod(value: OpenAPIRuntime.HTTPMethod.(unknown context at $106e9fd78).OpenAPIHTTPMethod.POST), header fields: [Host: localhost:8080, User-Agent: curl\/7.88.1, Accept: *\/*, Content-Type: application\/json, Content-Length: 64], body (prefix): { \"name\": \"foo\"}, metadata: path parameters: [:], query parameters: [], operationInput: <nil>, operationOutput: <nil>, underlying error: Unexpected Content-Type header: application\/json"}
```

### Test Plan

I have made the necessary changes to adapt to the newly generated code on the reference server side and have confirmed that it passes the tests locally.

### DIFF
--- a/Tests/PetstoreConsumerTests/Test_Server.swift
+++ b/Tests/PetstoreConsumerTests/Test_Server.swift
@@ -231,6 +231,46 @@ final class Test_Server: XCTestCase {
         )
     }
 
+    
+    func testCreatePet_withIncorrectContentType() async throws {
+        client = .init(
+            createPetBlock: { input in
+                return .created(
+                    .init(
+                        headers: .init(
+                            X_Extra_Arguments: .init(code: 1)
+                        ),
+                        body: .json(
+                            .init(id: 1, name: "Fluffz")
+                        )
+                    )
+                )
+            }
+        )
+        
+        do {
+            _ = try await server.createPet(
+                .init(
+                    path: "/api/pets",
+                    method: .post,
+                    headerFields: [
+                        .init(name: "x-extra-arguments", value: #"{"code":1}"#),
+                        .init(name: "content-type", value: "text/plain; charset=utf-8"),
+                    ],
+                    encodedBody: #"""
+                        {
+                          "name" : "Fluffz"
+                        }
+                        """#
+                ),
+                .init()
+            )
+            XCTFail("The method should have thrown an error.")
+        } catch {
+            XCTAssertNotNil(error)
+        }
+    }
+    
     func testUpdatePet_204_withBody() async throws {
         client = .init(
             updatePetBlock: { input in
--- a/Tests/PetstoreConsumerTests/Test_Server.swift
+++ b/Tests/PetstoreConsumerTests/Test_Server.swift
@@ -236,6 +236,7 @@ final class Test_Server: XCTestCase {
         client = .init(
             createPetBlock: { input in
                 XCTFail("The handler should not have been called")
+                fatalError("Unreachable")
             }
         )
         
--- a/Tests/PetstoreConsumerTests/Test_Server.swift
+++ b/Tests/PetstoreConsumerTests/Test_Server.swift
@@ -235,16 +235,7 @@ final class Test_Server: XCTestCase {
     func testCreatePet_withIncorrectContentType() async throws {
         client = .init(
             createPetBlock: { input in
-                return .created(
-                    .init(
-                        headers: .init(
-                            X_Extra_Arguments: .init(code: 1)
-                        ),
-                        body: .json(
-                            .init(id: 1, name: "Fluffz")
-                        )
-                    )
-                )
+                XCTFail("The handler should not have been called")
             }
         )
         
@@ -266,9 +257,7 @@ final class Test_Server: XCTestCase {
                 .init()
             )
             XCTFail("The method should have thrown an error.")
-        } catch {
-            XCTAssertNotNil(error)
-        }
+        } catch {}
     }
     
     func testUpdatePet_204_withBody() async throws {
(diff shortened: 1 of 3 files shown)


czechboy0 (changes requested):
Hi @takeshi-1000, thanks for taking this up!

Overall, your change looks good, can you just add a unit test? For example, you could create a variant of the test for `createPet` ([here](https://github.com/apple/swift-openapi-generator/blob/22a7d7af0be75447625e091d5ddf73bdd192c96d/Tests/PetstoreConsumerTests/Test_Server.swift#L129)) that purposefully passes the incorrect content type and verifies that an error was thrown with `XCTAssertThrowsError` (but no need to test the contents of the error, as it's an internal type).

takeshi-1000:
@czechboy0 
Thank you for reviewing . 
I tried to add, but `XCTAssertThrowsError` seems not to supported async method. 
https://developer.apple.com/documentation/xctest/1500795-xctassertthrowserror

2023-07-26 8 48のイメージ

I am considering the following code, but is there any better alternative?

```
 client = .init(
            createPetBlock: { input in
                return .created(
                    .init(
                        headers: .init(
                            X_Extra_Arguments: .init(code: 1)
                        ),
                        body: .json(
                            .init(id: 1, name: "Fluffz")
                        )
                    )
                )
            }
        )
        
        do {
            _ = try await server.createPet(
                .init(
                    path: "/api/pets",
                    method: .post,
                    headerFields: [
                        .init(name: "x-extra-arguments", value: #"{"code":1}"#),
                        .init(name: "content-type", value: "text/plain; charset=utf-8"),
                    ],
                    encodedBody: #"""
                        {
                          "name" : "Fluffz"
                        }
                        """#
                ),
                .init()
            )
            XCTFail("The method should have thrown an error.")
        } catch {
            XCTAssertNotNil(error)
        }
```

czechboy0:
Ah you're right. Yeah, a manual do/catch block, where you XCTFail at the end of the do block, and do nothing in the catch block is how I've worked around this limitation.

Tests/PetstoreConsumerTests/Test_Server.swift
```diff
@@ -231,6 +231,46 @@ final class Test_Server: XCTestCase {
...
+                          "name" : "Fluffz"
+                        }
+                        """#
+                ),
+                .init()
+            )
+            XCTFail("The method should have thrown an error.")
+        } catch {
+            XCTAssertNotNil(error)
+        }
```
czechboy0, line 271:
```suggestion
        } catch {}
```

You can remove the non-nil check, as the `error` variable in the `catch` block is of type `Error`, not `Error?`.

Tests/PetstoreConsumerTests/Test_Server.swift
```diff
@@ -231,6 +231,46 @@ final class Test_Server: XCTestCase {
...
+                return .created(
+                    .init(
+                        headers: .init(
+                            X_Extra_Arguments: .init(code: 1)
+                        ),
+                        body: .json(
+                            .init(id: 1, name: "Fluffz")
+                        )
+                    )
+                )
```
czechboy0, line 247:
```suggestion
                XCTFail("The handler should not have been called")
```

czechboy0 (changes requested):
Two more suggestions, just to simplify the tests a bit more, otherwise looks good. 👍

takeshi-1000:
Thank you for checking ! I applied your suggestion  .

czechboy0 (approved):
Looks great, thank you!

czechboy0:

Tests/PetstoreConsumerTests/Test_Server.swift
```diff
@@ -231,6 +231,35 @@ final class Test_Server: XCTestCase {
         )
     }
 
+    
+    func testCreatePet_withIncorrectContentType() async throws {
+        client = .init(
+            createPetBlock: { input in
+                XCTFail("The handler should not have been called")
```
czechboy0, line 238: The CI is complaining about this, we need to return Never here, for example like this.
```suggestion
                XCTFail("The handler should not have been called")

                fatalError("Unreachable")
```

czechboy0 (changes requested):

Tests/PetstoreConsumerTests/Test_Server.swift
takeshi-1000: Oh sorry, after applied the suggestion, test looks passed locally.

takeshi-1000:
I missed https://github.com/apple/swift-openapi-generator/blob/main/CONTRIBUTING.md#run-scriptssoundnesssh, and am confirming .

simonjbeaumont (approved):
Thanks @takeshi-1000 for getting involved!

PR #141, opened 2023-07-25
Move ContentType.identifier logic into the translator

czechboy0:
### Motivation

In preparation for changing the mapping of content type names onto enum case names, we need to have the `ContentType.identifier` logic on a translator, which has access to the feature flags, so that we can later conditionalize the logic.

### Modifications

Takes the first step - moves the existing logic onto the translator.

### Result

NFC

### Test Plan

Added unit tests both for the existing logic, and also for the stubbed out logic behind the feature flag.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/Content/ContentSwiftName.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Content/ContentSwiftName.swift
@@ -1,0 +1,36 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftOpenAPIGenerator open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftOpenAPIGenerator project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import OpenAPIKit30
+
+extension FileTranslator {
+
+    /// Returns a Swift-safe identifier used as the name of the content
+    /// enum case.
+    ///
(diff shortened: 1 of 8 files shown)


Sources/_OpenAPIGeneratorCore/Translator/Content/ContentSwiftName.swift
```diff
@@ -0,0 +1,36 @@
...
+            switch contentType {
+            case .json:
+                return "json"
+            case .text:
+                return "text"
+            case .binary:
+                return "binary"
+            }
+        }
+    }
```
simonjbeaumont, line 35: I'm not sure I understand why this has been moved here. Couple of reasons:

- This PR is in aid of the _multiple_ content types but it is operating on a single `ContentType` as it was when it was a computed property on `ContentType`.
- This flow is no longer colocated with all the other logic, e.g. `headerValueForString`, which remain on the `ContentType`—I understand you need access to the config, but couldn't this be a method on `ContentType` that took a config?

czechboy0: We'll be changing all content type -> Swift identifier mangling in the multiple content type feature, not just when multiple content types are involved. Since it's also a breaking change (so is generating all content types instead of just the first one), I'm staging in these two changes together. But yeah, until the subsequent PRs come, this doesn't seem fully justified. (I just wanted to break things out for easier review.)

Regarding where the method is, the extensions of translators pattern is what we use everywhere else, so I followed it here. We could have it extended in ContentType.swift, but that feels a bit more like a layering violation, as ContentType is mostly just a currency type, and it knowing about types like Config seems to be the wrong direction.

Also, we'll be able to move this logic _back_ once we tag the next breaking version, as it won't require access to the Config anymore again. So it's a bit of a dance to stage this in safely, but it'll end up in the old location eventually.

simonjbeaumont (approved):

Issue #71, opened 2023-06-16
Support wildcard content types


knellr:
We have an OpenAPI spec that contains wildcard content types, especially for non-200 responses:
```
          "404" : {
            "content" : {
              "*/*" : {
                "schema" : {
                  "type" : "object"
                }
              }
            },
            "description" : "Not Found"
          },
```
 The client generated currently enforces these with the following generated code:
```
try converter.validateContentTypeIfPresent(
    in: response.headerFields,
    substring: "*/*"
)
```
As suggested, the code above is looking for a substring, causing the request to fail:
```
  underlyingError = unexpectedContentTypeHeader (unexpectedContentTypeHeader = "application/json")
```
My reading of the spec is that wildcards are supported (there are multiple examples including `*/*`), and my expectation would be that either the validation should be done using a wildcard, or validation should be relaxed.

czechboy0:
Thanks for reporting the issue - you're right, currently the generator doesn't handle content maps for bodies that contain wildcards in the content type names.

czechboy0:
Hi @knellr – I've been thinking about this issue and I'm still not sure what kind of Swift code we should generate for it.

Can you talk more about what contexts you use wildcards in content types in the OpenAPI document for? Especially when paired with the `schema` field, it needs to be a JSON content type for `Codable` types to be generated for you, so why not explicitly specify it as such?

I do see `*/*` as an example OpenAPI specification, they just don't clarify how it's meant to be handled by code generators.

Any info would be appreciated here, thanks! 🙏

knellr:
I've just been able to dig into this, and in our specific case the issue is that we use springdoc-openapi-ui and it defauts its mime type to `*/*` (https://springdoc.org/core-properties.html).

I'm not sure what a genuine use case would be here, but it seems we can work around this by annotating our backend fully or setting the `springdoc.default-produces-media-type` setting to `application/json`.

In practice, if I didn't have control of the backend, I think I'd expect the mime type to be validated correctly, and as a minimum to be given access to the actual mime-type and the underlying `Data`.

czechboy0:
Glad to see you were able to get things working.

> I'm not sure what a genuine use case would be here, but it seems we can work around this by annotating our backend fully or setting the springdoc.default-produces-media-type setting to application/json.

[...]

PR #146, opened 2023-07-27
[Generator] Multiple content types support

czechboy0:
Depends on https://github.com/apple/swift-openapi-runtime/pull/29

### Motivation

Up until now, when an OpenAPI document contained more than one content type in either a request or a response body, we would only generate code for now, and ignored the other ones.

However, that was a temporary workaround until we could add proper support for multiple content types, which is what this PR is about.

Addresses #6 and #7, except for the "Accept header" propagation, which will be addressed separately and have its own PR and even a proposal. That's one of the reasons this feature is landing disabled, hidden behind the `multipleContentTypes` feature flag.

A tip for reviewing this change: first check out how the test OpenAPI document and the reference files changed, that shows the details of how we encode and decode multiple content types. I also added comments in the code for easier review.

### Modifications

- Main: all supported content types found in a `content` map in request and response bodies are now generated in the `Body` enum.
- Added a method `supportedTypedContents` that returns all supported content types, used now instead of the `bestTypedContent` method we used before. What is a "supported" content type? One that has a schema that doesn't return `false` to `isSchemaSupported`. (This is a pre-existing concept.)
- Updated the logic for generating request and response bodies to use the new method, both on client and server.
- Updated content type -> Swift enum case name logic, but that will go through a full proposal, so please hold feedback for the proposal, this is just a first pass based on some collected data of most commonly used content types.
- 

### Open Questions
- What to do in an operation with multiple request/response content types when _no_ `content-type` header is provided? Previously, with up to 1 content type support, we still went ahead and tried to parse it. But here, we don't know which content to parse it as, so we just try it with the first in the list (we respect the order in the YAML dictionary, so what the user put first will be what we try to use when no `content-type` header is provided). If an invalid `content-type` value is provided, we throw an error (pre-existing behavior). Q: Is choosing the first reasonable? What would be better? Note that, unfortunately, many servers don't send `content-type` today.

### Result

When an adopter provides multiple content types in their OpenAPI document, we generate one case in the Body enum for each! However, the new functionality is disabled by default, has to be explicitly requested by enabling the `multipleContentTypes` feature flag either on the CLI or in the config file.

### Test Plan

Added `setStats` and `getStats` operations that employ multiple content types, one in a request body, the other in a response body. Added unit tests for it in `PetstoreConsumerTests` to verify it all works correctly at runtime.

Deleted `Tests/OpenAPIGeneratorCoreTests/Translator/RequestBody/Test_translateRequestBody.swift` as snippet tests do this job better, and I didn't want to spend the time updating the test.

Adds `PetstoreConsumerTests_FF_MultipleContentTypes`, a variant of `PetstoreConsumerTests` for when the `multipleContentTypes` feature flag is enabled, so we test both how the existing logic handles multiple content types (it picks one), and the new logic (it generates code for all supported ones it found).

## TODOs
- [x] rename `IfConditionPair` to `IfBranch`
- [x] break out the first `if` from the subsequence `else if`s in `IfStatement`, to 3 properties: first if, then an array of else ifs, then an optional else; to avoid creating invalid code
- [x] create an SPI type `ContentType` to avoid repeatedly parsing the received content type


.swift-format
```diff
@@ -43,7 +43,7 @@
     "OnlyOneTrailingClosureArgument" : true,
     "OrderedImports" : false,
     "ReturnVoidInsteadOfEmptyTuple" : true,
-    "UseEarlyExits" : true,
+    "UseEarlyExits" : false,
```
czechboy0, line 46: This is a case where the generated code is made a bit more confusing when this rule is applied there, but I think it makes sense for our code in general. So either we diverge and _not_ share the same config for generator and generated code (going against #40), or we disable this option for both as I did here for expediency. I'm fine with either.

Sources/_OpenAPIGeneratorCore/Layers/StructuredSwiftRepresentation.swift
```diff
@@ -664,6 +664,38 @@ struct SwitchDescription: Equatable, Codable {
     var cases: [SwitchCaseDescription]
 }
 
+/// A description of an if condition and the corresponding code block.
+///
+/// For example: in `if foo { bar }`, the condition pair represents
+/// `foo` + `bar`.
+struct IfConditionPair: Equatable, Codable {
```
czechboy0, line 671: I'm not loving the name, ideas welcome.

Sources/_OpenAPIGeneratorCore/Translator/Content/ContentInspector.swift
```diff
@@ -72,6 +144,7 @@ extension FileTranslator {
     ///   - map: The content map from the OpenAPI document.
     ///   - excludeBinary: A Boolean value controlling whether binary content
     ///   type should be skipped, for example used when encoding headers.
+    ///   - foundIn: The location where this content is parsed.
     /// - Returns: the detected content type + schema, nil if no supported
     /// schema found or if empty.
     func bestSingleContent(
```
czechboy0, line 150: Reason we're keeping this and not deprecating is because we still use it for parameters. We don't yet have plans for multiple content type support for parameters, so they continue using `bestSingleContent`.

Sources/_OpenAPIGeneratorCore/Translator/Content/ContentSwiftName.swift
```diff
@@ -21,7 +21,25 @@ extension FileTranslator {
...
+                return "text"
+            case "*/*":
+                return "any"
+            case "application/xml":
+                return "xml"
+            case "application/octet-stream":
+                return "binary"
+            default:
+                return swiftSafeName(for: rawMIMEType)
+            }
```
czechboy0, line 42: These special-cased names will go through a full proposal, this is here as my first draft, but please hold back feedback for the proposal.

Sources/_OpenAPIGeneratorCore/Translator/RequestBody/TypedRequestBody.swift
```diff
@@ -25,8 +25,8 @@ struct TypedRequestBody {
     /// A Boolean value indicating whether the response is inlined.
     var isInlined: Bool
 
-    /// The validated content.
-    var content: TypedSchemaContent
+    /// The validated contents.
+    var contents: [TypedSchemaContent]
```
czechboy0, line 29: This is the gist of the whole feature: going from 1 to N contents in a body.

Tests/PetstoreConsumerTests_FF_MultipleContentTypes/TestClient.swift
```diff
@@ -0,0 +1,112 @@
+//===----------------------------------------------------------------------===//
```
czechboy0, line 1: Everything in `Tests/PetstoreConsumerTests_FF_MultipleContentTypes` will be consolidated back into the main test bundle once we're ready to tag the next breaking version. Until then, we test both the feature flag off (default) and on (new behavior).

simonjbeaumont:
Before digging into the review proper, thought I'd pick up on this question:

> What to do in an operation with multiple request/response content types when no content-type header is provided? Previously, with up to 1 content type support, we still went ahead and tried to parse it. But here, we don't know which content to parse it as, so we just try it with the first in the list (we respect the order in the YAML dictionary, so what the user put first will be what we try to use when no content-type header is provided). If an invalid content-type value is provided, we throw an error (pre-existing behavior). Q: Is choosing the first reasonable? What would be better? Note that, unfortunately, many servers don't send content-type today.

I think it's reasonable for us to throw an error if `Content-Type` _is_ provided but doesn't match what we get. However, in the case the server omits `Content-Type`, shouldn't we be trying _all_ of them, not just the first, similar to how we handle `anyOf`/`oneOf`? Failing if it doesn't match the first seems to be overly restrictive since it will reject a response from the server which is "in spec"? Or have I misunderstood?

Sources/_OpenAPIGeneratorCore/Translator/Content/ContentInspector.swift
```diff
@@ -123,4 +196,62 @@ extension FileTranslator {
...
+        foundIn: String
+    ) -> SchemaContent? {
+        if contentKey.isJSON,
+            let contentType = ContentType(contentKey.typeAndSubtype)
+        {
+            diagnostics.emitUnsupportedIfNotNil(
+                contentValue.encoding,
+                "Custom encoding for JSON content",
+                foundIn: "\(foundIn), content \(contentKey.rawValue)"
+            )
```
simonjbeaumont, line 228: The `encoding` property in the `content` map should only be used in more specific cases[^1]. Should we be more precise about when we consider it?

> The encoding object SHALL only apply to requestBody objects when the media type is multipart or application/x-www-form-urlencoded.

[^1]: https://spec.openapis.org/oas/v3.0.3#mediaTypeObject

Sources/_OpenAPIGeneratorCore/Translator/RequestBody/TypedRequestBody.swift
```diff
@@ -95,12 +95,11 @@ extension FileTranslator {
...
-                inParent: typeName
-            )
-        else {
+        let contents = try supportedTypedContents(
+            request.content,
+            inParent: typeName
+        )
+        if contents.isEmpty {
             return nil
         }
```
simonjbeaumont, line 104: Took me a while to grok this, because I was expecting the feature flag switch to be here, but then realised that you fallback to the old behaviour in the _new_ function, based on the feature flag.

What was the reasoning behind this? Could we move it here, which would also allow the new function, that does multiple content types, to be simplified to always do the new thing?

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
```diff
@@ -616,7 +616,79 @@ final class SnippetBasedReferenceTests: XCTestCase {
         )
     }
 
-    func testComponentsResponsesResponseWithOptionalHeader() throws {
+    func testComponentsResponsesResponseMultipleContentTypes() throws {
```
simonjbeaumont, line 619: Thanks for adding these tests!

simonjbeaumont (commented):
Thanks for taking the time to rebase this PR for easier review. It's looking really good. I've added a few questions, otherwise I think it's looking really good!

Sources/_OpenAPIGeneratorCore/Translator/Content/ContentInspector.swift
czechboy0: Possibly. This code only emits a diagnostics if it appears in the OpenAPI document, but we don't support the `encoding` map yet. Are you saying we should only emit the diagnostic specifically if the content type is `multipart` or `application/x-www-form-urlencoded`, and not emit the diagnostic if it's provided with other content types?

Sources/_OpenAPIGeneratorCore/Translator/Content/ContentInspector.swift
```diff
@@ -61,6 +61,78 @@ extension FileTranslator {
...
+        )
+        return try contents.compactMap { content in
+            guard
+                try validateSchemaIsSupported(
+                    content.schema,
+                    foundIn: parent.description
+                )
+            else {
+                return nil
+            }
```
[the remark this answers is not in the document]
czechboy0: Yes, each emits a separate diagnostic. 1 emits `Unsupported content` and 2 emits `Unsupported schema`.

Sources/_OpenAPIGeneratorCore/Layers/StructuredSwiftRepresentation.swift
czechboy0: I think we've been breaking out types like this whenever a larger wrapper contains a collection of smaller items. It makes the renderer easier to test, as we have `rendered...` function for each of these types.

And yeah, "branch" is a better term – how about `IfBranch` or `IfConditionBranch`?

Sources/_OpenAPIGeneratorCore/Layers/StructuredSwiftRepresentation.swift
```diff
@@ -1202,6 +1248,23 @@ extension Expression {
...
 
+    /// Returns an if statement, with optional else if's and an else
+    /// statement attached.
+    /// - Parameters:
+    ///   - branches: The conditional branches.
+    ///   - elseBody: The body of an else block.
+    static func ifStatement(
+        branches: [IfConditionPair],
+        elseBody: [CodeBlock]? = nil
+    ) -> Self {
```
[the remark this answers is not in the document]
czechboy0: Sure, I can break it out.

Tests/OpenAPIGeneratorReferenceTests/Resources/Docs/petstore.yaml
```diff
@@ -113,6 +113,31 @@ paths:
...
+    post:
+      operationId: postStats
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/PetStats'
+          text/plain: {}
+          application/octet-stream: {}
```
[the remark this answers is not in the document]
czechboy0: Yup, tracked by #71.

Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore_FF_MultipleContentTypes/Client.swift
```diff
@@ -256,7 +256,7 @@ public struct Client: APIProtocol {
                 try converter.setHeaderFieldAsText(
                     in: &request.headerFields,
                     name: "accept",
-                    value: "application/json"
+                    value: "application/json, text/plain, application/octet-stream"
```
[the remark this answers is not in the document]
czechboy0: > I realise this might change later, but, to confirm, the plan here is to send all content types the client knows about—that is, all those in the OpenAPI doc—in the accept header?

Yes, starting with this PR, we send all the _concrete_ _supported_ types in the _request_'s `Accept` header that we know might be coming in the body of the _response_. Note the _supported_ here, so if there's a content with an unsupported schema, we will not send its MIME type, as we'd fail to deserialize it anyway. We could change this to send all the content types found in the OpenAPI doc for the response, but it'd still fail for the user at response parsing time, so I'm happy to go with either option, the user will end up in the same place.

> Looking at the OAS, it appears there's some specified precedence when there's overlap:

> > REQUIRED. The content of the request body. The key is a media type or [media type range](https://tools.ietf.org/html/rfc7231#appendix-D) and the value describes it. For requests that match multiple keys, only the most specific key is applicable. e.g. text/plain overrides text/*

This is about request body content, and I'd consider the overlap covered by #71, as there cannot be overlap between concrete content types. But for sure we'll have to implement this prioritization logic once we're doing #71.

gjcairo:
> Q: Is choosing the first reasonable? What would be better?

Answering this question from the description: I think this is reasonable. The only alternative is to send supported ones, I guess, but I'm not sure what benefit this would have over sending just the first one.

Sources/_OpenAPIGeneratorCore/Translator/Content/ContentInspector.swift
czechboy0: > The encoding can be specified by a user using an encoding field in their document[1](https://github.com/apple/swift-openapi-generator/pull/146#user-content-fn-1-f66cc33fda18976965e1b0f07becbdf6). I don't think we support that today (correct me if I'm wrong)—are we tracking it?

We don't support the `encoding` feature today, I don't believe we even have an issue for it yet (probably worth filing, and if we can find an example of it in the wild, even better).

> I'm not convinced we're following that precisely—again this would be a different issue, separate from this PR.

Yeah, I suspect you're right. Up until now, the `bestSingleContent` was mainly created for body contents, which will stop using it when multiple body content types lands, at which point only parameters and response headers will use it. Then we could probably look into supporting `encoding` and removing any heuristics we use today for parameters/response headers.

Tests/PetstoreConsumerTests/Test_Client.swift
```diff
@@ -342,6 +342,120 @@ final class Test_Client: XCTestCase {
...
+                ],
+                encodedBody: #"""
+                    count_is_1
+                    """#
+            )
+        }
+        do {
+            _ = try await client.getStats(.init())
+            XCTFail("Should have thrown an error")
+        } catch {}
```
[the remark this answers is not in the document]
czechboy0: It's too much work and produces a lot of churn when it changes, so while we do have some test assertions for exact errors being thrown, in many places we don't bother. We don't want all tests to be too tightly coupled to the implementation details they're testing.

Sources/_OpenAPIGeneratorCore/Translator/RequestBody/translateRequestBody.swift
```diff
@@ -16,23 +16,18 @@ import OpenAPIKit30
...
     ) throws -> [Declaration] {
-        let content = requestBody.content
         let decl = try translateSchema(
             typeName: content.resolvedTypeUsage.typeName,
             schema: content.content.schema,
-            overrides: .init(
-                isOptional: !requestBody.request.required,
-                userDescription: requestBody.request.description
-            )
+            overrides: .none
```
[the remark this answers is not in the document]
czechboy0: Good question – I think it was wrong/not working before! We didn't have any tests for it in our reference suite, and it doesn't make sense now as a single request body might have multiple content values in it. So it probably wasn't supposed to be there in the first place, and got introduced through some copy pasting of code from parameters/headers, etc.

gjcairo (approved):

Sources/_OpenAPIGeneratorCore/Layers/StructuredSwiftRepresentation.swift
czechboy0: Ended up going for `IfBranch`.

Sources/_OpenAPIGeneratorCore/Layers/StructuredSwiftRepresentation.swift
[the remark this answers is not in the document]
czechboy0: Updated.

Sources/_OpenAPIGeneratorCore/Translator/Content/ContentInspector.swift
czechboy0: I'm of two minds. If the adopter has an `encoding` map for a content type that doesn't support it, they should probably fix their OpenAPI doc, in that sense, our diagnostic will actually _help_ them do that. On the other hand, it can also make it sound like we don't support a feature that doesn't exist in the OpenAPI specification itself. Since we haven't thought much about how the `encoding` feature should be implemented, I don't feel I have enough info to make a definite call just yet, but I'd be leaning towards helping the adopter fix up their doc, as they might otherwise be surprised why the feature isn't working, and no diagnostic is emitted.

simonjbeaumont: > but I'd be leaning towards helping the adopter fix up their doc

Hm, I thought our stance was that we'd try and keep the logic in the translator only concerned with handling valid OpenAPI documents.

IMO the right place for helping an adopter understand why generation is failing is by having this emitted at the start of the pipeline, in the preflight validation.

If the generation doesn't go as expected (i.e. it crashes or the generated code has things missing) then I'd be looking at the warnings, and draw this conclusion?

czechboy0: Yeah that's fair – so I'll update the code to only emit the diagnostic for these specific content types.

czechboy0: Fixed: https://github.com/apple/swift-openapi-generator/pull/146/commits/556fcb62986cea6c9dd335f76e77bdd8c02608b9

Sources/_OpenAPIGeneratorCore/Translator/RequestBody/TypedRequestBody.swift
simonjbeaumont: OK, I don't love it but am happy enough.

czechboy0: Yeah, the SPI functions on Converter aren't designed with the same goals as normal APIs, so sometimes they do more than one thing. In normal code, you don't try to optimize for avoiding any repetitive code in your caller's code, but here we do, to keep the generated code as high signal with things that come from the OpenAPI document, and minimize any constant, "stamped out" code that doesn't change between OpenAPI documents. That means we have some of these "omnibus" methods that we normally wouldn't write by hand.

But even today, updating the reference suite is a lot of repetitive work, so I think it makes sense for us to push as much logic as we can into the Converter functions in runtime, instead of generating snippets of logic into the adopter's target.

simonjbeaumont (approved):

Issue #128, opened 2023-07-18
Support Path Item references


petershaw:
A file that works in openapi generator does not work in openapi-generator. The section that causes the error is: 
```yaml
paths:
  /documents:
    $ref: './paths/documents.yaml#/documents'
  /documents/{document-ident}:
    $ref: './paths/documents.yaml#/documents-detail'
```
in the root file. 

Error: 
```
error: Inconsistency encountered when parsing `Vendor Extension` under the `/documents` path: Found at least one vendor extension property that does not begin with the required 'x-' prefix. Invalid properties: [ $ref ].
```

We always have our paths in separate files for various reasons. This never was a problem before so I think swift-openapi-generator should implement it. ;)

czechboy0:
cc @mattpolzin as I suspect this support would first need to be added in OpenAPIKit

czechboy0:
This is specifically about Path Item references, not external references, which we're discussing separately in #.

mattpolzin:
Tracking this as a bug here: https://github.com/mattpolzin/OpenAPIKit/issues/280

mattpolzin:
This bug should be fixed with the new alpha release. Note the breaking changes described on the Release page: https://github.com/mattpolzin/OpenAPIKit/releases/tag/3.0.0-alpha.9

czechboy0:
Amazing, thank you @mattpolzin. I marked this issue as unblocked now, for anyone who wants to pick it up.

czechboy0:
Fixed in #147.

czechboy0:
To clarify, in 3.0 there isn't much we can do, as we don't yet support external references, which is the only way to have a Path Item reference.

In 3.1, there is work to do, but we don't support that yet. Maybe that's what we should use this issue to track, let me reopen and retitle.

dannflor:
Would like to see this happen soon. Now that I'm trying to generate from multiple documents to work around #551 it's non-negotiable that I need to be able to share schemas between the documents and external `$ref` is the only way to do that without pre-build tooling and a lot of pain.

czechboy0:
Hi @dannflor, note that this issue is only about Path references. If you're looking for an issue tracking splitting your document into multiple documents that reference each other, that's https://github.com/apple/swift-openapi-generator/issues/25.

simonjbeaumont:
> it's non-negotiable

[...]

PR #147, opened 2023-07-27
Update to OpenAPIKit 3.0.0-alpha.9

simonjbeaumont:
### Motivation

Investigation into #128 led to a new release of OpenAPIKit, which supports path references. Supporting path references isn't something we want to do right now because:

1. OpenAPI 3.0.3 only supports external path references (cf. 3.1, which supports internal references too).
2. Swift OpenAPI Generator currently only supports OpenAPI 3.0.x.
3. Swift OpenAPI Generator currently doesn't support external references.

The OpenAPIKit update comes with a new API, because the paths are now an `Either` to support being a `PathItem` or a reference to a `PathItem`.

For now we'd like to be keeping up to date with OpenAPIKit so that, when we support OpenAPI 3.1 and/or external references, we have the APIs we need in the upstream package.

### Modifications

- Update to OpenAPIKit 3.0.0-alpha.9.
- Attempt to resolve the path item during translation.

### Result

An error will be thrown when using an OpenAPI document with a path reference.

### Test Plan

Added reference tests.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateAllAnyOneOf.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateAllAnyOneOf.swift
@@ -55,10 +55,10 @@ extension FileTranslator {
                 switch type {
                 case .allOf:
                     // AllOf uses all required properties.
-                    propertyType = rawPropertyType
+                    propertyType = rawPropertyType.withOptional(false)
                 case .anyOf:
                     // AnyOf uses all optional properties.
-                    propertyType = rawPropertyType.asOptional
+                    propertyType = rawPropertyType.withOptional(true)
                 }
                 let comment: Comment? = .property(
                     originalName: key,
--- a/Package.swift
+++ b/Package.swift
@@ -64,7 +64,7 @@ let package = Package(
         // Read OpenAPI documents
         .package(
             url: "https://github.com/mattpolzin/OpenAPIKit.git",
-            exact: "3.0.0-alpha.7"
+            exact: "3.0.0-alpha.9"
         ),
         .package(
             url: "https://github.com/jpsim/Yams.git",
--- a/Sources/_OpenAPIGeneratorCore/Translator/Operations/OperationDescription.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Operations/OperationDescription.swift
@@ -59,13 +59,20 @@ extension OperationDescription {
     ///   - components: The components from the OpenAPI document.
     ///   - asSwiftSafeName: A converted function from user-provided strings
     ///   to strings safe to be used as a Swift identifier.
+    /// - Throws: if `map` contains any references; see discussion for details.
+    ///
+    /// This function will throw an error if `map` contains any references, because:
+    /// 1. OpenAPI 3.0.3 only supports external path references (cf. 3.1, which supports internal references too)
+    /// 2. Swift OpenAPI Generator currently only supports OpenAPI 3.0.x.
+    /// 3. Swift OpenAPI Generator currently doesn't support external references.
     static func all(
         from map: OpenAPI.PathItem.Map,
         in components: OpenAPI.Components,
         asSwiftSafeName: @escaping (String) -> String
-    ) -> [OperationDescription] {
-        map.flatMap { path, value in
-            value.endpoints.map { endpoint in
+    ) throws -> [OperationDescription] {
+        try map.flatMap { path, value in
+            let value = try value.resolve(in: components)
+            return value.endpoints.map { endpoint in
                 OperationDescription(
                     path: path,
                     endpoint: endpoint,
--- a/Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/TypesFileTranslator.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/TypesFileTranslator.swift
@@ -41,19 +41,17 @@ struct TypesFileTranslator: FileTranslator {
             + config.additionalImports
             .map { ImportDescription(moduleName: $0) }
 
-        let apiProtocol = translateAPIProtocol(doc.paths)
+        let apiProtocol = try translateAPIProtocol(doc.paths)
 
         let serversDecl = translateServers(doc.servers)
 
         let components = try translateComponents(doc.components)
 
-        let operationDescriptions =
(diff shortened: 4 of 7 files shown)


czechboy0 (approved):
lgtm!

simonjbeaumont:
Interesting test failures:

> ```
> Test Case 'Test_YamsParser.testEmitsComplexOpenAPIParsingError' started at 2023-08-01 10:26:00.831
> /code/Tests/OpenAPIGeneratorCoreTests/Parser/Test_YamsParser.swift:123: error: Test_YamsParser.testEmitsComplexOpenAPIParsingError : XCTAssertEqual failed: ("/foo.yaml: error: Found neither a $ref nor a PathItem in Document.paths['/system']. 
> 
> PathItem could not be decoded because:
> Expected to find `responses` key for the **GET** endpoint under `/system` but it is missing..") is not equal to ("/foo.yaml: error: Expected to find `responses` key for the **GET** endpoint under `/system` but it is missing.") -
> Test Case 'Test_YamsParser.testEmitsComplexOpenAPIParsingError' failed (0.008 seconds)
> ```

This one is easy—just an error message changed in OpenAPIKit.

> ```
> ============================================================
> Performing reference test
> --- begin test config --------------------------------------
> docFilePath: Docs/petstore.yaml
> mode: types
> additionalImports: []
> featureFlags: []
> referenceOutputDirectory: ReferenceSources/Petstore
> 
> --- end test config ----------------------------------------
> /code/Tests/OpenAPIGeneratorReferenceTests/FileBasedReferenceTests.swift:108: error: FileBasedReferenceTests.testPetstore : failed - 
> Directory contents not equal:
>   File under test: Types.swift
>   Reference file: ReferenceSources/Petstore/Types.swift
>   Diff output: (see below)
> --- begin diff output --------------------------------------
> diff --git a/ReferenceSources/Petstore/Types.swift b/tmp/C40D09CF-8552-4A43-A718-A5E9EAD097F6/Types.swift
> index 06c6d29..5f18bc9 100644
> --- a/ReferenceSources/Petstore/Types.swift
> +++ b/tmp/C40D09CF-8552-4A43-A718-A5E9EAD097F6/Types.swift
> @@ -169,16 +169,16 @@ public enum Components {
>              /// Extra information about the error.
>              ///
>              /// - Remark: Generated from `#/components/schemas/Error/extraInfo`.
>              public struct extraInfoPayload: Codable, Equatable, Hashable, Sendable {
>                  /// - Remark: Generated from `#/components/schemas/Error/extraInfo/value1`.
> -                public var value1: Components.Schemas.ExtraInfo
> +                public var value1: Components.Schemas.ExtraInfo?
>                  /// Creates a new `extraInfoPayload`.
>                  ///
>                  /// - Parameters:
>                  ///   - value1:
> -                public init(value1: Components.Schemas.ExtraInfo) { self.value1 = value1 }
> +                public init(value1: Components.Schemas.ExtraInfo? = nil) { self.value1 = value1 }
>                  public init(from decoder: any Decoder) throws { value1 = try .init(from: decoder) }
>                  public func encode(to encoder: any Encoder) throws {
>                      try value1.encode(to: encoder)
>                  }
>              }
> 
> --- end diff output ----------------------------------------
> ```

This one is more interesting... 

This is converting from this OpenAPI:

```yaml
        extraInfo:
          description: Extra information about the error.
          allOf:
            - $ref: '#/components/schemas/ExtraInfo'
```


But an `allOf` must have "all of" the definitions present IIUC:

> The OpenAPI Specification allows combining and extending model definitions using the `allOf` property of JSON Schema, in effect offering model composition. `allOf` takes an array of object definitions that are validated independently but together compose a single object.

This will need looking into further.

simonjbeaumont:
> This will need looking into further.

OK, so I've created some tests which isolate the issue. They pass without the OpenAPIKit upgrade, and fail with it. Doesn't necessarily mean it's a bug in OpenAPIKit—we could be making some incorrect assumptions.

I'm going to rebase this PR with those commits, one at a time (tests first, then upgrade), to show the issue.

Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateAllAnyOneOf.swift
```diff
@@ -55,10 +55,10 @@ extension FileTranslator {
                 switch type {
                 case .allOf:
                     // AllOf uses all required properties.
-                    propertyType = rawPropertyType
+                    propertyType = rawPropertyType.withOptional(false)
                 case .anyOf:
                     // AnyOf uses all optional properties.
-                    propertyType = rawPropertyType.asOptional
+                    propertyType = rawPropertyType.withOptional(true)
```
simonjbeaumont, line 61: @czechboy0 this ended up being the fix. I didn't dig deeply enough to determine why this is now required with the OpenAPIKit update, but I think the logic is correct.

czechboy0: I like it, it feels right - `allOf` is all subschemas being required, `anyOf` is all subschemas being optional.

czechboy0 (approved):
:shipit:

Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateAllAnyOneOf.swift
mattpolzin: Woops. Sorry this caused some confusion. In hindsight I should have called this change out more loudly than just "fix allOf schema simplification" in the release notes. 

Your interpretation of `allOf` makes sense in that "all of these things are true" but OpenAPIKit just so happened to "fix" things such that "if a property is required, allOf its schemas must be required; if a property is optional, allOf its schemas must be optional" which in a sense is also true because if whatever object comprises those combined schemas is omitted then certainly the individual schemas were not required to be there. There's no right interpretation here since the components of `allOf` do form an informal product type but they also don't form concrete discrete sub-objects but rather all represent the same single object. It may be more accurate to say they don't have identities of their own and are neither required nor not in and of themselves, but that might just be the philosopher in me 😁.

Details that you may not care to know:
OpenAPIKit actually has some built-in (optional) "simplification" logic that removes `allOf` schemas by literally building a new schema that combines them and that logic defaults the combined schema to optional until a required entry is hit -- it only takes one required entry to "elevate" the combined schema to being required. Anyway, in fixing that logic I introduced a breaking change I didn't think to call out in the release notes.

PR #150, opened 2023-07-29
Calculate required reponse header parameters for initialize

### LINKED ISSUE #30: Recurse when calculating whether an init param is required
Having even one optional header means we have to explicitly provide them when constructing a response - that's not great.

For example:

```swift
return .ok(.init(body: .json(.init(foo: "bar"))))
```

Now, even when we don't want to provide any headers, we'll have to provide the param.

This should only happen when there is at least one _required_ header.

takeshi-1000:
### Motivation

Resolve #30

### Modifications

To ensure the initialization of response headers when there is at least one required response header, 
I added private method for calculate necessory parameters, and applied it to the Structure Declalation returned by the `translateStructBlueprint` method.

With the current `PropertyBlueprint` structure, it seems difficult to address this issue.
This is because the `defaultValue` of the PropertyBlueprint is used for member variables and initialization variables, and the method of generating the Structure Blueprint seems to be common. 

There may be a way to reconsider the PropertyBlueprint structure, but considering the impact and current needs, adopted the implementation as a private method once. It seems to be a complicated implementation, so if you have any other suggestions please let me know. 

### Result

If one or more required request header parameters are present,
Generates code that needs to set request header parameters when create response.

For example you define the yaml below

```yaml
      responses:
        '200':
          description: A success response with a greeting.
          headers:
            X-Extra-Arguments:
              required: false
              description: "A description here."
              content:
                application/json:
                  schema:
                    $ref: '#/components/schemas/Greeting'
```

Types.swift is Generated like blow, and you should not set response headers when initialize

スクリーンショット 2023-07-29 14 44 45

If you set required true
```yaml
            X-Extra-Arguments:
              required: true

```

Types.swift is Generated like blow, and you should set reponse headers when initialize

スクリーンショット 2023-07-29 14 52 55

### Test Plan

Modify tests to verify whether code generation is correct and generated code can be executed in cases of mandatory parameters and optional parameters

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
@@ -102,20 +102,23 @@ extension TypesFileTranslator {
             asSwiftSafeName: swiftSafeName
         )
 
-        let responseStructDecl = translateStructBlueprint(
-            .init(
-                comment: nil,
-                access: config.access,
-                typeName: typeName,
-                conformances: Constants.Operation.Output.Payload.conformances,
-                properties: [
-                    headersProperty,
-                    contentProperty,
-                ]
-            )
+        let structBlueprint: StructBlueprint = .init(
+            comment: nil,
+            access: config.access,
+            typeName: typeName,
+            conformances: Constants.Operation.Output.Payload.conformances,
+            properties: [
+                headersProperty,
+                contentProperty,
+            ]
         )
 
-        return responseStructDecl
+        let responseStructDecl = _calculateRequiredHeadersForInitialize(
+            with: headers,
+            from: translateStructBlueprint(structBlueprint)
+        )
+
+        return responseStructDecl.deprecate(if: structBlueprint.isDeprecated)
     }
 
     /// Returns a list of declarations for the specified reusable response
@@ -137,5 +140,83 @@ extension TypesFileTranslator {
             typeName: typeName,
             response: response
         )
+    }
+
+    /// Calculate the necessary parameters for initializing the response headers
+    /// and return the list of specified reusable response declarations.
+    /// - Parameters:
+    ///   - headers: the typed response headers
+    ///   - responseStructDecl:　A structure declaration before calculate.
+    /// - Returns: A structure declaration.
+    private func _calculateRequiredHeadersForInitialize(
+        with headers: [TypedResponseHeader],
+        from responseStructDecl: Declaration
+    ) -> Declaration {
+        guard case .commentable(let comment, let structDec) = responseStructDecl,
+            case .struct(let structDescription) = structDec
+        else {
+            return responseStructDecl
+        }
+
+        let newMembers: [Declaration] = structDescription
+            .members
+            .reduce(into: [Declaration]()) { partialResult, member in
+                let labelHeaders = "headers"
+
+                if case .commentable(let memberComment, let memberDecl) = member,
+                    case .function(let memberFuncDescription) = memberDecl,
+                    memberFuncDescription.signature.kind == .initializer,
+                    memberFuncDescription.signature.parameters.first(where: { $0.label == labelHeaders }) != nil
+                {
+
+                    let initParameters: [ParameterDescription] = memberFuncDescription
+                        .signature
+                        .parameters
+                        .map { parameterDesc in
+                            guard parameterDesc.label == labelHeaders else {
+                                return parameterDesc
+                            }
+                            let defaultValue: Expression? = {
+                                if headers.isEmpty {
+                                    return PropertyBlueprint.DefaultValue.emptyInit.asExpression
+                                }
+                                if headers.first(where: { !$0.isOptional }) != nil {
+                                    return nil
+                                }
+                                return PropertyBlueprint.DefaultValue.emptyInit.asExpression
+                            }()
+                            return .init(
+                                label: parameterDesc.label,
+                                name: parameterDesc.name,
+                                type: parameterDesc.type,
+                                defaultValue: defaultValue
+                            )
+                        }
+
+                    let initDescription: FunctionDescription = .init(
+                        accessModifier: memberFuncDescription.signature.accessModifier,
+                        kind: memberFuncDescription.signature.kind,
+                        parameters: initParameters,
+                        keywords: memberFuncDescription.signature.keywords,
+                        returnType: memberFuncDescription.signature.returnType,
+                        body: memberFuncDescription.body
+                    )
+
+                    partialResult.append(
+                        .commentable(memberComment, .function(initDescription))
+                    )
+                } else {
+                    partialResult.append(member)
+                }
+            }
+
+        let structDescriptionWithCalc = StructDescription(
+            accessModifier: structDescription.accessModifier,
+            name: structDescription.name,
+            conformances: structDescription.conformances,
+            members: newMembers
+        )
+
+        return .commentable(comment, .struct(structDescriptionWithCalc))
     }
 }
--- a/Tests/OpenAPIGeneratorReferenceTests/Resources/Docs/petstore.yaml
+++ b/Tests/OpenAPIGeneratorReferenceTests/Resources/Docs/petstore.yaml
@@ -107,6 +107,13 @@ paths:
                 application/json:
                   schema:
                     $ref: '#/components/schemas/CodeError'
+            X-Extra-Arguments2:
+              required: true
+              description: "A description here."
+              content:
+                application/json:
+                  schema:
+                    $ref: '#/components/schemas/CodeError'
           content:
             application/json:
               schema:
--- a/Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Client.swift
+++ b/Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Client.swift
@@ -174,6 +174,11 @@ public struct Client: APIProtocol {
                             in: response.headerFields,
                             name: "X-Extra-Arguments",
                             as: Components.Schemas.CodeError.self
+                        ),
+                        X_Extra_Arguments2: try converter.getRequiredHeaderFieldAsJSON(
+                            in: response.headerFields,
+                            name: "X-Extra-Arguments2",
+                            as: Components.Schemas.CodeError.self
                         )
                     )
                     try converter.validateContentTypeIfPresent(
--- a/Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Server.swift
+++ b/Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Server.swift
@@ -220,6 +220,11 @@ fileprivate extension UniversalServer where APIHandler: APIProtocol {
                         name: "X-Extra-Arguments",
                         value: value.headers.X_Extra_Arguments
                     )
+                    try converter.setHeaderFieldAsJSON(
+                        in: &response.headerFields,
+                        name: "X-Extra-Arguments2",
+                        value: value.headers.X_Extra_Arguments2
+                    )
                     try converter.validateAcceptIfPresent(
                         "application/json",
                         in: request.headerFields
--- a/Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Types.swift
+++ b/Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Types.swift
@@ -686,7 +686,7 @@ public enum Components {
             ///   - headers: Received HTTP response headers
             ///   - body: Received HTTP response body
             public init(
-                headers: Components.Responses.ErrorBadRequest.Headers,
+                headers: Components.Responses.ErrorBadRequest.Headers = .init(),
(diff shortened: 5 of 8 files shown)


Tests/OpenAPIGeneratorReferenceTests/Resources/Docs/petstore.yaml
```diff
@@ -107,6 +107,13 @@ paths:
                 application/json:
                   schema:
                     $ref: '#/components/schemas/CodeError'
+            X-Extra-Arguments2:
+              required: true
+              description: "A description here."
+              content:
+                application/json:
+                  schema:
+                    $ref: '#/components/schemas/CodeError'
```
czechboy0, line 116: I don't think we need this change to the integration test project, `listPets/My-Response-UUID` already is an example of a required response header.

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
```diff
@@ -519,7 +519,7 @@ final class SnippetBasedReferenceTests: XCTestCase {
         )
     }
 
-    func testComponentsResponsesResponseWithHeader() throws {
+    func testComponentsResponsesResponseWithOptionalHeader() throws {
```
czechboy0, line 522: These changes are great, covering both in snippet tests gives us the confidence it works correctly without having to expand the file-based reference test.

Tests/PetstoreConsumerTests/Test_Client.swift
```diff
@@ -192,6 +193,7 @@ final class Test_Client: XCTestCase {
             return
         }
         XCTAssertEqual(value.headers.X_Extra_Arguments, .init(code: 1))
+        XCTAssertEqual(value.headers.X_Extra_Arguments2, .init(code: 9999))
```
czechboy0, line 196: Let's undo these changes, as per the comment in `petstore.yaml`.

Tests/PetstoreConsumerTests/Test_Server.swift
```diff
@@ -166,7 +166,7 @@ final class Test_Server: XCTestCase {
         XCTAssertEqual(
             response.headerFields,
             [
-                .init(name: "X-Extra-Arguments", value: #"{"code":1}"#),
+                .init(name: "X-Extra-Arguments2", value: #"{"code":1}"#),
```
czechboy0, line 169: Let's undo these changes, as per the comment in `petstore.yaml`.

Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Types.swift
```diff
@@ -985,12 +985,18 @@ public enum Operations {
...
                     /// - Parameters:
                     ///   - X_Extra_Arguments:
-                    public init(X_Extra_Arguments: Components.Schemas.CodeError? = nil) {
+                    ///   - X_Extra_Arguments2:
+                    public init(
+                        X_Extra_Arguments: Components.Schemas.CodeError? = nil,
+                        X_Extra_Arguments2: Components.Schemas.CodeError
+                    ) {
                         self.X_Extra_Arguments = X_Extra_Arguments
+                        self.X_Extra_Arguments2 = X_Extra_Arguments2
```
czechboy0, line 999: Let's undo these changes, as per the comment in `petstore.yaml`.

Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Server.swift
```diff
@@ -220,6 +220,11 @@ fileprivate extension UniversalServer where APIHandler: APIProtocol {
                         name: "X-Extra-Arguments",
                         value: value.headers.X_Extra_Arguments
                     )
+                    try converter.setHeaderFieldAsJSON(
+                        in: &response.headerFields,
+                        name: "X-Extra-Arguments2",
+                        value: value.headers.X_Extra_Arguments2
+                    )
```
czechboy0, line 227: Let's undo these changes, as per the comment in `petstore.yaml`.

Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Client.swift
```diff
@@ -174,6 +174,11 @@ public struct Client: APIProtocol {
                             in: response.headerFields,
                             name: "X-Extra-Arguments",
                             as: Components.Schemas.CodeError.self
+                        ),
+                        X_Extra_Arguments2: try converter.getRequiredHeaderFieldAsJSON(
+                            in: response.headerFields,
+                            name: "X-Extra-Arguments2",
+                            as: Components.Schemas.CodeError.self
```
czechboy0, line 181: Let's undo these changes, as per the comment in `petstore.yaml`.

czechboy0 (changes requested):
Hi @takeshi-1000, thanks for picking this up!

I think the fix can be achieved using a little more general method, which will be usable in more places than just response headers.

Step 1: Define a computed property on `StructBlueprint` that tells you whether the struct has an empty initializer. For example:

```swift
extension StructBlueprint {
    
    /// A Boolean value indicating whether the struct can be initialized using
    /// an empty initializer.
    ///
    /// For example, when all the properties of the struct have a default value,
    /// the struct can be initialized using `Foo()`. This is important for
    /// other types referencing this type.
    var hasEmptyInit: Bool {
        // If at least one property requires an explicit value, this struct
        // cannot have an empty initializer.
        !properties.contains(where: { $0.defaultValue == nil })
    }
}
```

Step 2: Use this computed property instead of checking if headers property array is empty (like we do today).

I haven't tested this fully, but my local diff when I tried it looked something like:

```diff
diff --git a/Sources/_OpenAPIGeneratorCore/Translator/CommonTypes/StructBlueprint.swift b/Sources/_OpenAPIGeneratorCore/Translator/CommonTypes/StructBlueprint.swift
index 244c269..55ffffa 100644
--- a/Sources/_OpenAPIGeneratorCore/Translator/CommonTypes/StructBlueprint.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/CommonTypes/StructBlueprint.swift
@@ -80,6 +80,21 @@ struct StructBlueprint {
     var properties: [PropertyBlueprint]
 }
 
+extension StructBlueprint {
+
+    /// A Boolean value indicating whether the struct can be initialized using
+    /// an empty initializer.
+    ///
+    /// For example, when all the properties of the struct have a default value,
+    /// the struct can be initialized using `Foo()`. This is important for
+    /// other types referencing this type.
+    var hasEmptyInit: Bool {
+        // If at least one property requires an explicit value, this struct
+        // cannot have an empty initializer.
+        !properties.contains(where: { $0.defaultValue == nil })
+    }
+}
+
 /// A structure that contains the information about an OpenAPI object property
 /// that is required to generate a matching Swift property.
 struct PropertyBlueprint {
diff --git a/Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift b/Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
index da2eada..60c7383 100644
--- a/Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
@@ -36,20 +36,21 @@ extension TypesFileTranslator {
         let headerProperties: [PropertyBlueprint] = try headers.map { header in
             try parseResponseHeaderAsProperty(for: header)
         }
+        let headersStructBlueprint: StructBlueprint = .init(
+            comment: nil,
+            access: config.access,
+            typeName: headersTypeName,
+            conformances: Constants.Operation.Output.Payload.Headers.conformances,
+            properties: headerProperties
+        )
         let headersStructDecl = translateStructBlueprint(
-            .init(
-                comment: nil,
-                access: config.access,
-                typeName: headersTypeName,
-                conformances: Constants.Operation.Output.Payload.Headers.conformances,
-                properties: headerProperties
-            )
+            headersStructBlueprint
         )
         let headersProperty = PropertyBlueprint(
             comment: .doc("Received HTTP response headers"),
             originalName: Constants.Operation.Output.Payload.Headers.variableName,
             typeUsage: headersTypeName.asUsage,
-            default: headerProperties.isEmpty ? .emptyInit : nil,
+            default: headersStructBlueprint.hasEmptyInit ? .emptyInit : nil,
             associatedDeclarations: [
                 headersStructDecl
             ],
```

See if the above actually works with your snippet-based tests (which are great, thanks for adding them!). The file-based reference tests can be reverted, I don't think we need to expand them here, the snippet tests cover this well.

takeshi-1000:
@czechboy0 
Thank you for your detailed suggestions!
I didn't come up with it, but it looks reasonable so I applied it.
I also changed the test

czechboy0:
Awesome, thank you very much @takeshi-1000! Can you just update the PR description to reflect the changes now, and otherwise this should be ready to go? 🙏

takeshi-1000:
@czechboy0 
I updated, could you check please?

czechboy0 (approved):

Sources/_OpenAPIGeneratorCore/Translator/CommonTypes/StructBlueprint.swift
```diff
@@ -80,6 +80,21 @@ struct StructBlueprint {
...
+    /// A Boolean value indicating whether the struct can be initialized using
+    /// an empty initializer.
+    ///
+    /// For example, when all the properties of the struct have a default value,
+    /// the struct can be initialized using `Foo()`. This is important for
+    /// other types referencing this type.
+    var hasEmptyInit: Bool {
+        // If at least one property requires an explicit value, this struct
+        // cannot have an empty initializer.
+        !properties.contains(where: { $0.defaultValue == nil })
```
simonjbeaumont, line 94: Nit: Thin it would read nicer to invert this check:

```suggestion
        properties.allSatisfy { $0.defaultValue != nil }
```

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
```diff
@@ -36,20 +36,21 @@ extension TypesFileTranslator {
         let headerProperties: [PropertyBlueprint] = try headers.map { header in
             try parseResponseHeaderAsProperty(for: header)
         }
+        let headersStructBlueprint: StructBlueprint = .init(
```
simonjbeaumont, line 39: Nit: This syntax makes the reader ask "is there some kind of upcasting going on here?", could we replace with:

```suggestion
        let headersStructBlueprint =  StructBlueprint(
```

simonjbeaumont (commented):
Just a couple of minor nits from me, otherwise looks super—thank you!

Sources/_OpenAPIGeneratorCore/Translator/CommonTypes/StructBlueprint.swift
czechboy0: Agreed, my fault though as I suggested the original syntax 😄

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
czechboy0: I think we use this pattern a lot in this area though, `let foo: Foo = .init(<newline>`, so if we want to change the style, we should probably do that holistically.

Sources/_OpenAPIGeneratorCore/Translator/CommonTypes/StructBlueprint.swift
takeshi-1000: Thank you. 🙏

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
takeshi-1000: It seems reasonable to change it, 
but I'm not sure what to do with this fix given the overall style, so I didn't.
(Also, I personally don't have much trouble reading this case.)

czechboy0: You know how I'd love to resolve this - can we file an issue and discuss there? I'd like to make the case for the `let foo: Type = .init` syntax, but I also don't want to hold back this PR 🙂

simonjbeaumont: Of course—I wouldn't want this PR to be held up on this. It's already approved, but since I was added as a reviewer I left some feedback, and explicitly called them outs as nits. The delay is caused by a debate about existing style in the project, which isn't one I started 😉

takeshi-1000: Thank you all for your kindness and detailed explanations.
I just made an issue. https://github.com/apple/swift-openapi-generator/issues/158

In Japan time, it's already the end of work, so my activity time is half a day later.

czechboy0: Thanks for filing #158, I'll open a PR for it in a bit.

czechboy0: Ok PRs for addressing #158 are open, so marking this as resolved.

PR #152, opened 2023-07-31
[Generator] Design away EncodableBodyContent

czechboy0:
### Motivation

See https://github.com/apple/swift-openapi-runtime/pull/30

Depends on Runtime 0.1.6

### Modifications

Adapted the generator code for the runtime changes. Also, since this was factored out of the big PR (#146) adding multiple content types, you can see the code was prepared for handling multiple content types, even though still in this PR it only ever gets N=1.

### Result

The generated code uses the new simplified runtime functions, and prepares us for multiple content types.

### Test Plan

Adapted file-based reference tests to generate the new syntax.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponseOutcome.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponseOutcome.swift
@@ -306,10 +306,13 @@ extension ServerFileTranslator {
         }
         codeBlocks.append(contentsOf: headerExprs.map { .expression($0) })
 
-        let typedContents = [try bestSingleTypedContent(
-            typedResponse.response.content,
-            inParent: bodyTypeName
-        )].compactMap { $0 }
+        let typedContents = [
+            try bestSingleTypedContent(
+                typedResponse.response.content,
+                inParent: bodyTypeName
+            )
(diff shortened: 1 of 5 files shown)


Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponseOutcome.swift
```diff
@@ -306,73 +306,63 @@ extension ServerFileTranslator {
...
-            let contentTypeHeaderValue = typedContent.content.contentType.headerValueForValidation
-            let validateAcceptHeader: Expression = .try(
-                .identifier("converter").dot("validateAcceptIfPresent")
-                    .call([
-                        .init(label: nil, expression: .literal(contentTypeHeaderValue)),
-                        .init(label: "in", expression: .identifier("request").dot("headerFields")),
-                    ])
-            )
-            codeBlocks.append(.expression(validateAcceptHeader))
+        )].compactMap { $0 }
```
czechboy0, line 312: This "wrap an optional in an array and then compact map it" ugliness is just temporary, it'll be replaced soon with a function that returns an array of supported content types. Just did this here to make the eventual change a one-liner.

simonjbeaumont (approved):

PR #153, opened 2023-07-31
Deprecation warnings not emitted when calling operations that are deprecated

### LINKED ISSUE #149: Deprecation warnings not emitted when calling operations that are deprecated
When an operation is marked deprecated, the generated gives the APIProtocol definition an availability annotation marking it as deprecated. However the method on the `Client` object which implements the protocol is not given an annotation. When it is called, no warning is emitted. It's not clear to me if this is an Xcode/Swift bug, or expected behaviour.

```swift
let client = Client(
               serverURL: URL(string:"https://api.something.com")!,
               configuration: .init(),
               transport: URLSessionTransport(),
               middlewares: []
)

let erasedClient : some APIProtocol = client

// Emits no warning in Xcode 15b5
let a = try await client.someDeprecatedMethod()

// Emits a warning 
let b = try await erasedClient.someDeprecatedMethod()
```

czechboy0:
### Motivation

Fixes #149.

Turns out that when a function is deprecated in a protocol, then calling the function on a concrete type that conforms to that protocol does _not_ emit a warning unless the function _implementation_ is also annotated as deprecated.

### Modifications

This PR emits deprecation annotations for the client and server implementations of `APIProtocol`'s functions.

### Result

Warnings are now correctly emitted regardless of whether the adopter calls methods on `let client: MyClient` where `MyClient: APIProtocol`, or on `let client: any APIProtocol`, leading to better communication between the author of the OpenAPI document and the adopter of the generated Swift code about the API evolution and future removal of operations.

### Test Plan

Adapted integration tests.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/ClientTranslator/translateClientMethod.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/ClientTranslator/translateClientMethod.swift
@@ -212,6 +212,7 @@ extension ClientFileTranslator {
                     .expression(sendExpr)
                 ]
             )
+            .deprecate(if: description.operation.deprecated)
         )
     }
 }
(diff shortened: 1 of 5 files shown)


simonjbeaumont (approved):
LGTM, if it doesn't trip up the integration test.

PR #157, opened 2023-07-31
[NFC] Refactor PetstoreConsumerTests to allow multiple versions

czechboy0:
### Motivation

In upcoming features that contain breaking changes, but we'll stage them in behind a feature flag, we'd like to have multiple variants of the `PetstoreConsumerTests` test target, allowing us to test both the existing behavior and the new behavior once a feature flag is enabled. These additional variants would be temporary, and would be deleted again (or rather, the main test suite would be updated) once the feature flag is enabled unconditionally. So the steady state number of `PetstoreConsumerTest` variants would continue to be 1, just for short periods of time, it might be higher.

### Modifications

This PR makes that possible by refactoring common functionality into a new internal target `PetstoreConsumerTestCore`. Note that all the variants of the test target share the same OpenAPI document, but generate different variants of the reference code.

Highlights:
- new `PetstoreConsumerTestCore` target, depended on by the existing `PetstoreConsumerTests`
- added an ability to _not_ fail the test when specific diagnostics are emitted (through the new `ignoredDiagnosticMessages: Set<String>` parameter), which allows us to test both existing and new behavior on the same OpenAPI document. Other, non-allowlisted diagnostics, still continue to fail the test, so new ones won't sneak through undetected.
- many test functions now optionally take `featureFlags` and `ignoredDiagnosticMessages`, again allowing us to test both existing and proposed behavior hidden behind a feature flag

### Result

It will be a lot easier to temporarily introduce other variants of the `PetstoreConsumerTests` test target to allow for thoroughly testing features even before they are enabled by default, giving us more confidence in the accuracy of proposals and their associated implementations.

### Test Plan

All tests continue to pass, this is an NFC, a pure refactoring, making the next PR much smaller. To see how this all fits together, check out the PR that has all the changes: #146. That said, these partial PRs are easier to review, as they're each focused on one task.


Sources/PetstoreConsumerTestCore/Common.swift
```diff
@@ -0,0 +1,131 @@
+//===----------------------------------------------------------------------===//
```
czechboy0, line 1: This whole file has contents moved from the other Common.swift, no code was actually changed.

Tests/PetstoreConsumerTests/Common.swift
```diff
@@ -11,127 +11,10 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 //===----------------------------------------------------------------------===//
-import OpenAPIRuntime
```
czechboy0, line 14: Moved into the Common.swift file of the new shared target.

czechboy0:

Tests/PetstoreConsumerTests/Common.swift
```diff
@@ -11,127 +11,10 @@
...
-        )
-    }
-}
+import XCTest
 
 extension Operations.listPets.Output {
     static var success: Self {
         .ok(.init(headers: .init(My_Response_UUID: "abcd"), body: .json([])))
     }
 }
```
gjcairo, line 20: Potentially naive question but why wouldn't this belong in the common target too?

gjcairo (approved):
LGTM, just one small question

Tests/PetstoreConsumerTests/Common.swift
czechboy0: Because it extends `Operations.listPets.Output`, which is a generated type. Each of the targets has its own copy of the generated types (potentially slightly different), so we can't extend them in the shared library, which isn't tied to a specific copy of generated types.

simonjbeaumont (approved):

PR #162, opened 2023-08-02
Hide strict OpenAPI validation behind a feature flag

czechboy0:
### Motivation

In https://github.com/apple/swift-openapi-generator/pull/130, we introduced stricted validation of input OpenAPI documents. Reflecting back, since this might start rejecting OpenAPI document that were previously accepted, that is considered a breaking change and should be hidden behind a feature flag, only to be enabled unconditionally in the next breaking version.

### Modifications

Hide the new validation behind a feature flag.

### Result

Without providing the feature flag explicitly, the generator will maintain its old behavior and not perform strict validation.

### Test Plan

N/A - this is a speculative fix to maintain backwards compatibility in the 0.1.x version.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
+++ b/Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
@@ -125,31 +125,33 @@ func makeGeneratorPipeline(
             postTransitionHooks: [
                 { doc in
 
-                    // Run OpenAPIKit's built-in validation.
-                    try doc.validate()
+                    if config.featureFlags.contains(.strictOpenAPIValidation) {
+                        // Run OpenAPIKit's built-in validation.
+                        try doc.validate()
 
-                    // Validate that the document is dereferenceable, which
-                    // catches reference cycles, which we don't yet support.
-                    _ = try doc.locallyDereferenced()
+                        // Validate that the document is dereferenceable, which
+                        // catches reference cycles, which we don't yet support.
+                        _ = try doc.locallyDereferenced()
 
-                    // Also explicitly dereference the parts of components
-                    // that the generator uses. `locallyDereferenced()` above
-                    // only dereferences paths/operations, but not components.
-                    let components = doc.components
-                    try components.schemas.forEach { schema in
-                        _ = try schema.value.dereferenced(in: components)
-                    }
-                    try components.parameters.forEach { schema in
-                        _ = try schema.value.dereferenced(in: components)
-                    }
-                    try components.headers.forEach { schema in
-                        _ = try schema.value.dereferenced(in: components)
-                    }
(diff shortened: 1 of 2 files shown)


Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
```diff
@@ -125,31 +125,33 @@ func makeGeneratorPipeline(
...
+                        }
+                        try components.headers.forEach { schema in
+                            _ = try schema.value.dereferenced(in: components)
+                        }
+                        try components.requestBodies.forEach { schema in
+                            _ = try schema.value.dereferenced(in: components)
+                        }
+                        try components.responses.forEach { schema in
+                            _ = try schema.value.dereferenced(in: components)
+                        }
```
simonjbeaumont, line 154: Do we need to be more selective here about what's gated and what isn't?

Didn't we add these originally to reject documents with reference cycles and/or external references that we cannot support, and would otherwise fail to generate anyway?

If that's the case then shouldn't we keep the `locallyDereferened()` and all the `defererenced(in:)` checks, as this won't be a regression in what we can handle, it will just fail earlier, and with a clearer reason.

I can entertain the idea that `doc.validate()` could be causing regressions, but only because we treat it as fatal.

Maybe the better change here would be for us to call that and emit a warning diagnostic if it fails, rather than letting it throw?

czechboy0: I think for backwards compatibility, if there are any docs that now get rejected by the generator, but they didn't before (even if they produced potentially non-compiling code), it could be considered a breaking change, so I wanted to restore the original 0.1.x behavior, and enable all the strictness in 0.2.0, which will need to be an explicit upgrade for adopters, and they can fix up their docs, or file an issue on us, if the stricter mode breaks them.

Regarding whether failing to validate or dereference should be downgraded to a warning, I don't have a strong opinion other than if some strange use of OpenAPI gets through validation, we'll have to handle it in the translator, which is a lot more expensive. If we discover genuine examples of overly strict validation, we can customize the validator (instead of using the default OpenAPIKit one), and adjust. But from me skimming the default validations, they all seemed pretty reasonable to verify the structure of the document, and it seemed unlikely that we'd successfully generate compiling code from a doc that fails validation. But again, maybe there are exceptions and I'm open to softening validation on case-by-case basis. But for maintainers, it's easier to _start_ strict in 0.2.x and see what adopters come back with 🙂

simonjbeaumont: While this errs on the side of caution, I guess that makes sense for a safe patch release.

simonjbeaumont (approved):

PR #164, opened 2023-08-02
Improve the diagnostic for when a schema is unsupported

czechboy0:
### Motivation

Previously, when a schema that was, or contained a subschema that was, unsupported, adopter would get a pretty opaque warning diagnostic telling them the "schema is unsupported", without more details of why, and which part of it.

### Modifications

This PR moves from a simple Boolean status on the `isSchemaSupported` class of utility functions, and moves to a richer enum, which includes the reason and the schema itself when unsupported, leading to much more informative diagnostics.

### Result

Diagnostics emitted for unsupported schemas will be a lot more meaningful.

### Test Plan

Adapted the unit tests to make sure all the reasons for unsupported schemas are now tested, and the reason matches.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/isSchemaSupported.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/isSchemaSupported.swift
@@ -13,6 +13,46 @@
 //===----------------------------------------------------------------------===//
 import OpenAPIKit30
 
+/// A result of checking whether a schema is supported.
+enum IsSchemaSupportedResult: Equatable {
+
+    /// The schema is supported and can be generated.
+    case supported
+
+    /// The reason a schema is unsupported.
+    enum UnsupportedReason: Equatable, CustomStringConvertible {
+
+        /// Describes when no subschemas are found in an allOf, oneOf, or anyOf.
+        case noSubschemas
+
+        /// Describes when the schema is not object-ish, in other words isn't
+        /// an object, a ref, or an allOf.
+        case notObjectish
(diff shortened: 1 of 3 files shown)


Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/isSchemaSupported.swift
```diff
@@ -13,6 +13,46 @@
...
+
+    /// The reason a schema is unsupported.
+    enum UnsupportedReason: Equatable, CustomStringConvertible {
+
+        /// Describes when no subschemas are found in an allOf, oneOf, or anyOf.
+        case noSubschemas
+
+        /// Describes when the schema is not object-ish, in other words isn't
+        /// an object, a ref, or an allOf.
+        case notObjectish
```
simonjbeaumont, line 30: But this is a valid schema that isn't "object-ish", right?

```yaml
              schema:
                type: string
```

Is this for when it's supposed to be "object-ish" but isn't?

simonjbeaumont (approved):

Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/isSchemaSupported.swift
czechboy0: Yup. It's because in some places we only support a subset of valid schemas, and this is meant to communicate why the generator rejected a valid schema on its own, but not in that context (e.g. `oneOf` with a discriminator must have object-ish subschemas, and not e.g. a string).

Issue #143, opened 2023-07-26
Including the '\r\n' newline in a description results in non-compiling code


MacFriis:
using asp.net to write a web api, with Swashbuckle to document the api. the swagger.yaml files puts in \r\n if the documentation is on multiple lines 
C# - .net 
```
/// <summary>
/// get all words within the category and language
/// if the user is registered, this list will include words from other users
/// </summary>
/// <param name="category"></param>
/// <param name="language"></param>
/// <returns></returns>
[HttpGet, AllowAnonymous]
public async Task<ActionResult<List<Word>>> GetWords(
   [FromQuery] string category = "general",
   [FromHeader] string language = "en"
   ) ...
```
the resulting yaml file will have this 

```
 get:
      tags:
        - Words
      summary: "get all words within the category and language\r\nif the user is registered, this list will include words from other users"
      parameters:
        - name: category
          in: query
          description: ''
          style: form
          schema:
            type: string
            default: general
        - name: language
         ...
```
This result in a compile error:
```
1157: }
1158: }
1159: }
---
514:153: expected ')' to end parameter clause
===
514:153: expected ')' to end parameter clause
Command PhaseScriptExecution failed with a nonzero exit code
```

czechboy0:
Hi @MacFriis, thanks for filing the issue. Could you paste a formatted (using the ``` markdown syntax) snippet of the YAML and the generated Swift? That'll make it easier for us to understand what the issue is. Thanks 🙏

czechboy0:
Thanks, turns out the `\r` character wasn't being handled correctly.

czechboy0:
https://github.com/apple/swift-openapi-generator/pull/169

PR #169, opened 2023-08-03
[Bug] Handle the '\r\n' newline correctly

czechboy0:
### Motivation

Fixes https://github.com/apple/swift-openapi-generator/issues/143, which tracks including the '\r\n' newline in a description in the OpenAPI document.

### Modifications

Replaces the '\r\n' character with '\n' before rendering comments, as '\n' is handled properly.

### Result

OpenAPI documents that include '\r\n' in a description produce compiling code now.

### Test Plan

Added a unit test for this case.


czechboy0:

Sources/_OpenAPIGeneratorCore/Renderer/TextBasedRenderer.swift
```diff
@@ -54,12 +54,15 @@ struct TextBasedRenderer: RendererProtocol {
             prefix = "// MARK:"
             commentString = string
         }
-        return commentString.transformingLines { line in
-            if line.isEmpty {
-                return prefix
+        return
+            commentString
+            .replacingOccurrences(of: "\r", with: "\n")
```
simonjbeaumont, line 59: In the presence of `\r\n` this produces two newlines but I think it should just be one.

Presumably the goal of this patch is to handle both newline conventions: `LF` or `CR+LF`. IIUC, the `\r` is a carriage return.

Does `String` have any utilities for splitting on "newlines" which might just handle this for us internally?

czechboy0: Oh I wasn't sure, let me see if we can only produce a single newline here.

czechboy0: Ok, how about this? https://github.com/apple/swift-openapi-generator/pull/169/commits/e2e7f3b2eaf42004e3e749bbcd046fa7d07b0cdc

Sources/_OpenAPIGeneratorCore/Renderer/TextBasedRenderer.swift
```diff
@@ -721,7 +723,7 @@ fileprivate extension String {
     /// Returns an array of strings, where each string represents one line
     /// in the current string.
     func asLines() -> [String] {
-        split(separator: "\n", omittingEmptySubsequences: false)
+        split(omittingEmptySubsequences: false, whereSeparator: \.isNewline)
```
simonjbeaumont, line 726: Should we be omitting empty sequences?

Realise this is scope creep because this was probably the behaviour before but I think that we should be preserving the number of newlines in comments. If you consider that a comment may be interpreted as markdown, you need two newlines to start a new block.

simonjbeaumont (approved):

Sources/_OpenAPIGeneratorCore/Renderer/TextBasedRenderer.swift
czechboy0: Yeah, this is existing behavior, and I don't remember all the reasons we explicitly added this. I suspect it was to preserve the newlines that adopters have in their comments, be it markdown or just plain text.

simonjbeaumont: Oh. I completely inverted this check in my head! You’re right.

Issue #52, opened 2023-06-08
Feature "Unexploded query params" is not supported.


barnard-b:
Using an Open API file that has a parameter with `explode: false` results in a warning of
```
Feature "Unexploded query params" is not supported, skipping [context: foundIn=Operations.launcher_list.Input.Query (#/paths/2.2.0/launcher/GET/query)/launcher_config__ids]
```

`false` values appear to be listed as unsupported in the documentation.
<img width="489" alt="image" src="https://github.com/apple/swift-openapi-generator/assets/44757473/fa2ebc63-a7e4-49c8-9746-6adc526dc913">

```yaml
  /launcher/:
    get:
      operationId: launcher_list
      parameters:
        - in: query
          name: launcher_config__ids
          schema:
            type: array
            items:
              type: integer
          description: Comma-separated launcher config IDs.
          explode: false
          style: form
      responses:
        "200":
          description: A success response with a greeting.
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Greeting"
```

I am open to creating a PR, but this is my first time using Open API and would need guidance.

czechboy0:
Hi @barnard-b, thanks for reporting this! Indeed, `explode: false` isn't supported yet. The issue https://github.com/apple/swift-openapi-generator/issues/18 covers both `explode: false` and other styles, but we can keep this issue to only track adding support for `explode: false`.

We would definitely appreciate a PR! To learn more about contributing, check out these docs: https://swiftpackageindex.com/apple/swift-openapi-generator/0.1.0/documentation/swift-openapi-generator/contributing-to-swift-openapi-generator

And if you have any questions, don't hesitate to tag me here 🙂

czechboy0:
This will be released shortly in 0.1.8.

PR #171, opened 2023-08-04
[Generator] Support unexploded query items

czechboy0:
### Motivation

Depends on https://github.com/apple/swift-openapi-runtime/pull/35.

Fixes https://github.com/apple/swift-openapi-generator/issues/52.

By default, query items are encoded as exploded (`key=value1&key=value2`), but OpenAPI allows explicitly requesting them unexploded (`key=value1,value2`). This feature missing has shown up in a few OpenAPI documents recently.

### Modifications

Adapt the generator to provide the two new `style` and `explode` parameters to query item encoding/decoding functions.

### Result

We now support unexploded query items.

### Test Plan

Expanded snippet-based tests to allow generating not just types, but also parts of the client/server. This has allowed us to not have to expand file-based reference tests here.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
@@ -192,6 +192,9 @@ extension FileTranslator {
                 .content
                 .contentType
                 .codingStrategy
+
+            // Defaults are defined by the OpenAPI specification:
+            // https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#fixed-fields-10
             switch parameter.location {
             case .query, .cookie:
                 style = .form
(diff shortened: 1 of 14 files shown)


Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
```diff
@@ -189,6 +192,14 @@ extension FileTranslator {
                 .content
                 .contentType
                 .codingStrategy
+            switch parameter.location {
+            case .query, .cookie:
+                style = .form
+                explode = true
+            case .path, .header:
+                style = .simple
+                explode = false
```
glbrntt, line 201: These are just the defaults from https://swagger.io/specification/#parameter-object, right?

Where we implement specs in our other projects we tend to drop in references to them so that if we return to the code later we can see why a decision was made (i.e. we are following the spec vs. just doing what we think is right).

glbrntt (approved):
From what I can tell, this all looks good. It might be worth getting some eyes with more knowledge of this project to look as well 👀

Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
czechboy0: Yup it's the default from the specification, sure let me add the link as well.

- [ ] Add a link to the specification

czechboy0:
Thanks for the review, @glbrntt, could you also take a look at the associated runtime PR please? https://github.com/apple/swift-openapi-runtime/pull/35

PR #173, opened 2023-08-07
Refactor ContentType + improve content type matching

czechboy0:
### Motivation

The `ContentType` type had two issues:
- the enum where all cases had the same associated string value was not a good representation of the data
- the parsing was too strict and didn't match e.g. `foo/bar+json` as a JSON content type

### Modifications

Refactored `ContentType` to split the raw value storage from the `Category` (new term, ideas of a better name are welcome) of content types (`json`, `text`, or `binary`).

Also, improved the mapping to match https://json-schema.org/draft/2020-12/json-schema-core.html#section-4.2

### Result

Now the type is easier to work with, and parsing correctly recognizes `foo/bar+json` as JSON.

### Test Plan

Added unit tests.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/Content/ContentType.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Content/ContentType.swift
@@ -17,103 +17,106 @@ import OpenAPIKit30
 ///
 /// Represents the serialization method of the payload and affects
 /// the generated serialization code.
-enum ContentType: Hashable {
+struct ContentType: Hashable {
 
-    /// A content type for JSON.
-    case json(String)
+    /// The category of a content type.
+    enum Category: Hashable {
 
-    /// A content type for any plain text.
-    case text(String)
+        /// A content type for JSON.
+        case json
 
-    /// A content type for raw binary data.
-    case binary(String)
+        /// A content type for any plain text.
+        case text
+
+        /// A content type for raw binary data.
+        case binary
+
+        /// Creates a category from the provided raw string.
+        ///
+        /// First checks if the provided content type is a JSON, then text,
+        /// and uses binary if none of the two match.
+        /// - Parameter rawValue: A string with the content type to create.
+        init(rawValue: String) {
+            // https://json-schema.org/draft/2020-12/json-schema-core.html#section-4.2
+            if rawValue == "application/json" || rawValue.hasSuffix("+json") {
+                self = .json
+                return
+            }
+            if rawValue.hasPrefix("text/") {
+                self = .text
+                return
+            }
+            self = .binary
+        }
+
+        /// The coding strategy appropriate for this content type.
+        var codingStrategy: CodingStrategy {
+            switch self {
+            case .json:
+                return .json
+            case .text:
+                return .text
+            case .binary:
+                return .binary
+            }
+        }
+    }
+
+    /// The underlying raw content type string.
+    private var rawValue: String
+
+    /// The mapped content type category.
+    private(set) var category: Category
 
     /// Creates a new content type by parsing the specified MIME type.
     /// - Parameter rawValue: A MIME type, for example "application/json".
-    init?(_ rawValue: String) {
-        if rawValue.hasPrefix("application/") && rawValue.hasSuffix("json") {
-            self = .json(rawValue)
-            return
-        }
-        if rawValue.hasPrefix("text/") {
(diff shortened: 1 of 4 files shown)


Sources/_OpenAPIGeneratorCore/Translator/Content/ContentType.swift
```diff
@@ -17,103 +17,106 @@ import OpenAPIKit30
...
+            case .binary:
+                return .binary
+            }
+        }
+    }
 
-    /// A content type for any plain text.
-    case text(String)
+    /// The underlying raw content type string.
+    private var rawValue: String
```
glbrntt, line 66: Is this ever modified? Wondering if it can be `private let` so that the `category` can be `internal let`?

Sources/_OpenAPIGeneratorCore/Translator/Content/ContentType.swift
```diff
@@ -17,103 +17,106 @@ import OpenAPIKit30
...
+            // https://json-schema.org/draft/2020-12/json-schema-core.html#section-4.2
+            if rawValue == "application/json" || rawValue.hasSuffix("+json") {
+                self = .json
+                return
+            }
+            if rawValue.hasPrefix("text/") {
+                self = .text
+                return
+            }
+            self = .binary
```
glbrntt, line 49: Why can we bucket everything else as `binary`? I don't think e.g. `xml` falls in to any of these categories.

gjcairo: I agree with this, it seems a bit odd to me to default to `binary`. 

And a bit of an aside, but on the topic of XML, looks like `+xml` is also a valid suffix: https://www.iana.org/assignments/media-types/media-types.xhtml

czechboy0: The old name of "binary" used to be "raw", it means we just collect the bytes without trying to interpret them at all.

The order is:
- if it's JSON, we use JSONDecoder 
- if it's text, we parse it as UTF-8 data into a string 
- otherwise, we don't introspect the bytes 

We currently don't support XML as a structured format, so it's just as raw as an octet-stream to us. That might change in the future though.

Sources/_OpenAPIGeneratorCore/Translator/Content/ContentType.swift
czechboy0: Good point, I'll make both let.

Sources/_OpenAPIGeneratorCore/Translator/Content/ContentType.swift
glbrntt: Right, that makes sense. I think `binary` is very misleading... I actually prefer the old "raw" name but maybe something more explicit like "uninterpreted" is better.

czechboy0: Yeah. That said, today genuine binary (`application/octet-stream`) and fallback binary (anything else) are treated the same way. If we wanted to rename and split, that'd likely have to propagate through the whole generator implementation.

gjcairo (approved):

Sources/_OpenAPIGeneratorCore/Translator/Content/ContentType.swift
glbrntt: That's fair. Would you mind expanding on the comment for the `binary` case?

glbrntt (approved):
Looks good but could you add a comment to the `binary` case before merging?

Sources/_OpenAPIGeneratorCore/Translator/Content/ContentType.swift
czechboy0: Done in https://github.com/apple/swift-openapi-generator/pull/173/commits/cd076256f750fc21410b1eca86be15c19461ac29

czechboy0:
@glbrntt, actually I made a few more changes in https://github.com/apple/swift-openapi-generator/pull/173/commits/91828acf3a6e47ee8f40e0cb0776e6d6b5575f7d, could you take a look if that still has your approval? It was aimed at avoiding callers accidentally using case sensitive comparisons of content types, plus it allows access to the parsed `type` and `subtype` components.

glbrntt (approved):

Issue #118, opened 2023-07-14
Nullable enums with an empty string fail to get generated


czechboy0:
In #59, we added support for an empty string as a valid case in a string enum, however, we only tested it on enum schemas _not_ marked as `nullable`.

Turns out, that when marked as `nullable`, OpenAPIKit parses the values slightly [differently](https://github.com/mattpolzin/OpenAPIKit/blob/release/3_0/Sources/OpenAPIKit30/Schema%20Object/JSONSchemaContext.swift#L662-L666), and treats the empty string as a nil string instead. That nil value, when converted to OpenAPIKit's `AnyCodable`, [gets turned](https://github.com/mattpolzin/OpenAPIKit/blob/release/3_0/Sources/OpenAPIKitCore/AnyCodable/AnyCodable.swift#L39) into a `Void`.

Before we introduce a workaround in the generator, @mattpolzin is this intentional behavior, and if so, can you say more about it please?

czechboy0:
To reproduce, we can just use:

```yaml
type: string
nullable: true
enum:
  - foo
  - ""
```

The second enum value will be an empty string when the schema is _not_ nullable, and `Void` when the schema is nullable.

mattpolzin:
Unfortunately this is a known issue with Yams. https://github.com/jpsim/Yams/issues/301

OpenAPIKit has special case decoding of nullable string enums because otherwise a null case gets read in as the string "null" instead of the value `null`, but it is unintentional that the empty string also gets read in as `null`. 

There is a proposed fix for the issue against Yams and a fork with that fix under the CreateAPI project but no indication of whether the fix will be upstreamed.

czechboy0:
Thanks for the find, @mattpolzin - I wonder if @jpsim can tell us more about the prioritization of landing that [fix](https://github.com/jpsim/Yams/pull/351)?

czechboy0:
I suspect we'll need to check for this ourselves in the generator, it's not a big deal, just wanted to try to see if Yams could fix it, but I suspect that'd be a breaking change anyway, so might not be likely.

Issue #82, opened 2023-06-20
Support nullable values


m4p:
Commit hash: f530c906215528a7450fc7602b9f93d9236eb4c0

Context:
While testing my application that uses with swift-openapi-generator, I noticed that `nullable` is not yet supported on Schema Objects values. This leads to a Decoder error for API calls that return `null` for nullable values.

Steps to reproduce:
1. Use a OpenAPI spec that relies on nullable values
2. Make an api call that returns some null values
3. Expected: Nullable values returned as optionals
4. Actual: `DecodingError: valueNotFound String - Expected String value but found null instead.`

$ swift --version
swift-driver version: 1.75.2 Apple Swift version 5.8.1 (swiftlang-5.8.0.124.5 clang-1403.0.22.11.100)
Target: arm64-apple-macosx13.0

$ uname -a
Darwin hayai.lan 22.5.0 Darwin Kernel Version 22.5.0: Mon Apr 24 20:52:24 PDT 2023; root:xnu-8796.121.2~5/RELEASE_ARM64_T6000 arm64

My system has IPv6 enabled.

czechboy0:
Hi @m4p, thanks for taking the time to file the issue.

Could you attach a short snippet of OpenAPI (ideally YAML) that shows more of the context of how you use nullable values? My understanding is that object properties not being required, and a schema being nullable, are similar in the sense that we want an optional value in Swift. (@mattpolzin can add more background here).

With the OpenAPI snippet, I'm also curious how you think the generated Swift code should change, to help us understand how and whether we need to treat non-required object properties and nullable schemas differently. Thanks! 🙏

m4p:
Yeah, removing nullable values from required results in the code working as expected, so it may be an issue with the quality of the spec.

czechboy0:
I believe it's legal to have a required property marked as nullable, it's just a question of how to represent that well in Swift, where `{}` and `{"foo": null}` is treated the same way for a Swift Decodable struct that looks like:

```swift
struct Foo: Decodable {
  var foo: String?
}
```

I suspect this is a genuine thing we should handle in Swift OpenAPI Generator, by not just checking whether a property schema is required, but also whether it's nullable.

Just to confirm my suspicion, the code that emits the error you had in the issue description, is that property generated as `foo: String` or `foo: String?`.

mattpolzin:
[...]

PR #176, opened 2023-08-08
Nullable enums with an empty string fail to get generated

czechboy0:
### Motivation

Fixes https://github.com/apple/swift-openapi-generator/issues/118.

The issue is that when a `type: string, enum: ...` schema is marked as `nullable: true`, _empty cases are parsed as Void_ instead of _nil_. See the issue for the investigation and identifying the root cause, but the TL;DR is that we need a workaround in the generator in the short term, and it's uncertain if the issue will ever be addressed in our dependency Yams (might be unlikely, as it'd probably be a breaking change for them at this point).

### Modifications

This PR adds code that catches this case and converts the `Void` into an empty string, which is what the OpenAPI document author actually put into their document in the first place.

### Result

Now `nullable: true` string enums with an empty case can be generated correctly. One example of such use is the [GitHub API](https://github.com/github/rest-api-description).

Note that this PR isn't really adding support for nullable schemas in general, that's tracked by https://github.com/apple/swift-openapi-generator/issues/82. We're just working around one specific scenario in which a relatively common input trips up the generator.

### Test Plan

Added a unit test for each of: non-nullable and nullable variants.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateStringEnum.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateStringEnum.swift
@@ -20,15 +20,23 @@ extension FileTranslator {
     ///   - typeName: The name of the type to give to the declared enum.
     ///   - openAPIDescription: A user-specified description from the OpenAPI
     ///   document.
+    ///   - isNullable: Whether the enum schema is nullable.
     ///   - allowedValues: The enumerated allowed values.
     func translateStringEnum(
         typeName: TypeName,
         userDescription: String?,
+        isNullable: Bool,
         allowedValues: [AnyCodable]
     ) throws -> Declaration {
 
         let rawValues = try allowedValues.map(\.value)
             .map { anyValue in
+                // In nullable enum schemas, empty strings are parsed as Void.
+                // This is unlikely to be fixed, so handling that case here.
+                // https://github.com/apple/swift-openapi-generator/issues/118
+                if isNullable && anyValue is Void {
+                    return ""
+                }
                 guard let string = anyValue as? String else {
                     throw GenericError(message: "Disallowed value for a string enum '\(typeName)': \(anyValue)")
                 }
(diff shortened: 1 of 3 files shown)


Tests/OpenAPIGeneratorCoreTests/Translator/CommonTranslations/Test_translateStringEnum.swift
```diff
@@ -0,0 +1,63 @@
...
+        )
+        XCTAssertEqual(names, ["a", "_empty", "undocumented"])
+    }
+
+    func testCaseValuesForNullableSchema() throws {
+        let names = try _caseValues(
+            .string(
+                nullable: true,
+                allowedValues: "a",
+                nil
```
czechboy0, line 35: I'm not loving how swift-format splits the var args across lines like this, but it's out of my hands 😄

Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateStringEnum.swift
```diff
@@ -20,15 +20,23 @@ extension FileTranslator {
...
     ) throws -> Declaration {
 
         let rawValues = try allowedValues.map(\.value)
             .map { anyValue in
+                // In nullable enum schemas, empty strings are parsed as Void.
+                // This is unlikely to be fixed, so handling that case here.
+                // https://github.com/apple/swift-openapi-generator/issues/118
+                if isNullable && anyValue is Void {
+                    return ""
+                }
```
simonjbeaumont, line 39: Do we even need the `isNullable` parameter? i.e. would it be safe to just test if `anyValue is Void` unilaterally?

simonjbeaumont (approved):

gjcairo (approved):

Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateStringEnum.swift
czechboy0: _Maybe_. I wanted to make it as focused as possible, so that we don't accidentally catch some case we don't know about. So if another Void case comes up, I'd like us to get an issue filed and consider it deliberately.

PR #178, opened 2023-08-08
Stop treating schema warnings as errors

czechboy0:
### Motivation

In https://github.com/apple/swift-openapi-generator/pull/130, we introduced extra validation by calling into OpenAPIKit's `validate` method. (It's hidden behind a feature flag, but I've been testing with it enabled.)

It looks for structural issues, such as non-unique operation ids, which would result in us generating non-compiling code anyway, so the validation helps catch these early and emit descriptive errors that adopters can use to fix their doc.

However, the method also takes a parameter `strict`, which defaults to `true`, and when enabled, it turns warnings emitted during schema parsing into errors. This part is _too_ strict for us and was rejecting OpenAPI documents that were valid enough for our needs.

An example of a schema warning is a schema having `minItems: 1` on a non-array schema. While it's not technically correct, it also doesn't impede our understanding of the non-array schema, as we never actually check what the value of `minItems` is. That's why these are just warnings, not errors, so we should stop promoting them to fatal errors that block an adopter from generating code.

### Modifications

This PR flips the `strict` parameter to `false`. This doesn't make us miss out on these warnings, as recently (in https://github.com/apple/swift-openapi-generator/pull/174), we started forwarding these schema warnings into generator diagnostics, so the adopter will see them still, and can address them on their own schedule.

### Result

Now documents with only schema warnings aren't rejected by the generator anymore.

### Test Plan

Added a unit test of the broken-out `validateDoc` function, ensures that a schema with warnings doesn't trip it up anymore.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
+++ b/Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
@@ -1,0 +1,58 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftOpenAPIGenerator open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftOpenAPIGenerator project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// Runs validation steps on the incoming OpenAPI document.
+/// - Parameters:
+///   - doc: The OpenAPI document to validate.
+///   - config: The generator config.
+/// - Throws: An error if a fatal issue is found.
+func validateDoc(_ doc: ParsedOpenAPIRepresentation, config: Config) throws {
+    guard config.featureFlags.contains(.strictOpenAPIValidation) else {
+        return
+    }
+    // Run OpenAPIKit's built-in validation.
+    // Pass `false` to `strict`, however, because we don't
+    // want to turn schema loading warnings into errors.
+    // We already propagate the warnings to the generator's
+    // diagnostics, so they get surfaced to the user.
+    // But the warnings are often too strict and should not
+    // block the generator from running.
+    // Validation errors continue to be fatal, such as
+    // structural issues, like non-unique operationIds, etc.
+    try doc.validate(strict: false)
+
+    // Validate that the document is dereferenceable, which
+    // catches reference cycles, which we don't yet support.
+    _ = try doc.locallyDereferenced()
+
+    // Also explicitly dereference the parts of components
+    // that the generator uses. `locallyDereferenced()` above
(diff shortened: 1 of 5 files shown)


Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
```diff
@@ -0,0 +1,69 @@
...
+    // Run OpenAPIKit's built-in validation.
+    // Pass `false` to `strict`, however, because we don't
+    // want to turn schema loading warnings into errors.
+    // We already propagate the warnings to the generator's
+    // diagnostics, so they get surfaced to the user.
+    // But the warnings are often too strict and should not
+    // block the generator from running.
+    // Validation errors continue to be fatal, such as
+    // structural issues, like non-unique operationIds, etc.
+    let warnings = try doc.validate(strict: false)
```
simonjbeaumont, line 33: Maybe we can provide an even more consistent experience by catching the errors from this call and emitting fatal `.error` diagnostics?

simonjbeaumont (approved):

Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
czechboy0: Good idea, I'll emit an error diagnostic but still also throw the error, to stop the execution. Diagnostics themselves don't have an easy way to abort the execution today.

czechboy0: Actually, I can't do that, as we're returning the diagnostics here, not submitting them to the collector, so we can't both return the diagnostic and throw an error. But I do want to think more about whether we use `.error` diagnostics today, and if not, remove them. Or, in turn, make them more powerful and migrate some error throwing to them.

simonjbeaumont: Wasn't suggesting you throw here. Was suggesting you add an `.error` diagnostic. But maybe I misunderstood the `.error` diagnostic, which I thought _did_ abort the generator. The docs attached to that enum case imply it's fatal. Maybe it isn't fatal at all? 😁

czechboy0: Right, that's what I mean - it depends on the collector implementation what it does, and we have two collectors used by the CLI: the YAML and stderr printing one, neither of which abort the generator. We should probably find a way where the collector can signal back to the main generator logic that it wants to abort, but it'll be something we have to build.

The `error` diagnostic was designed to be fatal, but it was never fully enforced across all collectors. So we should probably either enforce it more, or remove the error severity altogether to avoid the confusion.

simonjbeaumont: I vote we keep it; maybe even rename it fatal. WDYT? Could you open an issue and we can converge on a stance? Even if we end up removing it, we can close the loop on this.

czechboy0: Filed as #180, we can continue there.

PR #179, opened 2023-08-08
Allow more types of subschemas in anyOf/allOf

czechboy0:
### Motivation

Previously, we enforced that child schemas of `anyOf` and `allOf` are object-ish (aka objects, refs to objects, or allOfs of objects) schemas. That made is easier to get started, but that limitation isn't needed anymore, the generated code can handle any child schemas nowadays, and a few high-profile OpenAPI docs are using this and were blocked on our limitation.

### Modifications

Remove the limitation that only object-ish schemas can be child schemas of anyOf and allOf.

> Note: oneOf is different, we always supported any schema when there is no discriminator specified, and require refs to object-ish schemas when a discriminator is specified.

### Result

OpenAPI documents that, for example do the following, can be generated now:

```
anyOf:
  - type: string
  - type: integer
```

### Test Plan

Updated the unit tests for `isSchemaSupported`, where most of the work was, plus amended anyOf/allOf snippet tests to also include non-object-ish schemas.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/isSchemaSupported.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/isSchemaSupported.swift
@@ -242,8 +242,9 @@ extension FileTranslator {
     /// - Parameter schema: A schema to check.
     func isRefToObjectishSchemaAndSupported(_ schema: JSONSchema) throws -> IsSchemaSupportedResult {
         switch schema.value {
-        case .reference:
-            return try isObjectishSchemaAndSupported(schema)
+        case let .reference(ref, _):
+            let referencedSchema = try components.lookup(ref)
+            return try isObjectishSchemaAndSupported(referencedSchema)
         default:
             return .unsupported(
                 reason: .notRef,
--- a/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/isSchemaSupported.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/isSchemaSupported.swift
@@ -141,7 +141,7 @@ extension FileTranslator {
                     schema: schema
                 )
             }
-            return try areObjectishSchemasAndSupported(schemas)
+            return try areSchemasSupported(schemas)
         case .any(of: let schemas, _):
             guard !schemas.isEmpty else {
                 return .unsupported(
@@ -149,7 +149,7 @@ extension FileTranslator {
                     schema: schema
                 )
             }
-            return try areObjectishSchemasAndSupported(schemas)
+            return try areSchemasSupported(schemas)
         case .one(of: let schemas, let context):
             guard !schemas.isEmpty else {
                 return .unsupported(
@@ -206,20 +206,6 @@ extension FileTranslator {
         return .supported
     }
 
-    /// Returns a result indicating whether the provided schemas
-    /// are reference, object, or allOf schemas and supported.
-    /// - Parameter schemas: Schemas to check.
-    /// - Returns: `.supported` if all schemas match; `.unsupported` otherwise.
-    func areObjectishSchemasAndSupported(_ schemas: [JSONSchema]) throws -> IsSchemaSupportedResult {
-        for schema in schemas {
-            let result = try isObjectishSchemaAndSupported(schema)
-            guard result == .supported else {
-                return result
-            }
-        }
-        return .supported
-    }
-
     /// Returns a result indicating whether the provided schema
     /// is an reference, object, or allOf (object-ish) schema and is supported.
     /// - Parameter schema: A schemas to check.
@@ -228,7 +214,7 @@ extension FileTranslator {
         case .object, .reference:
             return try isSchemaSupported(schema)
         case .all(of: let schemas, _):
-            return try areObjectishSchemasAndSupported(schemas)
+            return try areSchemasSupported(schemas)
         default:
             return .unsupported(
                 reason: .notObjectish,
--- a/Sources/swift-openapi-generator/Documentation.docc/Articles/Supported-OpenAPI-features.md
+++ b/Sources/swift-openapi-generator/Documentation.docc/Articles/Supported-OpenAPI-features.md
@@ -142,12 +142,12 @@ Supported features are always provided on _both_ client and server.
 - [x] enum
 - [x] type
 - [x] allOf
-    - a wrapper struct is generated and children must be object schemas
+    - a wrapper struct is generated, children can be any schema
 - [x] oneOf
-    - if a discriminator is specified (recommended), children must be object schemas
+    - if a discriminator is specified, each child must be a reference to an object schema
     - if no discriminator is specified, children can be any schema
 - [x] anyOf
-    - children must be object schemas
+    - a wrapper struct is generated, children can be any schema
 - [ ] not
 - [x] items
 - [x] properties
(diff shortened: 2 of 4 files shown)


Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
```diff
@@ -1223,6 +1251,13 @@ extension SnippetBasedReferenceTests {
         line: UInt = #line
     ) throws {
         let translator = try makeTypesTranslator(componentsYAML: componentsYAML)
+        let schemas = translator.components.schemas
+        for (key, schema) in schemas {
+            XCTAssertTrue(
+                try translator.isSchemaSupported(schema) == .supported,
+                "Schema \(schema.prettyDescription) for key \(key) is not supported"
+            )
+        }
```
simonjbeaumont, line 1260: Is this a preflight check of the test, rather than the test itself? What would be the behaviour if we removed this; wouldn't it just throw in `translator.translateSchemas`? IMO, that would be fine, because we can keep the logic simpler here and have tests that `XCTAssertThrowsError` for things we don't support today, if we'd like to keep track of them.

Sources/swift-openapi-generator/Documentation.docc/Articles/Supported-OpenAPI-features.md
```diff
@@ -142,12 +142,12 @@ Supported features are always provided on _both_ client and server.
 - [x] enum
 - [x] type
 - [x] allOf
-    - a wrapper struct is generated and children must be object schemas
+    - a wrapper struct is generated, children can be any schema
 - [x] oneOf
-    - if a discriminator is specified (recommended), children must be object schemas
+    - if a discriminator is specified, each child must be a reference to an object schema
```
simonjbeaumont, line 147: > When using the discriminator, inline schemas will not be considered.

Just want to check that we're not being over-strict here. We shouldn't consider inline schemas, sure; but hopefully we're not failing if we find any?

Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/isSchemaSupported.swift
```diff
@@ -256,8 +242,9 @@ extension FileTranslator {
     /// - Parameter schema: A schema to check.
     func isRefToObjectishSchemaAndSupported(_ schema: JSONSchema) throws -> IsSchemaSupportedResult {
         switch schema.value {
-        case .reference:
-            return try isObjectishSchemaAndSupported(schema)
+        case let .reference(ref, _):
+            let referencedSchema = try components.lookup(ref)
+            return try isObjectishSchemaAndSupported(referencedSchema)
```
simonjbeaumont, line 247: This lookup isn't required IIUC, as AFAICT `isObjectishSchemaAndSupported` calls `isSchemaSupported` which also does the lookup, in the case of a `.reference`.

Tests/OpenAPIGeneratorCoreTests/Translator/TypeAssignment/Test_isSchemaSupported.swift
```diff
@@ -60,28 +60,42 @@ class Test_isSchemaSupported: XCTestCase {
...
             .object(properties: [
                 "Foo": .string
             ]),
             .reference(.component(named: "MyObj")),
+            .string,
+            .array(items: .string),
         ]),
 
-        // oneOf with two schemas
+        // oneOf with a discriminator with two objectish schemas
```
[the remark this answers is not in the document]
czechboy0: It's okay. The relationship is an implication: if the `oneOf` has a `discriminator`, then all the schemas MUST be objectish. But not the other way around, objectish schemas can appear in a `oneOf` even without a `discriminator`.

Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/isSchemaSupported.swift
czechboy0: It was required, as we need to make sure the target of the reference is `isObjectishSchemaAndSupported`, not just `isSchemaSupported`. But yup, it won't work if there's a reference to a reference, but I think that's fine.

Sources/swift-openapi-generator/Documentation.docc/Articles/Supported-OpenAPI-features.md
czechboy0: We are - because if a discriminator is provided, the subschemas must be references to objects, because the discriminator tells you the name of the schema to decode as, but inline schemas don't have a name.

OpenAPI 3.1 does support the concept of a "const" value, which potentially could allow inline object schemas with a discriminator, but 3.0.x doesn't, so my interpretation is that it's not supported. But as always, if we find an example of where that doesn't hold, we'll keep improving this.

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
czechboy0: I think you're right that it's not necessary, I'll remove it.

czechboy0: Done: https://github.com/apple/swift-openapi-generator/pull/179/commits/8385093820b542c75cdb5b9900e95f4b8e900bc6

Sources/swift-openapi-generator/Documentation.docc/Articles/Supported-OpenAPI-features.md
simonjbeaumont: What I'm saying is that the text of the spec, doesn't say that this invariant needs to hold in the OpenAPI document, just that the tools should maybe discard inline specs:

> When using the discriminator, inline schemas will not be considered.

czechboy0: Ah gotcha. Hmm, due to how we separate validation from generation, we're not currently set up to do that easily. I'll file it separately and we can fix that later. I think that's okay, as we've always rejected inline schemas in oneOfs with a discriminator, so this PR doesn't regress that.

czechboy0: Filed: https://github.com/apple/swift-openapi-generator/issues/181

Tests/OpenAPIGeneratorCoreTests/Translator/TypeAssignment/Test_isSchemaSupported.swift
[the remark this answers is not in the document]
czechboy0: Ah sorry, you're right, I got confused and thought the diff was your suggestion, not my diff 🤦‍♂️ Will add the discriminator, thanks.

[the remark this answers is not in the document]
czechboy0: Fixed: https://github.com/apple/swift-openapi-generator/pull/179/commits/6535e7ecf9b64877cbef2c06f66af54cc2aec5dd

Sources/swift-openapi-generator/Documentation.docc/Articles/Supported-OpenAPI-features.md
simonjbeaumont: > we've always rejected inline schemas in oneOfs with a discriminator, so this PR doesn't regress that.

OK, given this isn't a regression in that regard, I think tracking this as a separate issue is acceptable; thanks for filing!

simonjbeaumont (approved):

Issue #168, opened 2023-08-03
Overriding path item parameters with operation parameters doesn't work


mh-dmclean:
Hello there,

I've recently started experimenting with the openapi-generator seeing if it could help make some things easier at work. The generator seems to work but leads to compilation errors when having overridden variables, Easily modifying the GreetingService example Apple provides can reproduce this problem. [Here](https://gist.github.com/mh-dmclean/14b26befc7ad5f3996bd53890ca4fbfc) is a link to a gist containing the modified example and the `Types.swift` that is outputted. The error itself is on Lines 58/59 on Types.swift in the gist.

Package.resolved revision: c181ed

czechboy0:
Hi @mh-dmclean, you're right, great catch!

From the OpenAPI 3.0.3 specification for the Path Item -> parameters [^1]:

> A list of parameters that are applicable for all the operations described under this path. These parameters can be overridden at the operation level, but cannot be removed there.

But today, we simply concatenate the parameters from the path item level with those at operation level here: https://github.com/apple/swift-openapi-generator/blob/main/Sources/_OpenAPIGeneratorCore/Translator/Operations/OperationDescription.swift#L150

However, we should instead _replace_ any parameters we find at the path item level with one found for the operation, if they have the same location + name.

A contribution with a fix and a unit test for this would be much appreciated 🙂 

[^1]: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#fixed-fields-7

arthurcro:
Hi @czechboy0! 😄 I'm intreseted in contributing to [swift-openapi-generator](https://github.com/apple/swift-openapi-generator).  I'll happily look at this issue if not picked up yet! 🙂

czechboy0:
Great, welcome, @arthurcro! Hopefully my previous comment has enough information to start on, and you might also want to check out https://swiftpackageindex.com/apple/swift-openapi-generator/0.1.7/documentation/swift-openapi-generator/contributing-to-swift-openapi-generator. Let us know if you need any help! 🙏

arthurcro:
I believe I found a first decent fix for the issue. However, I'm struggling finding the right place to write a unit test to confirm it propoerly resolves the issue. 😄 
I did not find isolated unit tests testing `OperationDescription` since it's mostly tested through other tests. Could you maybe help me find the right place to add a great unit test? 🙂 

> However, we should instead replace any parameters we find

[...]

PR #183, opened 2023-08-09
Retain operation level parameters over path level ones

arthurcro:
### Motivation

Fixes https://github.com/apple/swift-openapi-generator/issues/168.

### Modifications

Previously, an operation's parameters were returned as a concatenation of the parameters from the path item level with those at operation level.

However, this is incorrect according to https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#fixed-fields-7. 
> A list of parameters that are applicable for all the operations described under this path. These parameters can be overridden at the operation level, but cannot be removed there. The list MUST NOT include duplicated parameters.

With this change, we are merging the two arrays of parameters using a unique identifier of the location + name of parameters. If duplicate parameters exist, only the parameters from the operation level are preserved.

### Result

After this change, an operation's parameters won't contain any duplicate. 

### Test Plan

Add tests for `OperationDescription.allParameters` at `Tests/OpenAPIGeneratorCoreTests/Translator/Operations/Test_OperationDescription.swift`.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/Operations/OperationDescription.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Operations/OperationDescription.swift
@@ -154,7 +154,7 @@ extension OperationDescription {
             var mergedParameters: [UnresolvedParameter] = []
             var uniqueIdentifiers: Set<String> = []
 
-            for parameter in pathParameters + operation.parameters {
+            for parameter in operation.parameters + pathParameters {
                 let resolvedParameter = try parameter.resolve(in: components)
                 let identifier = resolvedParameter.location.rawValue + ":" + resolvedParameter.name
 
--- a/Sources/_OpenAPIGeneratorCore/Translator/Operations/OperationDescription.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Operations/OperationDescription.swift
@@ -151,15 +151,16 @@ extension OperationDescription {
     /// - Throws: When an invalid JSON reference is found.
     var allParameters: [UnresolvedParameter] {
         get throws { 
-            var mergedParameters = [UnresolvedParameter]()
-            var uniqueIdentifiers = Set<String>()
+            var mergedParameters: [UnresolvedParameter] = []
+            var uniqueIdentifiers: Set<String> = []
 
             for parameter in pathParameters + operation.parameters {
                 let resolvedParameter = try parameter.resolve(in: components)
-                let identifier = resolvedParameter.name + resolvedParameter.location.rawValue
+                let identifier = resolvedParameter.location.rawValue + ":" + resolvedParameter.name
 
-                guard !uniqueIdentifiers.contains(identifier) 
-                else { continue }
+                guard !uniqueIdentifiers.contains(identifier) else { 
+                    continue 
+                }
 
                 mergedParameters.append(parameter)
                 uniqueIdentifiers.insert(identifier)
--- a/Sources/_OpenAPIGeneratorCore/Translator/Operations/OperationDescription.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Operations/OperationDescription.swift
@@ -150,23 +150,24 @@ extension OperationDescription {
     /// - Returns: An array of merged path item and operation level parameters without duplicates.
     /// - Throws: When an invalid JSON reference is found.
     var allParameters: [UnresolvedParameter] {
-        get throws { 
+        get throws {
             var mergedParameters: [UnresolvedParameter] = []
             var uniqueIdentifiers: Set<String> = []
 
-            for parameter in operation.parameters + pathParameters {
+            let allParameters = pathParameters + operation.parameters
+            for parameter in allParameters.reversed() {
                 let resolvedParameter = try parameter.resolve(in: components)
                 let identifier = resolvedParameter.location.rawValue + ":" + resolvedParameter.name
 
-                guard !uniqueIdentifiers.contains(identifier) else { 
-                    continue 
+                guard !uniqueIdentifiers.contains(identifier) else {
+                    continue
                 }
 
                 mergedParameters.append(parameter)
                 uniqueIdentifiers.insert(identifier)
             }
 
-            return mergedParameters
+            return mergedParameters.reversed()
         }
     }
 
--- a/Sources/_OpenAPIGeneratorCore/Translator/Operations/OperationDescription.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Operations/OperationDescription.swift
@@ -144,10 +144,29 @@ extension OperationDescription {
         )
     }
 
-    /// Returns parameters from both the path item level
-    /// and the operation level.
+    /// Merged parameters from both the path item level and the operation level.
+    /// If duplicate parameters exist, only the parameters from the operation level are preserved.
+    ///
+    /// - Returns: An array of merged path item and operation level parameters without duplicates.
+    /// - Throws: When an invalid JSON reference is found.
     var allParameters: [UnresolvedParameter] {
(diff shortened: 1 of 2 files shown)


Sources/_OpenAPIGeneratorCore/Translator/Operations/OperationDescription.swift
```diff
@@ -144,10 +144,29 @@ extension OperationDescription {
...
-    /// and the operation level.
+    /// Merged parameters from both the path item level and the operation level.
+    /// If duplicate parameters exist, only the parameters from the operation level are preserved.
+    ///
+    /// - Returns: An array of merged path item and operation level parameters without duplicates.
+    /// - Throws: When an invalid JSON reference is found.
     var allParameters: [UnresolvedParameter] {
-        pathParameters + operation.parameters
+        get throws { 
+            var mergedParameters = [UnresolvedParameter]()
```
czechboy0, line 154:
```suggestion
            var mergedParameters: [UnresolvedParameter] = []
```

Nit: prefer a literal.

Sources/_OpenAPIGeneratorCore/Translator/Operations/OperationDescription.swift
```diff
@@ -144,10 +144,29 @@ extension OperationDescription {
...
+            var uniqueIdentifiers = Set<String>()
```
czechboy0, line 155:
```suggestion
            var mergedParameters: [UnresolvedParameter] = []
            var uniqueIdentifiers: Set<String> = []
```

Style nit: prefer a literal.

Sources/_OpenAPIGeneratorCore/Translator/Operations/OperationDescription.swift
```diff
@@ -144,10 +144,29 @@ extension OperationDescription {
...
+
+            for parameter in pathParameters + operation.parameters {
+                let resolvedParameter = try parameter.resolve(in: components)
+                let identifier = resolvedParameter.name + resolvedParameter.location.rawValue
```
czechboy0, line 159:
```suggestion
                let identifier = resolvedParameter.location.rawValue + ":" + resolvedParameter.name
```

Nit: it reads better if the larger scope (location) goes first, then the name that only needs to be unique within the location, and let's separate it with a character for easier readability when debugging.

Sources/_OpenAPIGeneratorCore/Translator/Operations/OperationDescription.swift
```diff
@@ -144,10 +144,29 @@ extension OperationDescription {
...
+
+                guard !uniqueIdentifiers.contains(identifier) 
+                else { continue }
```
czechboy0, line 162:
```suggestion
                guard !uniqueIdentifiers.contains(identifier) else { 
                    continue 
                }
```

Style nit.

czechboy0:
The change looks good, just a few minor style suggestions.

For testing, you can create a new unit test file `Tests/OpenAPIGeneratorCoreTests/Translator/Operations/Test_OperationDescription.swift`, follow the style seen e.g. in `Tests/OpenAPIGeneratorCoreTests/Translator/Content/Test_ContentType.swift`, and create an `OperationDescription` with duplicated params in the path and operation in the unit test and verify that `allParameters` returns just the unique ones.

arthurcro:
Thank you for the nice review, I applied all suggestions! 🙂
I'll add a test shortly and mark this as ready for review.

Tests/OpenAPIGeneratorCoreTests/Translator/Operations/Test_OperationDescription.swift
```diff
@@ -0,0 +1,112 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import OpenAPIKit30
+import XCTest
+@testable import _OpenAPIGeneratorCore
+
+final class Test_OperationDescription: Test_Core {
+
+    func testAllParameters_duplicates_retainOnlyOperationParameters() throws {
```
arthurcro, line 20: @czechboy0 I'm not sure about usign the `YamsParser` here. I could not find an easier way to construct an `OperationDescription`. It needs an `OpenAPI.Endpoint` but it does not have a `public` initializer. 
I thought about constructing a `OpenAPI.Document` or `OpenAPI.PathItem`manually but I found it make the test harder to read and understand. What do you think? 🙂

czechboy0: I think constructing `OpenAPI.PathItem` is better than creating a document from YAML. See if you can get that to work to initialize an `OperationDescription`.

czechboy0: From `OpenAPI.PathItem`, you can get the `OpenAPI.PathItem.Endpoint` for your tested operation.

czechboy0:

Tests/OpenAPIGeneratorCoreTests/Translator/Operations/Test_OperationDescription.swift
arthurcro: Alright, thanks for the feedback. I'll work on updating the tests to use that instead!

arthurcro: I've pushed the changes. Thanks to the tests I also noticed I was retaining the path parameter over the operation ones which was incorrect according to https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#path-item-object so I fixed that too!

czechboy0: I think the order was right before, the shared path params should come before the operation specific ones.

If you flip the order back, you should be able to just slightly change your logic - do (pathParams + operationParams), get a reversed(), iterate that, then reverse the result again. I think that should get us the original order but also retain the last occurrence of any location+name pair.

arthurcro: Ah right, the order should be kept as well. I'll adjust that part, thank you!

Tests/OpenAPIGeneratorCoreTests/Translator/Operations/Test_OperationDescription.swift
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    private func _test(_ pathItem: OpenAPI.PathItem) throws -> [UnresolvedParameter] {
+        guard let endpoint = pathItem.endpoints.first else {
+            XCTFail("Unable to retrieve the path item first endpoint.")
+            return []
+        }
+
+        let operationDescription = OperationDescription(
+            path: .init(["\test"]),
```
czechboy0, line 87:
```suggestion
            path: .init(["/test"]),
```

Tests/OpenAPIGeneratorCoreTests/Translator/Operations/Test_OperationDescription.swift
```diff
@@ -0,0 +1,96 @@
...
+                parameters: [operationLevelParameter],
+                requestBody: .b(.init(content: [:])),
+                responses: [:]
+            ),
+            vendorExtensions: [:]
+        )
+        let allParameters = try _test(pathItem)
+
+        XCTAssertEqual(allParameters, [operationLevelParameter])
+    }
```
czechboy0, line 48: Can you please add one more test that verifies the correct ordering?

Could be:
- path item level: A, B
- operation level: B, C
- expected result: A, B, C

czechboy0 (changes requested):
Just two more comments about the tests, otherwise looks great!

PR #184, opened 2023-08-09
Fixes #114 - Add scoped imports

### LINKED ISSUE #114: Replace `@preconcurrency import Foundation` with scoped import
As discussed here: https://github.com/apple/swift-openapi-runtime/pull/22#discussion_r1261117471.

We'd like to import just the types we need from Foundation when using `@preconcurrency import` to help us track which types are lacking senbability.

PARAIPAN9:
### Motivation

Fixes #114
### Modifications

Add scoped imports
### Result

Specific imports
### Test Plan

Run CI and Tests


czechboy0:

czechboy0 (changes requested):
I think it'll also need a change in the generator itself so that it produces these scoped imports instead of one large one for Foundation.

This would be the place to start: https://github.com/apple/swift-openapi-generator/blob/ed61e741d262a894bc1303d2e92506368b3da394/Sources/_OpenAPIGeneratorCore/Translator/CommonTypes/Constants.swift#L29

PARAIPAN9:
The PR  isn't fully prepared at the moment. I've made some changes to ensure I'm heading in the right direction. Currently, the types from the module are being generated in all the files, added in Constants. I need to ensure that the files named Client, Server, and Types have their own separate imports to keep everything  functioning properly and not import unesd types.

czechboy0:
No problem, take your time. Just moved the PR back to draft to better communicate its status.

PARAIPAN9:
Just a heads up to let you know that this was updated.

czechboy0 (approved):
Looks great, thanks! 🙏

PARAIPAN9:
My pleasure!!

PR #185, opened 2023-08-10
[Generator] Accept header

### LINKED ISSUE #160: Add Accept header information to the Input struct
This is a follow-up task to #6.

- Synthesizing an Input parameter that's a type-safe variant of the "accept" header
- The synthesized accept enum should only contain supported content types that we actually generate Output cases for

This way, the server implementer, conforming their type to APIProtocol, will have a clear signal from the user about which content type they want.

Things to keep in mind:
- Client might not care
- Client might specify `*/*`
- Client might specify `application/*`
- Client might specify multiple, and use priorities (the `q` parameter)

The synthesized enum might not be concrete types, but some representation like:

```swift
enum Accept {
  case any
  case anySubtype(String)

  enum SupportedContentTypes {
    case json
    case yaml
  }

  case concrete(SupportedContentTypes)

  // The order here is significant, sorted from most to least desired, according to the rules in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept
  case multiple([SupportedContentTypes]
}
```

This will require a proposal, as there are many options and it needs to be carefully considered.

czechboy0:
### Motivation

Fixes #160.

SOAR-0003 was accepted, this is the generator side of the implementation. The runtime side, which must land first, is at: https://github.com/apple/swift-openapi-runtime/pull/37.

### Modifications

- Adapted the generator logic for the changes - see the file-based reference tests for concrete examples of what the generated `AcceptableContentType` enum looks like.
- Introduced `translateRawRepresentableEnum`, which allows sharing logic between generating this new enum and other string-based enums.
- Explicitly skip parameters that match reserved headers, as dictated by the specification.

### Result

SOAR-0003 as proposed, working behind the `multipleContentTypes` feature flag.

### Test Plan

Adapted PetstoreConsumerTests to test the new behavior, adapter reference tests.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateRawRepresentableEnum.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateRawRepresentableEnum.swift
@@ -14,7 +14,7 @@
 import OpenAPIKit30
 
 extension FileTranslator {
-    
+
     /// Returns a declaration of the specified raw representable enum.
     /// - Parameters:
     ///   - typeName: The name of the type to give to the declared enum.
@@ -35,16 +35,17 @@ extension FileTranslator {
         unknownCaseDescription: String?,
         customSwitchedExpression: (Expression) -> Expression = { $0 }
     ) throws -> Declaration {
-        
+
         let generateUnknownCases = unknownCaseName != nil
-        let knownCases: [Declaration] = cases
+        let knownCases: [Declaration] =
+            cases
             .map { caseName, rawValue in
                 .enumCase(
                     name: caseName,
                     kind: generateUnknownCases ? .nameOnly : .nameWithRawValue(rawValue)
                 )
             }
-        
+
         let otherMembers: [Declaration]
         if let unknownCaseName {
             let undocumentedCase: Declaration = .commentable(
@@ -59,20 +60,20 @@ extension FileTranslator {
             let rawRepresentableInitializer: Declaration
             do {
                 let knownCases: [SwitchCaseDescription] = cases.map { caseName, rawValue in
-                        .init(
-                            kind: .case(.literal(rawValue)),
-                            body: [
-                                .expression(
-                                    .assignment(
-                                        Expression
-                                            .identifier("self")
-                                            .equals(
-                                                .dot(caseName)
-                                            )
-                                    )
+                    .init(
+                        kind: .case(.literal(rawValue)),
+                        body: [
+                            .expression(
+                                .assignment(
+                                    Expression
+                                        .identifier("self")
+                                        .equals(
+                                            .dot(caseName)
+                                        )
                                 )
-                            ]
-                        )
+                            )
+                        ]
+                    )
                 }
                 let unknownCase = SwitchCaseDescription(
                     kind: .default,
@@ -115,18 +116,18 @@ extension FileTranslator {
                     )
                 )
             }
-            
+
             let rawValueGetter: Declaration
             do {
                 let knownCases: [SwitchCaseDescription] = cases.map { caseName, rawValue in
-                        .init(
-                            kind: .case(.dot(caseName)),
-                            body: [
-                                .expression(
-                                    .return(.literal(rawValue))
-                                )
-                            ]
-                        )
+                    .init(
+                        kind: .case(.dot(caseName)),
+                        body: [
+                            .expression(
+                                .return(.literal(rawValue))
+                            )
+                        ]
+                    )
                 }
                 let unknownCase = SwitchCaseDescription(
                     kind: .case(
@@ -148,7 +149,7 @@ extension FileTranslator {
                         )
                     ]
                 )
-                
+
                 let variableDescription = VariableDescription(
                     accessModifier: config.access,
                     kind: .var,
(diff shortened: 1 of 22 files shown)


Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore_FF_MultipleContentTypes/Types.swift
```diff
@@ -1066,8 +1066,48 @@ public enum Operations {
...
+                    case text
+                    case binary
+                    case other(String)
+                    public static var defaultValues: [Operations.getStats.Input.Headers.AcceptableContentType] {
+                        [.json, .text, .binary]
+                    }
+                    public init?(rawValue: String) {
+                        let lowercasedRawValue = rawValue.lowercased()
+                        switch lowercasedRawValue {
+                        case "application/json":
```
gjcairo, line 1080: Should we consider all types ending in `+json` as JSON too?

Tests/PetstoreConsumerTestsFFMultipleContentTypes/Test_Client.swift
```diff
@@ -61,6 +67,48 @@ final class Test_Client: XCTestCase {
...
+                ],
+                encodedBody: #"""
+                    count is 1
+                    """#
+            )
+        }
+        let response = try await client.getStats(
+            .init(
+                headers: .init(accept: [
+                    .init(quality: 0.8, contentType: .json),
```
glbrntt, line 95: This feels backwards, I'd expect to specify the content type and then it's q-factor.

glbrntt (commented):
This seems like a pretty sensible shape. Not sure about the ordering of `quality` and `contentType` though.

Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore_FF_MultipleContentTypes/Types.swift
czechboy0: This string is what the author of the OpenAPI document specified, and we translate it to a Swift name. The logic for doing so is the subject of [SOAR-0002](https://github.com/czechboy0/swift-openapi-generator/blob/hd-soar-0002-content-type-naming/Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0002.md).

We only match the content type documented by the author. So if the author said `application/json`, and the payload's content-type is `application/json`, it'll match. If the payload has `application/vnd.gus+json`, then we will not match it, and it's considered a different content type.

Same the other way around, if the author said to expect `application/vnd.gus+json`, and the payload has anything else (including `application/json`), it will be treated as a different content type.

The strictness is important, as the OpenAPI author might specify both `application/json` and `application/vnd.gus+json` and each would have a different schema, so we need to make sure we keep the separation.

We only use the `+json` checking when deciding how to serialize the payload, and that logic is here: https://github.com/apple/swift-openapi-generator/blob/8f647275acc49a28bd91b3c10c1c2cdd6a2d1e40/Sources/_OpenAPIGeneratorCore/Translator/Content/ContentType.swift#L60

Tests/PetstoreConsumerTestsFFMultipleContentTypes/Test_Client.swift
czechboy0: Ha, you're totally right, I'll flip them.

czechboy0:
@gjcairo @glbrntt if you have more time, would love your thoughts on the associated changes in the runtime library: https://github.com/apple/swift-openapi-runtime/pull/37

simonjbeaumont (commented):
Thanks Honza!

Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore_FF_MultipleContentTypes/Types.swift
```diff
@@ -1066,8 +1066,48 @@ public enum Operations {
...
+                        case .text:
+                            return "text/plain"
+                        case .binary:
+                            return "application/octet-stream"
+                        case .other(let value):
+                            return value
+                        }
+                    }
+                }
+                public typealias Accept = [AcceptHeaderContentType<AcceptableContentType>]
```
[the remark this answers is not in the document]
czechboy0: Removed in the last iteration.

Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateRawRepresentableEnum.swift
```diff
@@ -0,0 +1,216 @@
...
+
+extension FileTranslator {
+
+    /// Returns a declaration of the specified raw representable enum.
+    /// - Parameters:
+    ///   - typeName: The name of the type to give to the declared enum.
+    ///   - conformances: The list of types the enum conforms to.
+    ///   - userDescription: The contents of the documentation comment.
+    ///   - cases: The list of cases to generate.
+    ///   - unknownCaseName: The name of the extra unknown case that preserves
```
gjcairo, line 24: Should we document that, if absent, we won't generate unknown cases?

glbrntt (approved):
Looks great!

Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateRawRepresentableEnum.swift
czechboy0: Will do.

czechboy0: Fixed in https://github.com/apple/swift-openapi-generator/pull/185/commits/b9ace9a1bcb9b9b9ec9fc96a40ffa4c6533efbe0

gjcairo (approved):

PR #188, opened 2023-08-10
Fixes #165 - Enable documentation comment validation in swift-format 

### LINKED ISSUE #165: Enable documentation comment validation in swift-format
(This is about the codebase of the generator, runtime, and transports, not about generated code.)

We should enable the following two rules in the `.swift-format` file of all our repos, to get more validation of documentation comments. I sometimes forget to rename properties in doc comments and this could help catch that. Plus it'll ensure that new code is documented properly.

- `ValidateDocumentationComments`
- `AllPublicDeclarationsHaveDocumentation`
- possibly other comments-related ones

PARAIPAN9:
### Motivation

This is for better documentation
Fixes #165

### Modifications

Enabled some rules in .swift-format

### Result

The .swift-format will be changed and the rules will reflect into the project

### Test Plan

Run tests and CI


czechboy0:
Sorry if the issue wasn't clear, part of the enablement will likely need to be to actually update all the places where the docs aren't passing validation. Otherwise our CI would go red.

PARAIPAN9:
There is also this rule: BeginDocumentationCommentWithOneLineSummary that is comment related, currently set to false.

.swift-format
```diff
@@ -16,7 +16,7 @@
   "prioritizeKeepingFunctionOutputTogether" : false,
   "respectsExistingLineBreaks" : true,
   "rules" : {
-    "AllPublicDeclarationsHaveDocumentation" : false,
+    "AllPublicDeclarationsHaveDocumentation" :true,
```
czechboy0, line 19:
```suggestion
    "AllPublicDeclarationsHaveDocumentation" : true,
```

Nit: whitespace.

czechboy0 (changes requested):
This will need updating the actual places missing docs (try running `./scripts/soundness` locally). That might be quite a bit of work, so if that seems like too much, feel free to put the issue back into the queue and someone else will pick it up later.

czechboy0:
> There is also this rule: BeginDocumentationCommentWithOneLineSummary that is comment related, currently set to false.

I don't think we need to set that, it's overly strict about the abstract of each symbol. Could be added later.

czechboy0:
Just setting the PR back to draft to communicate its status.

PARAIPAN9:
Just a heads up to let you know that I'm still working on it, it will take a little bit of time.

czechboy0:
No rush, @PARAIPAN9 – thanks 🙏

PARAIPAN9:
Hello, I have another update about this PR. I've come to a stage where I need to include comment documentation for the generated files and by doing this there is reference test that is failing. What would be the best approach here? Should I modify the generator to produce the documentation, similar to how it does it for "Remark" comments?

czechboy0:
No, let's exclude the files from being reformatted and linted. They get formatted as part of the generation process anyway, so this doesn't remove any value.

I suspect updating the script for formatting should do the trick.

czechboy0:
@PARAIPAN9 Once you'd like this reviewed, please move the PR into Ready for review.

scripts/run-swift-format.sh
```diff
@@ -23,16 +23,17 @@ REPO_ROOT="$(git -C "${CURRENT_SCRIPT_DIR}" rev-parse --show-toplevel)"
 
 SWIFTFORMAT_BIN=${SWIFTFORMAT_BIN:-$(command -v swift-format)} || fatal "❌ SWIFTFORMAT_BIN unset and no swift-format on PATH"
 
-git -C "${REPO_ROOT}" ls-files -z '*.swift' \
-  | xargs -0 "${SWIFTFORMAT_BIN}" lint --parallel --strict \
+"${SWIFTFORMAT_BIN}" lint \
+  --parallel --recursive --strict \
+  "${REPO_ROOT}/Sources" \
```
czechboy0, line 28: What was the issue with before? I think this would disable linting the Test and other directories, which I don't think we want. cc @simonjbeaumont

czechboy0 (changes requested):
Just one question about the script, otherwise looks great, thank you!

scripts/run-swift-format.sh
PARAIPAN9: The problem was that it was also complaining about the generated files.

czechboy0: We should then only exclude generated files but keep all the other ones. There are examples, integration tests and unit tests that we want to cover.

PARAIPAN9: Yes that's right, then I’ll tweak the script to exclude only the generated files.

scripts/run-swift-format.sh
```diff
@@ -24,6 +24,7 @@ REPO_ROOT="$(git -C "${CURRENT_SCRIPT_DIR}" rev-parse --show-toplevel)"
 SWIFTFORMAT_BIN=${SWIFTFORMAT_BIN:-$(command -v swift-format)} || fatal "❌ SWIFTFORMAT_BIN unset and no swift-format on PATH"
 
 git -C "${REPO_ROOT}" ls-files -z '*.swift' \
+  | grep -z -v 'Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore' \
```
czechboy0, line 27:
```suggestion
  | grep -z -v 'Tests/OpenAPIGeneratorReferenceTests/Resources' \
```

And please also exclude and remove any changes to the Sources/swift-openapi-generator/Documentation.docc. Otherwise will be ready to go.

czechboy0 (changes requested):

czechboy0 (approved):
Thank you, @PARAIPAN9!

czechboy0:
@PARAIPAN9 If you'd be interested, it'd be good to get this change into swift-openapi-runtime as well, but no pressure 🙂

PARAIPAN9:
Sure, no problem!!

PR #189, opened 2023-08-11
Ignore inline schemas in oneOf with discriminator

### LINKED ISSUE #181: Filter out inline schemas in oneOf + discriminator
Today, we skip oneOf that have a discriminator and also contain at least one inline schema.

The OpenAPI specification (https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#discriminator-object) however prescribes softer handling of this invalid case:

> When using the discriminator, inline schemas will not be considered.

We should filter out the inline schemas rather than skipping the whole oneOf to be more compliant with the specification.

simonjbeaumont:
### Motivation

Currently, when provided with a `oneOf` with a `discriminator`, the generator will fail if any of the schemas are not references. However, the OpenAPI specification states that inline schemas should be ignored when using a discriminator:

> When using the discriminator, inline schemas will not be considered.
> — https://spec.openapis.org/oas/v3.0.3#discriminator-object

The generator shouldn't fail when presented with a spec-compliant OpenAPI document, however misguided it might be.

### Modifications

Filter out inline schemas during schema validation and translation.

### Result

Can handle documents with `oneOf` types with a discriminator, containing inline schemas.

### Test Plan

- Updated unit tests for supported/unsupported schemas.
- Updated snippet test with an example, which failed before this patch.

### Resolves

- Fixes #181

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateAllAnyOneOf.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateAllAnyOneOf.swift
@@ -124,9 +124,17 @@ extension FileTranslator {
         discriminator: OpenAPI.Discriminator?,
         schemas: [JSONSchema]
     ) throws -> Declaration {
+        // > When using the discriminator, inline schemas will not be considered.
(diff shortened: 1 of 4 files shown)


czechboy0 (approved):
Looks great, thanks!

PR #170, opened 2023-08-04
SOAR-0002: Improved naming of content types

czechboy0:
### Motivation

See proposal text.

### Modifications

N/A

### Result

N/A

### Test Plan

N/A

PR #194, opened 2023-08-14
SOAR-0002 implementation

czechboy0:
### Motivation

Implements the approved SOAR-0002 proposal: https://github.com/apple/swift-openapi-generator/pull/170

### Modifications

- updates the function that computes content type names to be in line with v3 of SOAR-0002
- refactors the `ContentType` type a bit to make it harder to accidentally perform case-sensitive comparions where case-insensitive ones are required

### Result

When an adopter passes the `multipleContentTypes` feature flag, they'll get the new logic as described in SOAR-0002.

### Test Plan

- adapted unit tests for the `contentSwiftName` function
- enabled the `proposal0001` feature flag as well, since they will be enabled together in 0.2.0, so makes sense to test the new world together
- this lead to a few more updates needed in the reference tests that weren't caused by 0.2.0 alone


glbrntt (approved):
LGTM

Issue #192, opened 2023-08-12
Create a URI encoder and decoder that works with Codable types


czechboy0:
Create a URI encoder and decoder that works with Codable types.

More details in this comment: https://github.com/apple/swift-openapi-generator/issues/182#issuecomment-1675870562

bfrearson:
This is definitely needed, but before we continue it would be good to reach a consensus on what we should do with a nested type such as: 

```
struct NestedType: Encodable {
    let name: String
    let ship: Ship
}

struct Ship: Encodable {
    let designation: String
    let crew: [String]
    let type: String
}
```

The expected output of our encoder, I assume, would always be key-value pairs in the form `key1=value1&key2=value2`. There is nothing in the openapi spec that prevents a schema like the example above from being used here, even though it doesn't match the architecture, so we need to have a way of handling this, even if it is bad practice.

There's multiple ways that we could handle these types, and in the long run it would be good to support a few different methods. However, I think for the moment, we should decide on a single implementation and expand from there.

What should the output of our encoder look like for a type that itself contains other Codable types like the example above (percent encoding omitted for clarity)?

- dot syntax: `name=Han Solo&ship.designation=Millennium Falcon&ship.crew=Chewie, Leia, C3PO&ship.type=YT-1300F light freighter`
- JSON serialized strings for lower-level data: `name=Han Solo&ship={crew = ( "Chewie, Leia, C3PO"); designation = "Millenium Falcon"; type = "YT-1300F light freighter";}`
- error: "url encoded forms must use a flat data structure"
- Something else?

For the very short term, I'd suggest we error out until the initial encoder is written, but I imagine that's not acceptable in the longer run.

czechboy0:
I'd like us to support any arbitrarily nested Codable type, as it's not going to be that much more work than just top level items. I have a design in mind, let me try to put together a prototype end of this or next week and we can take it from there.

czechboy0:
Early encoder prototype I'm working on: https://github.com/apple/swift-openapi-runtime/pull/41

czechboy0:
So, after actually reading the 6570 and 1866 RFCs, turns out nesting containers is not a supported feature. All the nesting approaches seem to be conventions that people invented on top.

The first implementation of URIEncoder/URIDecoder will _not_ support nesting containers, but I'd be open to enhancing it in the future to support nesting, assuming we can have some confidence that what we implement is actually used widely.

bfrearson:
[...]

Issue #182, opened 2023-08-09
Support URL form encoded request bodies


benfrearson:
There may be a legitimate reason why this isn't already implemented, or perhaps I'm not using the generator for its intended use case - if so, I'd appreciate some clarity.

It appears to me that URL form encoded requests have their data structure erased in the generated Input.body type, which blindly expects `Data`.

``` 
# openapi.yaml
paths:
  /login:
    post:
      tags:
        - Auth
      summary: Login
      description: Login
      operationId: login
      requestBody:
        required: true
        content:
          application/x-www-form-urlencoded:
            schema:
              type: object
              required:
                - device_name
                - email
                - password
              properties:
                device_name:
                  type: string
                email:
                  type: string
                password:
                  type: string
```

```
//Types.swift

//...

/// API operations, with input and output types, generated from `#/paths` in the OpenAPI document.
public enum Operations {
    /// Login
    ///
    /// Login
    ///
    /// - Remark: HTTP `POST /login`.
    /// - Remark: Generated from `#/paths//login/post(login)`.
    public enum login {
        public static let id: String = "login"
        public struct Input: Sendable, Equatable, Hashable {
            
           // ...

            @frozen public enum Body: Sendable, Equatable, Hashable { case binary(Foundation.Data) }
            public var body: Operations.login.Input.Body
            /// Creates a new `Input`.
            ///
            /// - Parameters:
            ///   - path:
            ///   - query:
            ///   - headers:
            ///   - cookies:
            ///   - body:
            public init(
                path: Operations.login.Input.Path = .init(),
                query: Operations.login.Input.Query = .init(),
                headers: Operations.login.Input.Headers = .init(),
                cookies: Operations.login.Input.Cookies = .init(),
                body: Operations.login.Input.Body
            ) {
                self.path = path
                self.query = query
                self.headers = headers
                self.cookies = cookies
                self.body = body
            }
        }
// ...
}
```

The issue here is that using the corresponding `Client` method now is unaware of the required url form fields. This is obviously fairly easy to work around, but it obfuscates the expected data from the users of the `Client` in a way that using a JSON encoded body does not.

It would be better if the generated Input type could handle encoding form bodies itself, such as by:
- Having an `Input.body` initializer that accepts arguments as dictated by the form fields.
- Having an intermediate property for the key-value pairs that is encoded when serializing the request similar to the `converter.setRequiredRequestBodyAsBinary()` method.

[...]

PR #195, opened 2023-08-14
Support url form encoded bodies

bfrearson:
### Motivation

Opening this as a preliminary approach to adding support for url encoded forms #182.

I notice that we're probably going to want to take a different approach to this implementation by first building out a custom URLFormEncoder (#192), so this might be mostly redundant at this point, but wanted to at least document this here as my first contribution to the project.

It may be that work on #192 can be merged into the `Converter` methods in the associated [runtime changes](https://github.com/apple/swift-openapi-runtime/pull/39) 

### Modifications

Add a new `urlEncodedForm` content type to support type safe x-www-form-urlencoded requests.

### Result

Url form requests will take a Codable object type, rather than a raw Data input

### Test Plan

- Updated and passed reference tests with example url form request
- Not yet added additional unit tests as this PR is only intended as a draft for feedback

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/CommonTypes/Constants.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/CommonTypes/Constants.swift
@@ -354,6 +354,9 @@ enum Constants {
 
         /// The substring used in method names for the binary coding strategy.
         static let binary: String = "Binary"
+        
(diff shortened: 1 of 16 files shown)


czechboy0:
Hi @benfrearson – thanks for getting started on this! Let me read it all and drop some comments, then we can figure out how we approach staging this support in.

PR #196, opened 2023-08-15
Emit JSON path reference in comments for all generated types

### LINKED ISSUE #13: Emit JSON path reference in comments for all generated types
This already works for many types and properties, but not for all.

We should ensure that all content from the OpenAPI doc generated in Swift includes a comment linking it back to its original JSON path in the OpenAPI doc.

For example:

```swift
/// - Remark: Generated from `#/paths/\/pets\/{petId}/patch/requestBody/content/application\/json/schema`.
```

takeshi-1000:
### Motivation

Resolve #13

### Modifications

Applied commentable to the declaration variable in Traslator so that a comment in Types.swift is added indicating that it was generated from the Json path of the openapi document

### Result

Some comments in Types.swift is generated, like `Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Types.swift`.

### Test Plan

Modify Type.swift in the reference source

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/Parameters/translateParameter.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Parameters/translateParameter.swift
@@ -43,7 +43,11 @@ extension TypesFileTranslator {
         } else {
             associatedDeclarations = []
         }
+        
+        let comment: Comment? = parent.docCommentWithUserDescription(nil,
+                                                                    subPath: "\(parameter.location.rawValue)/\(parameter.name)")
         return .init(
+            comment: comment,
             isDeprecated: parameter.parameter.deprecated,
             originalName: parameter.name,
             typeUsage: parameter.typeUsage,
--- a/Sources/_OpenAPIGeneratorCore/Translator/Parameters/translateParameter.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Parameters/translateParameter.swift
@@ -43,7 +43,7 @@ extension TypesFileTranslator {
         } else {
             associatedDeclarations = []
         }
-        
+
         let comment: Comment? = parent.docCommentWithUserDescription(
             nil,
             subPath: "\(parameter.location.rawValue)/\(parameter.name)"
--- a/Sources/_OpenAPIGeneratorCore/Translator/Parameters/translateParameter.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Parameters/translateParameter.swift
@@ -44,8 +44,10 @@ extension TypesFileTranslator {
             associatedDeclarations = []
         }
         
-        let comment: Comment? = parent.docCommentWithUserDescription(nil,
-                                                                    subPath: "\(parameter.location.rawValue)/\(parameter.name)")
+        let comment: Comment? = parent.docCommentWithUserDescription(
+            nil,
+            subPath: "\(parameter.location.rawValue)/\(parameter.name)"
+        )
         return .init(
             comment: comment,
             isDeprecated: parameter.parameter.deprecated,
--- a/Sources/_OpenAPIGeneratorCore/Translator/Parameters/translateParameter.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Parameters/translateParameter.swift
@@ -43,13 +43,8 @@ extension TypesFileTranslator {
         } else {
             associatedDeclarations = []
         }
-
-        let comment: Comment? = parent.docCommentWithUserDescription(
-            nil,
-            subPath: "\(parameter.location.rawValue)/\(parameter.name)"
-        )
         return .init(
-            comment: comment,
+            comment: parameter.docComment(parent: parent),
             isDeprecated: parameter.parameter.deprecated,
             originalName: parameter.name,
             typeUsage: parameter.typeUsage,
--- a/Sources/_OpenAPIGeneratorCore/Translator/RequestBody/translateRequestBody.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/RequestBody/translateRequestBody.swift
@@ -41,6 +41,7 @@ extension TypesFileTranslator {
     /// - Returns: A list of declarations; empty if the request body is
     /// unsupported.
     func requestBodyContentCases(
+        typeName: TypeName,
         for requestBody: TypedRequestBody
     ) throws -> [Declaration] {
         var bodyMembers: [Declaration] = []
@@ -54,13 +55,20 @@ extension TypesFileTranslator {
             }
             let identifier = contentSwiftName(content.content.contentType)
             let associatedType = content.resolvedTypeUsage
-            let contentCase: Declaration = .enumCase(
-                .init(
+            
+            let subPath: String = {
+                let contentPath = typeName.isComponent ? "content" : "requestBody/content"
+                return "\(contentPath)/\(content.content.contentType.lowercasedTypeAndSubtypeWithEscape)"
+            }()
+            
+            let contentCase: Declaration = .commentable(
+                typeName.docCommentWithUserDescription(nil, subPath: subPath),
+                .enumCase(.init(
                     name: identifier,
                     kind: .nameWithAssociatedValues([
                         .init(type: associatedType.fullyQualifiedNonOptionalSwiftName)
                     ])
-                )
+                ))
             )
             bodyMembers.append(contentCase)
         }
@@ -131,7 +139,8 @@ extension TypesFileTranslator {
         requestBody: TypedRequestBody
     ) throws -> Declaration {
         let type = requestBody.typeUsage.typeName
-        let members = try requestBodyContentCases(for: requestBody)
+        let members = try requestBodyContentCases(typeName: type,
+                                                  for: requestBody)
         return translateRequestBodyInTypes(
             typeName: type,
             members: members
@@ -154,7 +163,14 @@ extension TypesFileTranslator {
             conformances: Constants.Operation.Output.conformances,
             members: members
         )
-        return bodyEnumDecl
+        let comment: Comment? = {
+            if typeName.isComponent {
+                return typeName.docCommentWithUserDescription(nil)
+            } else {
+                return typeName.docCommentWithUserDescription(nil, subPath: "requestBody")
+            }
+        }()
+        return .commentable(comment, bodyEnumDecl)
     }
 }
 
--- a/Sources/_OpenAPIGeneratorCore/Translator/RequestBody/translateRequestBody.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/RequestBody/translateRequestBody.swift
@@ -41,7 +41,6 @@ extension TypesFileTranslator {
     /// - Returns: A list of declarations; empty if the request body is
     /// unsupported.
     func requestBodyContentCases(
-        typeName: TypeName,
         for requestBody: TypedRequestBody
     ) throws -> [Declaration] {
         var bodyMembers: [Declaration] = []
@@ -63,7 +62,7 @@ extension TypesFileTranslator {
             
             let contentCase: Declaration = .commentable(
                 typeName.docCommentWithUserDescription(nil, subPath: subPath),
-                .enumCase(.init(
+                .enumCase(
                     name: identifier,
                     kind: .nameWithAssociatedValues([
                         .init(type: associatedType.fullyQualifiedNonOptionalSwiftName)
@@ -139,8 +138,7 @@ extension TypesFileTranslator {
         requestBody: TypedRequestBody
     ) throws -> Declaration {
         let type = requestBody.typeUsage.typeName
-        let members = try requestBodyContentCases(typeName: type,
-                                                  for: requestBody)
+        let members = try requestBodyContentCases(for: requestBody)
         return translateRequestBodyInTypes(
             typeName: type,
             members: members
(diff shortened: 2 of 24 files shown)


Sources/_OpenAPIGeneratorCore/Translator/Parameters/translateParameter.swift
```diff
@@ -43,7 +43,11 @@ extension TypesFileTranslator {
         } else {
             associatedDeclarations = []
         }
+        
+        let comment: Comment? = parent.docCommentWithUserDescription(nil,
+                                                                    subPath: "\(parameter.location.rawValue)/\(parameter.name)")
```
czechboy0, line 48:
```suggestion
        let comment: Comment? = parent.docCommentWithUserDescription(
            nil,
            subPath: "\(parameter.location.rawValue)/\(parameter.name)"
        )
```

Nit: whitespace

Sources/_OpenAPIGeneratorCore/Translator/RequestBody/translateRequestBody.swift
```diff
@@ -41,6 +41,7 @@ extension TypesFileTranslator {
     /// - Returns: A list of declarations; empty if the request body is
     /// unsupported.
     func requestBodyContentCases(
+        typeName: TypeName,
```
czechboy0, line 44:
```suggestion
```

You shouldn't need to pass this type in, it's already present on `requestBody.typeUsage.typeName`.

Sources/_OpenAPIGeneratorCore/Translator/RequestBody/translateRequestBody.swift
```diff
@@ -54,13 +55,20 @@ extension TypesFileTranslator {
...
-                .init(
+            
+            let subPath: String = {
+                let contentPath = typeName.isComponent ? "content" : "requestBody/content"
+                return "\(contentPath)/\(content.content.contentType.lowercasedTypeAndSubtypeWithEscape)"
+            }()
+            
+            let contentCase: Declaration = .commentable(
+                typeName.docCommentWithUserDescription(nil, subPath: subPath),
+                .enumCase(.init(
```
czechboy0, line 66:
```suggestion
                .enumCase(
```

I think we have a utility function on `Declaration` that allows passing in the parameters directly, don't we?

Sources/_OpenAPIGeneratorCore/Translator/RequestBody/translateRequestBody.swift
```diff
@@ -131,7 +139,8 @@ extension TypesFileTranslator {
         requestBody: TypedRequestBody
     ) throws -> Declaration {
         let type = requestBody.typeUsage.typeName
-        let members = try requestBodyContentCases(for: requestBody)
+        let members = try requestBodyContentCases(typeName: type,
+                                                  for: requestBody)
```
czechboy0, line 143:
```suggestion
        let members = try requestBodyContentCases(
            typeName: type,
            for: requestBody
        )
```

Nit: whitespace

czechboy0, line 143:
```suggestion
        let members = try requestBodyContentCases(for: requestBody)
```

As commented above, I don't think we need to pass it in, the type name should already be present on `requestBody`.

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponseHeader.swift
```diff
@@ -22,10 +22,18 @@ extension TypesFileTranslator {
     ///   - header: A response parameter.
     /// - Returns: A property blueprint.
     func parseResponseHeaderAsProperty(
+        responseKind: String = "",
```
czechboy0, line 25:
```suggestion
        responseKind: ResponseKind,
```

If the response kind is needed, pass it in as the typesafe type `ResponseKind` and don't provide a default value, we want to make sure all callers are updated.

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
```diff
@@ -34,10 +35,19 @@ extension TypesFileTranslator {
...
-            try parseResponseHeaderAsProperty(for: header)
+            try parseResponseHeaderAsProperty(
+                responseKind: responseKind,
+                typeName: typeName,
+                for: header)
         }
+        let headerStructComment: Comment? = {
+            let subPath = typeName.isComponent
+                  ? "headers"
+                  : "reponses/\(responseKind)/headers"
```
czechboy0, line 46:
```suggestion
                  : "responses/\(responseKind)/headers"
```

Typo

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
```diff
@@ -20,6 +20,7 @@ extension TypesFileTranslator {
     ///   - typedResponse: The typed response to declare.
     /// - Returns: A structure declaration.
     func translateResponseInTypes(
+        responseKind: String = "",
```
czechboy0, line 23:
```suggestion
        responseKind: ResponseKind,
```

As below, let's pass in the typesafe variant and not provide a default.

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponseHeader.swift
```diff
@@ -22,10 +22,18 @@ extension TypesFileTranslator {
...
+        typeName: TypeName,
         for header: TypedResponseHeader
     ) throws -> PropertyBlueprint {
+        let comment: Comment? = {
+            let subPath: String = typeName.isComponent
+            ? "headers/\(header.name)"
+            : "reponses/\(responseKind)/headers/\(header.name)"
```
czechboy0, line 32:
```suggestion
            : "responses/\(responseKind)/headers/\(header.name)"
```

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponseOutcome.swift
```diff
@@ -42,6 +42,7 @@ extension TypesFileTranslator {
         let responseStructDecl: Declaration?
         if typedResponse.isInlined {
             responseStructDecl = try translateResponseInTypes(
+                responseKind: responseKind.jsonPathComponent,
```
czechboy0, line 45:
```suggestion
                responseKind: responseKind,
```

Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeName.swift
```diff
@@ -126,6 +126,10 @@ struct TypeName: Equatable {
         precondition(components.count >= 1, "Cannot get the parent of a root type")
         return .init(components: components.dropLast())
     }
+    
+    var isComponent: Bool {
+        description.contains("#/components")
```
czechboy0, line 131: I think you could instead check if the first two JSON components are `#` and `components`, that'd at least use the internal `components` storage and be a little more reliable.

czechboy0 (changes requested):
This is definitely the right direction! Added a few comments, we can get this PR cleaned up and landed, and then you can continue to address any remaining missing comments in a separate PR.

takeshi-1000:
Thanks for the quick confirmation! The fix is ​​still incomplete, but I just put in something that can be applied immediately.

czechboy0:

czechboy0:
> Thanks for the quick confirmation! The fix is ​​still incomplete, but I just put in something that can be applied immediately.

Sure – can you please just resolve the CI failures and we could land this PR? Then you can continue to work on the missing bits in a separate PR.

Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeName.swift
takeshi-1000: Exactly. I fixed.
https://github.com/apple/swift-openapi-generator/pull/196/commits/4188460d7e724641a0bda1a85e42990d2a26f613#diff-9d2dcac6bf1aa7b2aee70f0b1ff4de6be4efe7e702a63f0145012a4f45a31ac9

takeshi-1000:
@czechboy0 
I modify some code, could you check please? 
Also As you suggested, I plan to address some other issues in another PR.
(There are places where userDescription is not reflected in comments, and characters are implemented with hard code)
Is it okay to open these issues?

Sources/_OpenAPIGeneratorCore/Translator/RequestBody/translateRequestBody.swift
```diff
@@ -54,8 +54,18 @@ extension TypesFileTranslator {
             }
             let identifier = contentSwiftName(content.content.contentType)
```
czechboy0, line 55:
```suggestion
            let contentType = content.content.contentType
            let identifier = contentSwiftName(contentType)
```

Sources/_OpenAPIGeneratorCore/Translator/RequestBody/translateRequestBody.swift
```diff
@@ -54,8 +54,18 @@ extension TypesFileTranslator {
...
+
+            let subPath: String = {
+                let contentPath = requestBody.typeUsage.typeName.isComponent ? "content" : "requestBody/content"
+                return "\(contentPath)/\(content.content.contentType.lowercasedTypeAndSubtypeWithEscape)"
+            }()
+
+            let contentCase: Declaration = .commentable(
+                requestBody
+                    .typeUsage
+                    .typeName
```
czechboy0, line 66:
```suggestion
            let typeName = requestBody.typeUsage.typeName
            let subPath: String = {
                let contentPath = (typeName.isComponent ? "" : "requestBody/") + "content"
                return "\(contentPath)/\(contentType.lowercasedTypeAndSubtypeWithEscape)"
            }()
            let contentCase: Declaration = .commentable(
                typeName
```

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
```diff
@@ -20,6 +20,7 @@ extension TypesFileTranslator {
     ///   - typedResponse: The typed response to declare.
     /// - Returns: A structure declaration.
     func translateResponseInTypes(
+        responseKind: ResponseKind?,
```
czechboy0, line 23: Actually, I wonder - don't we already have this information in `typeName`'s JSON components? Why do we need to pass `responseKind`, especially since it's optional so won't always be there?

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
```diff
@@ -76,22 +88,45 @@ extension TypesFileTranslator {
...
+                )
             )
             bodyCases.append(bodyCase)
         }
+        let subPathForContent: String = {
+            if let responseKind = responseKind?.jsonPathComponent, !typeName.isComponent {
+                return "responses/\(responseKind)/content"
+            } else {
+                return "content"
+            }
```
czechboy0, line 116: There's quite a lot of duplication of prefixes of the comments, could we make it a bit more readable by building it up gradually using some local variables, and concatenating more components as needed? Otherwise it's difficult to read how the different branches of the if/else are different.

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
```diff
@@ -136,6 +171,7 @@ extension TypesFileTranslator {
             of: OpenAPI.Response.self
         )
         return try translateResponseInTypes(
+            responseKind: nil,
```
czechboy0, line 174: Well, this is unfortunate. Now I wonder if we don't already have this info in `typeName`?

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponseHeader.swift
```diff
@@ -22,10 +22,22 @@ extension TypesFileTranslator {
...
+        let comment: Comment? = {
+            let subPath: String = {
+                if let responseKind = responseKind?.jsonPathComponent, !typeName.isComponent {
+                    return "responses/\(responseKind)/headers/\(header.name)"
+                } else {
+                    return "headers/\(header.name)"
+                }
+            }()
+            return typeName.docCommentWithUserDescription(nil, subPath: subPath)
+        }()
```
czechboy0, line 38: If the comment is the only reason we have to start passing these other two params, let's move the creation of the comment to the caller of this method, this muddies up the purpose of the method.

Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeName.swift
```diff
@@ -126,6 +126,14 @@ struct TypeName: Equatable {
         precondition(components.count >= 1, "Cannot get the parent of a root type")
         return .init(components: components.dropLast())
     }
+
+    /// Returns a bool value indicating whether the list of JSON path components contains "#" first and "components" second.
+    var isComponent: Bool {
```
czechboy0, line 131:
```suggestion
    var isInComponents: Bool {
```

czechboy0 (changes requested):
A few more stylistic nits to ensure better readability.

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
takeshi-1000: I think so, I could not find something which access responseKind from typeName or typeResponse.

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
takeshi-1000: I try to fix. https://github.com/apple/swift-openapi-generator/pull/196/commits/5ff67b8a1795b52ceefeae5d2c840008642e96ff

czechboy0 (approved):
Updated with one more commit, lgtm.

takeshi-1000:
Thank you for suporting!

Issue #197, opened 2023-08-15
Range responses appear to lead to non-compiling code on the server


czechboy0:
Range responses appear to lead to non-compiling code on the server.

PR #198, opened 2023-08-15
Fix range responses

czechboy0:
### Motivation

Fixes https://github.com/apple/swift-openapi-generator/issues/197, broken range responses, such as `4xx`.

### Modifications

Fix the bug, seems to have been some old copy paste gone wrong, and we didn't have a test for it.

### Result

Now range responses work again.

### Test Plan

Changed one of the status code responses to be a range response, updated fixtures.


czechboy0:

PR #200, opened 2023-08-16
[Docs] Clarify what constitutes a breaking change of generated code

czechboy0:
### Motivation

As new features are introduced to the generator, it is limiting to not be able to add new defaulted properties to generated structs without it being an API-breaking change.

### Modifications

Clarify that adding a defaulted property to a generated struct is not considered a breaking change in the Swift OpenAPI Generator project, and adopters should avoid capturing initializers of generated structs for that reason.

### Result

With this clarification, we can add new defaulted properties without breaking API.

### Test Plan

N/A

### DIFF
--- a/Sources/swift-openapi-generator/Documentation.docc/Articles/API-stability-of-generated-code.md
+++ b/Sources/swift-openapi-generator/Documentation.docc/Articles/API-stability-of-generated-code.md
@@ -27,7 +27,7 @@ Below is a table of example changes you might make to an OpenAPI document, and w
 | Remove a required property | ❌ | ❌ | ❌ |
 | Rename a schema | ✅ | ❌ | ❌ |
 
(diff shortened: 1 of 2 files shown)


Sources/swift-openapi-generator/Documentation.docc/Articles/API-stability-of-generated-code.md
```diff
@@ -27,7 +27,7 @@ Below is a table of example changes you might make to an OpenAPI document, and w
 | Remove a required property | ❌ | ❌ | ❌ |
 | Rename a schema | ✅ | ❌ | ❌ |
 
-> †: Safe change to make as long as no adopter captured the Swift function signature of the initializer of the generated struct, which gains a new parameter. Rare, but something to be aware of.
+> †: Safe change to make as long as no adopter captured the Swift function signature of the initializer of the generated struct, which gains a new parameter. Rare, but something to be aware of. Note that when upgrading the generator to a newer version, we reserve the right to add new defaulted properties to generated structs, so such a change is considered non-breaking. For that reason, avoid capturing the function signature of generated structs.
```
FranzBusch, line 30: > function signature of generated structs

Do you mean the function signature of generated initialisers?

glbrntt (approved):
I think this trade-off is fine. The chances of users taking a function ref to inits is low and it helps reduce the complexity of the generator.

Issue #204, opened 2023-08-16
Stop generating an undocumented case for enums/oneOfs


czechboy0:
Consider not generating an undocumented case for enums/oneOfs by default.

It seemed like a good idea to generate the undocumented case before, as it allows for API evolution - as adding an enum or oneOf case maybe shouldn't be a breaking change.

However, OpenAPI/JSON Schema enums/oneOfs are closed by default, so we're actually diverging from the specification here.

And it's causing issues when these enums/oneOfs are nested within other allOfs/oneOfs/anyOfs and it "successfully" parses even though it parses into an undocumented case.

Talk about this more, but we might need to undo this, and follow the specification closely, and also generate closed enums by default. Possibly it could be an option to generate an undocumented case? Or we should document that open enums can be represented as:

```yaml
MyOpenEnum:
  anyOf:
    - type: string
      enum:
        - foo
        - bar
        - baz
    - type: string
```

And an open oneOf with:

```yaml
MyOpenOneOf:
  anyOf:
    - oneOf:
        - #/components/schemas/Foo
        - #/components/schemas/Bar
        - #/components/schemas/Baz
    - {}
```

And as an adopter, you first check the first schema if it's not nil. If it's not nil, the closed enum/oneOf parsed successfully. If the first element is nil, the second will be filled in, which is the fallback value.

czechboy0:
It would simplify the composition of types to be able to stop generating open enums/oneOfs, especially since adopters have a pattern to work around it. Plus, we're not specification-compliant now. So we should probably remove this.

iDevPro:
Hello, how to build enum Int raw value but named cases like

```swift
enum SizeValue: Int {
    case mini = 10
    case mid = 30
    case max = 100
}
```

Thx for answer :)

PR #205, opened 2023-08-16
[Generator] Stop generating an undocumented case for enums/oneOfs

czechboy0:
### Motivation

Fixes https://github.com/apple/swift-openapi-generator/issues/204.

Depends on https://github.com/apple/swift-openapi-runtime/pull/40.

### Modifications

This introduces the feature flag `closedEnumsAndOneOfs` which stops generating the `undocumented` case for enums and oneOfs (no other places are affected, such as responses, and so on.). The default behavior in 0.1.x stays the same, but in 0.2.0 we'll stop generating `undocumented` cases unconditionally.

This fixes our divergence from the OpenAPI specification, as we treated enums and oneOfs as open, but they're closed by default.

I also added docs on this pattern to make it easier for adopters to know how to create open enums and oneOfs on purpose.

### Result

We get closer to being spec-compliant.

### Test Plan

Created new snippet tests to verify the new functionality, will adapt file-based tests when we're preparing 0.2.0.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -78,7 +78,7 @@ let package = Package(
         ),
 
         // Tests-only: Runtime library linked by generated code
-        .package(url: "https://github.com/apple/swift-openapi-runtime", .upToNextMinor(from: "0.1.8")),
+        .package(url: "https://github.com/apple/swift-openapi-runtime", .upToNextMinor(from: "0.1.9")),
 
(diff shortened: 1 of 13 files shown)


glbrntt (approved):
LGTM. Helpful docs too!

gjcairo (approved):

Issue #211, opened 2023-08-18
Additional 'case' blocks cannot appear after the 'default' block of a 'switch'


dpasirst:
It's possible the upstream tool is the root cause as it's using Aide in Rust creating OpenAPI 3.1 that is then being downgraded to OpenAPI 3.0 - but other generators don't seem to have a problem with this.  
The problem is that when the openapi code has `default` specified before something like a `200` response, the generated code creates the `default` in the Swift switch case before the other options which is not valid in the Swift language. I'm not sure if having default first is correct or not in the OpenAPI spec, but it does not seem to be a problem with other generators, spec validators, and tools like apimatic.io - its also automatically generated this way so manually maintaining a different version is less than ideal.

In this case, the generated OpenAPI spec has the following:

```yaml
paths:
  /api/info:
    get:
      summary: version
      description: server version information
      operationId: version
      parameters: []
      responses:
        default:
          description: ''
          headers: {}
          content:
            application/json:
              schema: {}
        '200':
          description: plain text
          headers: {}
          content:
            text/plain; charset=utf-8:
              schema: {}
        '401':
          description: not authorized if secured
          headers: {}
          content: {}
      deprecated: false
```

The result when the swift open-api-generator runs is:
`Additional 'case' blocks cannot appear after the 'default' block of a 'switch'`

as the generated code looks like this:
```swift
    /// version
    ///
    /// server version information
    ///
    /// - Remark: HTTP `GET /api/info`.
    /// - Remark: Generated from `#/paths//api/info/get(version)`.
    public func version(_ input: Operations.version.Input) async throws -> Operations.version.Output
    {
        try await client.send(
            input: input,
            forOperation: Operations.version.id,
            serializer: { input in
                let path = try converter.renderedRequestPath(template: "/api/info", parameters: [])
                var request: OpenAPIRuntime.Request = .init(path: path, method: .get)
                suppressMutabilityWarning(&request)
                try converter.setHeaderFieldAsText(
                    in: &request.headerFields,
                    name: "accept",
                    value: "application/json, text/plain"
                )
                return request
            },
            deserializer: { response in
                switch response.statusCode {
                default:
                    let headers: Operations.version.Output.Default.Headers = .init()
                    let contentType = converter.extractContentTypeIfPresent(
                        in: response.headerFields
                    )
                    let body: Operations.version.Output.Default.Body
                    if try contentType == nil
                        || converter.isMatchingContentType(
                            received: contentType,
[...]

PR #212, opened 2023-08-19
Additional 'case' blocks cannot appear after the 'default' block of a 'switch'

dpasirst:
### Motivation

BUG: https://github.com/apple/swift-openapi-generator/issues/211
This results in invalid Swift code that is not possible to compile.

### Modifications

As described by @czechboy0 in the thread linked above, a new computed property `responseOutcomes` on `OperationDescription` was added such that if a `default` is present, it is moved to the end. Other than `default`, the order is preserved.

### Result

The generated code produced is correct swift code allowing it to build.

### Test Plan

I did a manual test using the openapi.yaml that resulted in the original problem that I highlighted in the issue linked above.


czechboy0 (approved):
Thanks, @dpasirst!

czechboy0:

czechboy0:

PR #213, opened 2023-08-19
All descriptions defined in the OpenAPI document are generated

### LINKED ISSUE #206: Not all descriptions defined in the OpenAPI document are generated
For example, like below image, user description is not reflected in comment parameter.

スクリーンショット 2023-08-17 9 50 38

※ I was thinking of address this in https://github.com/apple/swift-openapi-generator/pull/196 , so that fix may also be helpful.

takeshi-1000:
### Motivation

Resolve #206 

### Modifications

Passed description defined in OpenAPI document to some `docCommentWithUserDescription`

### Result

Parameter descriptions such as query, request header, response header are now generated in Types.swift

### Test Plan

- Modified Types.swift in Petstore so that test passed
- Added description for openapi.yaml in test code to ensure this fix


czechboy0 (approved):
Thanks, @takeshi-1000!

PR #219, opened 2023-08-25
Support OpenAPI 3.1 in addition to 3.0

### LINKED ISSUE #39: Support OpenAPI 3.1
Today, we assume OpenAPI 3.0.x, but we should explore how we can also support 3.1, and continue to support both in parallel for some time, as 3.0.x is still the most adopted one out in the wild.

czechboy0:
### Motivation

Fixes #39. We need to start accepting 3.1 documents in addition to 3.0.

It also has useful features, like being able to add descriptions to references.

### Modifications

Migrated the codebase to use the `OpenAPIKit` instead of the `OpenAPIKit30` module, and updated the parser to handle both 3.0 and 3.1, and convert 3.0 to 3.1 documents in memory transparently.

### Result

OpenAPI 3.1 documents are now accepted, instead of rejected. Any 3.1 specific features can be addressed separately, this is the MVP support.

### Test Plan

Updated the file-based reference test to use version 3.1.0, and the parser tests that 3.0.x documents are loaded and converted successfully. So no need to duplicate our reference tests or anything.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
+++ b/Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
@@ -12,7 +12,9 @@
 //
 //===----------------------------------------------------------------------===//
 import Foundation
+import OpenAPIKit
 import OpenAPIKit30
+import OpenAPIKitCompat
 import Yams
 
 /// A parser that uses the Yams library to parse the provided
@@ -44,21 +46,27 @@ struct YamsParser: ParserProtocol {
         guard let openAPIVersion = versionedDocument.openapi else {
             throw Diagnostic.openAPIMissingVersionError(location: .init(filePath: input.absolutePath.path))
         }
-        switch openAPIVersion {
-        case "3.0.0", "3.0.1", "3.0.2", "3.0.3":
-            break
-        default:
-            throw Diagnostic.openAPIVersionError(
-                versionString: "openapi: \(openAPIVersion)",
-                location: .init(filePath: input.absolutePath.path)
-            )
-        }
-
         do {
-            return try decoder.decode(
-                OpenAPI.Document.self,
-                from: input.contents
-            )
+            let document: OpenAPIKit.OpenAPI.Document
+            switch openAPIVersion {
+            case "3.0.0", "3.0.1", "3.0.2", "3.0.3":
+                let openAPI30Document = try decoder.decode(
+                    OpenAPIKit30.OpenAPI.Document.self,
+                    from: input.contents
+                )
+                document = openAPI30Document.convert(to: .v3_1_0)
+            case "3.1.0":
+                document = try decoder.decode(
+                    OpenAPIKit.OpenAPI.Document.self,
+                    from: input.contents
+                )
+            default:
+                throw Diagnostic.openAPIVersionError(
+                    versionString: "openapi: \(openAPIVersion)",
+                    location: .init(filePath: input.absolutePath.path)
+                )
+            }
+            return document
         } catch DecodingError.dataCorrupted(let errorContext) {
             try checkParsingError(context: errorContext, input: input)
             throw DecodingError.dataCorrupted(errorContext)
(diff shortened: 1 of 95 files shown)


Tests/OpenAPIGeneratorReferenceTests/Resources/Docs/petstore.yaml
```diff
@@ -1,4 +1,4 @@
-openapi: "3.0.3"
```
Kyle-Ye, line 1: Since we are supporing both 3.0.x and and 3.1, maybe add a new petstore_3_1.yaml file and a new test case while keeping the original 3.0.3 test case is better IMO.

Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
```diff
@@ -104,7 +112,7 @@ extension Diagnostic {
     static func openAPIVersionError(versionString: String, location: Location) -> Diagnostic {
         return error(
             message:
-                "Unsupported document version: \(versionString). Please provide a document with OpenAPI versions in the 3.0.x set.",
+                "Unsupported document version: \(versionString). Please provide a document with OpenAPI versions in the 3.0.x or 3.1.x sets.",
```
Kyle-Ye, line 115: When the user accidentally uses the non-existent version 3.0.4, the error message here may be confusing.

Kyle-Ye (approved):

Tests/OpenAPIGeneratorReferenceTests/Resources/Docs/petstore.yaml
czechboy0: We already test in the parser tests that we can load 3.0.3 (and other 3.0.x) documents. I don't think we want to maintain two almost identical integration tests, which are expensive to update.

Sources/_OpenAPIGeneratorCore/Parser/YamsParser.swift
czechboy0: Possibly, but the only alternative would be to spell out every single supported version in the error message, which I'm not sure is needed?

czechboy0: Also, if the user uses 3.0.4, every single other OpenAPI tool will fail too. So I'm not too worried about people providing malformed documents.

Sources/swift-openapi-generator/Documentation.docc/Articles/Supported-OpenAPI-features.md
```diff
@@ -4,9 +4,9 @@ Learn which OpenAPI features are supported by Swift OpenAPI Generator.
 
 ## Overview
 
-Swift OpenAPI Generator is currently focused on supporting [OpenAPI 3.0.3][0].
+Swift OpenAPI Generator is currently focused on supporting [OpenAPI 3.0.3][0] and [OpenAPI 3.1.0][1].
```
glbrntt, line 7: The yams parser suggests `3.0.{0,1,2}` are also supported. Should they be listed here as well? 

More generally how good are OpenAPI document vendors at switching to newer versions?

Sources/swift-openapi-generator/Documentation.docc/Articles/Supported-OpenAPI-features.md
```diff
@@ -155,7 +155,7 @@ Supported features are always provided on _both_ client and server.
 - [x] description
 - [x] format
 - [ ] default
-- [ ] nullable
+- [ ] nullable (only in 3.0, removed in 3.1)
```
glbrntt, line 158: Does this mean the types generated from a 3.0.x document prior to this change will be different to that of the same document after this change as it's converted to 3.1.x first?

glbrntt (approved):

Sources/swift-openapi-generator/Documentation.docc/Articles/Supported-OpenAPI-features.md
czechboy0: It's meant to mean what we're focused on, i.e. if someone wants something from 3.0.2 that isn't relevant in 3.0.3, it'd likely be out of scope. But it doesn't mean we _only_ support these versions (i.e. that we'd reject them in the generator).

See the updates to README/landing page, where I added a table of exact supported versions.

PR #225, opened 2023-08-28
Allow non-unique values in oneOf.discriminator.mapping

### LINKED ISSUE #223: In oneOf.discriminator.mapping, allow non-unique values (only enforce unique keys)
In oneOf.discriminator.mapping, allow non-unique values (only enforce unique keys).

Today, we get `In discriminator.mapping, found multiple keys for the same value` when value types repeat, but that's wrong - we should only enforce unique keys.

czechboy0:
### Motivation

Fixes #223, see the issue for details.

### Modifications

Changed the logic for deriving the enum cases to take both the mapping and the list of subschemas into account. Documented it more in code.

### Result

Now adopter can actually have different keys pointing to the same types in their mapping - originally reported by an adopter, so it's a real issue.

### Test Plan

Adapted tests.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateAllAnyOneOf.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateAllAnyOneOf.swift
@@ -124,75 +124,71 @@ extension FileTranslator {
         discriminator: OpenAPI.Discriminator?,
         schemas: [JSONSchema]
     ) throws -> Declaration {
-        // > When using the discriminator, inline schemas will not be considered.
-        // > — https://spec.openapis.org/oas/v3.0.3#discriminator-object
-        let includedSchemas: [JSONSchema]
-        if discriminator != nil {
-            includedSchemas = schemas.filter(\.isReference)
-        } else {
-            includedSchemas = schemas
-        }
-
-        // The list of cases isn't dependent only on the list of subschemas, but
-        // also on the optional discriminator.mapping property, which can
-        // actually map from multiple keys to the same value.
-        // At the same time, the mapping doesn't have to mention all the
-        // subschemas, in which case the default behavior (use the JSON
-        // path of the schema) is used.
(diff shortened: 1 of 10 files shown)


Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateAllAnyOneOf.swift
```diff
@@ -124,55 +124,71 @@ extension FileTranslator {
...
-                if schema.isReference {
-                    // Use the type name.
-                    caseName = childType.typeName.shortSwiftName
-                } else {
-                    // Use a position-based key.
-                    caseName = key
+        let cases: [(String, [String]?, Comment?, TypeUsage, [Declaration])]
+        if let discriminator {
+            // > When using the discriminator, inline schemas will not be considered.
+            // > — https://spec.openapis.org/oas/v3.0.3#discriminator-object
```
glbrntt, line 130: Might be worth referencing the 3.1.x docs now

glbrntt (approved):

Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateAllAnyOneOf.swift
czechboy0: Yeah I considered it, at the same time we wrote the code with 3.0.3 in mind, and it'll be a learning exercise for all of us maintainers to get more fully up to speed with 3.1 and start updating the code. Basically - since I don't know exactly if there aren't some nice additions in 3.1 in all of these areas, which might mean the implementation could be better, I didn't want to just update the links as it'd make it seem like we _read_ the 3.1 specification when implementing it, but that's not the case.

Once the bigger items are done, I'm planning to more fully diff the 3.0 and 3.1 specifications and write up how we can _actually_ start taking advantage of 3.1, whereas now we only just implemented being able to ingest 3.1, without yet fully taking advantage of it.

Does that make sense? 🙂

PR #235, opened 2023-08-30
Stop generating empty input/output structs (query, headers, ...)

### LINKED ISSUE #229: Consider not generating empty input/output structs (e.g. paths, query, etc)
Consider not generating empty input/output structs (e.g. paths, query, etc).

Today, we generate them all, even when they're empty.

To try to see impact of not doing so on real-world docs, I prototyped a version of the generator that skips those empty structs on my fork ([diff](https://github.com/czechboy0/swift-openapi-generator/compare/hd-prep-0.2.0...czechboy0:swift-openapi-generator:hd-skip-generating-empty-input-output-structs?expand=1)).

The result: it removes about 10% of the code of GitHub's OpenAPI document, about 50k LOC.

Saving lines of code is a nice benefit, but I actually find that the code is more readable as well, and code completion guides you better (as you can see which containers have some parameters, whereas today you can't).

Opening this issue to start the discussion, happy to take this through a proposal if there are varying views.

czechboy0:
### Motivation

Fixes #229.

### Modifications

Stops generating the structs and properties when they're empty.

### Result

About 10% reduction in generated code, more clear generated code.

### Test Plan

Adapted tests, check those out first to see how the generated code is changing.


glbrntt (approved):
Nice!

PR #236, opened 2023-08-31
Include parameters in the content type name

### LINKED ISSUE #232: Include any parameters in content type unique names
The k8s API has the following:

```json
        "responses": {
          "200": {
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/io.k8s.api.core.v1.ComponentStatusList"
                }
              },
              "application/json;stream=watch": {
                "schema": {
                  "$ref": "#/components/schemas/io.k8s.api.core.v1.ComponentStatusList"
                }
              },
```

The generated code doesn't compile, because we generate duplicate enum cases, as the `;stream=watch` is stripped.

It should instead be included and hashed into the unique name.

Since this is technically breaking, we should do it as part of 0.2.0.

czechboy0:
### Motivation

Fixes #232, which was blocking the Kubernetes OpenAPI doc from successfully generating and building.

### Modifications

Include parameters, such as `application/json; foo=bar` in the content type name, e.g. `application_json_foo_bar`.

### Result

#232 is fixed, verified already locally.

### Test Plan

Updated unit and snippet tests.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/Content/ContentType.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Content/ContentType.swift
@@ -108,15 +108,15 @@ struct ContentType: Hashable {
         originallyCasedSubtype.lowercased()
     }
 
-    /// The parameter section, if any.
+    /// The parameter key-value pairs.
     ///
     /// Preserves the casing from the input, do not use this
-    /// for equality comparisons, use `lowercasedParameters` instead.
+    /// for equality comparisons, use `lowercasedParameterPairs` instead.
     let originallyCasedParameterPairs: [String]
 
-    /// The component after ';' of the MIME type, as a lowercase string.
+    /// The parameter key-value pairs, lowercased.
     ///
-    /// The raw value in its original casing is only provided by `originallyCasedParameters`.
+    /// The raw value in its original casing is only provided by `originallyCasedParameterPairs`.
     var lowercasedParameterPairs: [String] {
         originallyCasedParameterPairs.map { $0.lowercased() }
     }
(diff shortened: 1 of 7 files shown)


Sources/_OpenAPIGeneratorCore/Translator/Content/ContentType.swift
```diff
@@ -143,21 +182,25 @@ struct ContentType: Hashable {
         "\(lowercasedType)/\(lowercasedSubtype)"
     }
 
-    /// Returns the type and subtype as a "<type>\/<subtype>" string.
+    /// Returns the type and subtype as a "<type>\/<subtype>[;<param>...]" string.
```
gjcairo, line 185: Small nit: maybe worth updating to something like this?
```suggestion
    /// Returns the type, subtype and parameters (if present) as a "<type>\/<subtype>[;<param>...]" string.
```

gjcairo (approved):

Tests/OpenAPIGeneratorCoreTests/Translator/Content/Test_ContentSwiftName.swift
```diff
@@ -39,6 +39,10 @@ final class Test_ContentSwiftName: Test_Core {
             // Generic names.
             ("application/myformat+json", "application_myformat_plus_json"),
             ("foo/bar", "foo_bar"),
+
+            // Names with a parameter.
+            ("application/foo", "application_foo"),
+            ("application/foo; bar=baz; boo=foo", "application_foo_bar_baz_boo_foo"),
```
glbrntt, line 45: Is it worth testing that additional whitespace is handled too? e.g. `"application/foo; bar = baz"`?

glbrntt (approved):

Tests/OpenAPIGeneratorCoreTests/Translator/Content/Test_ContentSwiftName.swift
czechboy0: Sure, let me add that.

czechboy0: Done in f4d95b7

PR #248, opened 2023-09-06
Add support for nullable schemas

### LINKED ISSUE #82: Support nullable values

czechboy0:
### Motivation

Fixes #82.

More background:

In JSON Schema, there are two fields that affect what we represent as optionality in Swift.
1. In object schemas, not including a property in the `required` array.
2. In all schemas, marking the schema as nullable. In OpenAPI 3.0, that used to be done using a dedicated `nullable` field, in OpenAPI 3.1, that field is removed and instead the same result is achieved by adding the `null` type in the `type` array. So for a non-nullable string, you'd use `type: string`, and for a nullable string, you'd use `type: [string, null]`.

Up until now, we've only supported (1) in how we decide which types are generated as optional, and that seems to be how most adopters control nullability. However, (2) is also supported in JSON Schema, so we should also support it, and as evidenced by the comments in #82, a few folks have hit the lack of nullability support in Swift OpenAPI Generator.

### Modifications

This PR takes nullability on the schema into account when deciding whether to generate a type as optional or not.

Since this is a source-breaking change, it's introduced behind the feature flag `nullableSchemas`, and it'll become default behavior in 0.3.0.

Most of this change is just propagating the feature flag to the right place.

### Result

Marking a schema as nullable will result in an optional type being generated.

### Test Plan

Added a snippet test, both with and without the feature flag, which shows the change.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeMatcher.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeMatcher.swift
@@ -20,6 +20,10 @@ struct TypeMatcher {
     /// safe to be used as a Swift identifier.
     var asSwiftSafeName: (String) -> String
 
+    /// A Boolean value indicating whether the `nullable` field on schemas
+    /// should be taken into account.
+    var supportNullableSchemas: Bool
+
     /// Returns the type name of a built-in type that matches the specified
     /// schema.
     ///
@@ -78,8 +82,11 @@ struct TypeMatcher {
                 guard case let .reference(ref, _) = schema else {
                     return nil
                 }
-                return try TypeAssigner(asSwiftSafeName: asSwiftSafeName)
-                    .typeName(for: ref).asUsage
+                return try TypeAssigner(
+                    asSwiftSafeName: asSwiftSafeName,
+                    supportNullableSchemas: supportNullableSchemas
+                )
+                .typeName(for: ref).asUsage
             },
             matchedArrayHandler: { elementType in
                 elementType.asArray
@@ -88,7 +95,7 @@ struct TypeMatcher {
                 TypeName.arrayContainer.asUsage
             }
         )?
-        .withOptional(!schema.required)
+        .withOptional(!schema.required || (supportNullableSchemas && schema.nullable))
     }
 
     /// Returns a Boolean value that indicates whether the schema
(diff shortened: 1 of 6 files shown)


Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeMatcher.swift
```diff
@@ -88,7 +95,7 @@ struct TypeMatcher {
                 TypeName.arrayContainer.asUsage
             }
         )?
-        .withOptional(!schema.required)
+        .withOptional(!schema.required || (supportNullableSchemas && schema.nullable))
```
czechboy0, line 98: This is the whole behavior change.

simonjbeaumont (approved):

Kyle-Ye:
If I understand correctly, (2) has not been solved after this PR. So maybe #82 should be reopened instead of closed.

czechboy0:
@Kyle-Ye please confirm that you're including the feature flag, the tests in this PR passed so at least some scenarios are fixed.

Kyle-Ye:
> @Kyle-Ye please confirm that you're including the feature flag, the tests in this PR passed so at least some scenarios are fixed.

Sorry, I wrongly thought it was already enabled by default in the current release.(0.2.1)

Following the [doc](https://swiftpackageindex.com/apple/swift-openapi-generator/0.2.1/documentation/swift-openapi-generator/configuring-the-generator) here, I got an error after adding the feature flag.

image

I fixed it by using a main branch dependency after reading the source code that 0.2.1 release does not include the nullableSchemas feature flag.

The nullable issue is indeed solved in my scenario. Waiting for the 0.3.0 release 🎉

thunderball1:
Does anyone know why it disappeared from main? There is only empty in enums for feature flags

czechboy0:
All feature flags got enabled by default in 0.3.0.

PR #255, opened 2023-09-08
[Proposal] SOAR-0005: Adopting the Swift HTTP Types package

czechboy0:
### Motivation

See the proposal at https://github.com/czechboy0/swift-openapi-generator/blob/hd-soar-0005/Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0005.md

### Modifications

N/A

### Result

N/A

### Test Plan

N/A


gjcairo:
This proposal is now In Review, which will run until 15th September - feel free to leave your feedback either on this PR or on the corresponding [Swift Forums post](https://forums.swift.org/t/proposal-soar-0005-adopting-the-swift-http-types-package/67185).

Thanks, and thanks @czechboy0 for the write-up!

gjcairo:
Overall this looks good to me, and I believe it's the right direction for the ecosystem to move in 👍🏼 

Just one question though: wouldn't it make sense to also make the `HTTPBody`s in `ClientTransport` and `ClientMiddleware` also optional?

czechboy0:
> Just one question though: wouldn't it make sense to also make the `HTTPBody`s in `ClientTransport` and `ClientMiddleware` also optional?

Great question, I actually tried to make that change, unfortunately unlike with request bodies (which OpenAPI documents as optional, and semantically, a nil body is distinct from an empty body), response bodies are required, and if there's no documented response body, it's considered empty. There's no concept of a nil response body.

The reasoning seems to come from HTTP semantics, where GET and DELETE requests are not supposed to have request bodies, unlike other methods, where a request body is allowed. In contrast, response bodies are never discouraged by the HTTP spec, so there's no nullability to communicate.

Consider the client transport implementing this and trying to make a decision of whether to return an empty or nil body - how do you make that? You can't, as there's no way to know whether the server sent a "nil body" or an empty body. That, again, is in contrast to a server transport, which can reliably, for GET and DELETE requests, provide a nil body (no need to even instantiate an empty `HTTPBody` value) without looking at the wire, as even if there are bytes, servers are supposed to ignore them for these methods.

gjcairo:
The review period is now over, and the proposal has been accepted. SOAR-0005 is now **Ready for Implementation**.

czechboy0:
Landing the proposals to avoid more conflicts, will update their status once the implementation lands.

PR #266, opened 2023-09-13
Bump OpenAPIKit to beta.5

czechboy0:
### Motivation

OpenAPIKit is getting ready to tag a 3.0.0 release candidate, so we need to keep on the latest version and ensure it works for us.

### Modifications

Bumped from beta.3 to beta.5, contains some new features like `const` support.

Adapted to the slight changes in behavior.

### Result

Now using the latest version.

### Test Plan

All tests pass again.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
+++ b/Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
@@ -12,6 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+import OpenAPIKit
+
 /// Runs validation steps on the incoming OpenAPI document.
 /// - Parameters:
 ///   - doc: The OpenAPI document to validate.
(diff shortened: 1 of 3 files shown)


Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
```diff
@@ -27,7 +29,10 @@ func validateDoc(_ doc: ParsedOpenAPIRepresentation, config: Config) throws -> [
     // block the generator from running.
     // Validation errors continue to be fatal, such as
     // structural issues, like non-unique operationIds, etc.
-    let warnings = try doc.validate(strict: false)
+    let warnings = try doc.validate(
+        using: Validator().validating(.operationsContainResponses),
```
simonjbeaumont, line 33: Looks like we're opting into more than the default now? Is that now _more_ strict than before?

Given we're emitting warnings for all of these, are there any more we can be enabling?

simonjbeaumont (approved):

Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
czechboy0: In beta.4, the `.operationsContainResponses` validation was _removed_ from the default set, so I just added it back to maintain the same behavior. I'm not aware of any validations that we're _not_ running but could be, from the ones built-in to OpenAPIKit. Of course, OpenAPIKit has a nice API for building additional ones, and I'm very open to introducing more validations if we find patterns difficult to detect later in the generator that we could be catching early.

PR #274, opened 2023-09-14
Escape negative numbers in integer-backed enums

### LINKED ISSUE #265: Integer backed enums can't handle negative numbers
Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateStringEnum.swift needs a fix, we should prefix negative numbers with e.g. "n", so "1" has the case "_1" and "-1" would have "_n1" or just "n1".

Right now if a negative number is used, the generated code won't compile.

PARAIPAN9:
### Motivation

Fixes #265

### Modifications

When handling a negative number, it takes its absolute value and add 'n' before it

### Result

Example: "-1" -> "n1"

### Test Plan

Ran and passed all the tests

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateStringEnum.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateStringEnum.swift
@@ -61,7 +61,7 @@ extension FileTranslator {
                     guard let rawValue = anyValue as? Int else {
                         throw GenericError(message: "Disallowed value for an integer enum '\(typeName)': \(anyValue)")
                     }
-                    let caseName = "_\(rawValue)"
+                    let caseName = rawValue < 0 ? "n\(abs(rawValue))" : "_\(rawValue)"
                     return (caseName, .int(rawValue))
                 }
             }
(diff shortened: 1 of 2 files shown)


Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateStringEnum.swift
```diff
@@ -61,7 +61,7 @@ extension FileTranslator {
                     guard let rawValue = anyValue as? Int else {
                         throw GenericError(message: "Disallowed value for an integer enum '\(typeName)': \(anyValue)")
                     }
-                    let caseName = "_\(rawValue)"
+                    let caseName = rawValue < 0 ? "n\(abs(rawValue))" : "_\(rawValue)"
```
czechboy0, line 64: Thoughts on whether we should also make positive numbers spelled "p1" for consistency or keep the current "_1", @glbrntt @simonjbeaumont @gjcairo?

czechboy0:
Please add a test case for this as well, just to be positive (ha) that it works.

Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateStringEnum.swift
czechboy0: That said, it'd be a breaking change, and I don't find `p1` significantly better than `_1`, so I'm also fine with keeping it as-is, and only fixing negative numbers, which today lead to non-compiling code, so no breaking change there.

czechboy0: Ok, let's go with what @simonjbeaumont suggested. @PARAIPAN9 please update the PR with that and add a few unit tests and we should be set! 🙏

PARAIPAN9: Perfect, thank you for the clarifications!!

czechboy0 (approved):
Great, thanks!

PR #280, opened 2023-09-15
Generate all requested modes in parallel

### LINKED ISSUE #227: Generate multiple files in parallel
Usually, adopters ask to generate more than 1 file, typically `types` and `client`, or `types` and `server`.

While internally, generating each file is pretty single-threaded, generating each file is parallelizable, so we should do that, and potentially cut generation time in 1/2 for large documents.

simonjbeaumont:
### Motivation

The generator itself is synchronous, but it can be run in one of more modes: `types`, `client`, `server`, and these runs can be parallelised. This is likely to have value to most adopters since most will use at least two modes. This is especially useful for large APIs.

### Modifications

Run all the requested modes of the generator in a task group.

### Result

When running with multiple modes, generation is faster.

Concretely, when using the Github API and generating types and client, it cut the overall generation time by 40%.

### Test Plan

CI.

### Resolves

Resolves #227.

### DIFF
--- a/Sources/swift-openapi-generator/runGenerator.swift
+++ b/Sources/swift-openapi-generator/runGenerator.swift
@@ -11,7 +11,14 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 //===----------------------------------------------------------------------===//
-import Foundation
+#if os(Linux)
+@preconcurrency import struct Foundation.URL
+@preconcurrency import struct Foundation.Data
+#else
+import struct Foundation.URL
+import struct Foundation.Data
+#endif
+import class Foundation.FileManager
 import ArgumentParser
 import _OpenAPIGeneratorCore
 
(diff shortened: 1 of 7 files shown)


Sources/swift-openapi-generator/runGenerator.swift
```diff
@@ -33,23 +33,29 @@ extension _Tool {
...
+            for config in configs {
+                group.addTask {
+                    try runGenerator(
+                        doc: doc,
+                        docData: docData,
+                        config: config,
+                        outputDirectory: outputDirectory,
+                        outputFileName: config.mode.outputFileName,
+                        isDryRun: isDryRun,
+                        diagnostics: diagnostics
```
czechboy0, line 54: Can you make the YAML diagnostics collector thread-safe please? I think that's the only one that could encounter problems with this change.

czechboy0 (changes requested):

Sources/swift-openapi-generator/runGenerator.swift
czechboy0: The stderr printing one might also need a way to serialize the printing to ensure we don't get garbled output. Possibly the whole diagnostics collector protocol should become Sendable.

simonjbeaumont: I've rebased on main following merging #281 and made it all clean of warnings.

czechboy0 (approved):

PR #283, opened 2023-09-19
Add url form encoder & decoder

bfrearson:
### Motivation

#182 Currently, the generated interface for request bodies that use URLEncodedForms accepts a `Data` object. This PR uses the URIEncoder/Decoder to instead provide an interface using codable types in the same way that a JSON body would.

### Modifications
Adds a new coding strategy that works with the corresponding [runtime converter updates](https://github.com/apple/swift-openapi-runtime/pull/53).

**Note:** This PR does not support encoding collections of primitives (i.e. an array of string values for a single key). This should be addressed in a future update, but it raised questions about the correct use of URIEncoder, so I didn't include it in this PR. Further discussion of the issue is in the [runtime PR](https://github.com/apple/swift-openapi-runtime/pull/53).

### Result
Generated interfaces now directly accept URLEncodedForm Codable types rather than untyped `Data`.

### Test Plan

Updated reference tests to include a request using a form body.

### DIFF
--- a/Tests/OpenAPIGeneratorReferenceTests/Resources/Docs/petstore.yaml
+++ b/Tests/OpenAPIGeneratorReferenceTests/Resources/Docs/petstore.yaml
@@ -182,7 +182,6 @@ paths:
   /probe/:
     post:
       operationId: probe
-      deprecated: true
       responses:
         '204':
           description: Ack
@@ -270,13 +269,6 @@ components:
       schema:
         type: integer
         format: int64
-    header.deprecatedHeader:
-      name: deprecatedHeader
-      in: header
-      deprecated: true
-      description: A deprecated header parameter
-      schema:
-        type: string
   schemas:
     Pet:
       type: object
@@ -318,7 +310,6 @@ components:
     PetKind:
       type: string
       description: "Kind of pet"
-      deprecated: true
       enum:
         - cat
         - dog
@@ -462,17 +453,6 @@ components:
         - $ref: '#/components/schemas/MessagedExercise'
       discriminator:
         propertyName: kind
-    DeprecatedObject:
-      deprecated: true
-      type: object
-      properties: {}
-      additionalProperties: false
-    ObjectWithDeprecatedProperty:
-      type: object
-      properties:
-        message:
-          type: string
-          deprecated: true
     PetStats:
       type: object
       properties:
--- a/Tests/OpenAPIGeneratorReferenceTests/Resources/Docs/petstore.yaml
+++ b/Tests/OpenAPIGeneratorReferenceTests/Resources/Docs/petstore.yaml
@@ -256,6 +256,24 @@ components:
           type: string
         kind:
           $ref: '#/components/schemas/PetKind'
+    MixedAnyOf:
+      anyOf:
+        - type: string
+          format: date-time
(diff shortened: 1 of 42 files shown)


Tests/PetstoreConsumerTests/TestClient.swift
```diff
@@ -82,7 +93,8 @@ struct TestClient: APIProtocol {
         return try await block(input)
     }
 
-    typealias UploadAvatarForPetSignature = @Sendable (Operations.uploadAvatarForPet.Input) async throws -> Operations
+    typealias UploadAvatarForPetSignature = @Sendable (Operations.uploadAvatarForPet.Input) async throws ->
+        Operations
         .uploadAvatarForPet.Output
     var uploadAvatarForPetBlock: UploadAvatarForPetSignature?
```
czechboy0, line 99: Please also add corresponding tests to Test_Client and Test_Server that ensures at runtime everything is correctly propagated.

Tests/OpenAPIGeneratorReferenceTests/Resources/Docs/petstore.yaml
```diff
@@ -113,6 +113,47 @@ paths:
                 $ref: '#/components/schemas/Pet'
         '4XX':
           $ref: '#/components/responses/ErrorBadRequest'
+  /pets/create:
```
czechboy0, line 116: Let's simplify this to only test what we want to test here - url encoded request bodies.
- remove X-Extra-Arguments from the request and response headers
- remove the 400 response
- make the success response be 204 and return no content at all

Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Client.swift
```diff
@@ -48,7 +48,8 @@ public struct Client: APIProtocol {
         try await client.send(
             input: input,
             forOperation: Operations.listPets.id,
-            serializer: { input in let path = try converter.renderedPath(template: "/pets", parameters: [])
+            serializer: { input in
+                let path = try converter.renderedPath(template: "/pets", parameters: [])
```
czechboy0, line 52: Could this be due to the swift-format version mismatch again? We'll upgrade at some point, but you'll likely need to revert this to pass CI :)

czechboy0 (changes requested):
Looks great, just a few suggestions, otherwise almost ready to go just like the runtime PR. 🙏

Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Client.swift
bfrearson: good spot thanks! Had already reverted to the correct version of swift-format, but looks like the updated version had already made these changes

czechboy0:

czechboy0 (approved):
I made a few more fixes, looks good now.

czechboy0:
@bfrearson most notably, I had to hide the feature under the `urlEncodedForm` feature flag, as changing the generated type is an API-breaking change, so must be opt-in in 0.2.x, and will become the default in 0.3.0.

PR #289, opened 2023-09-21
Move warning about strict concurrency into PlatformChecks.swift

simonjbeaumont:
### Motivation

In #191 we enabled strict concurrency in CI. We also attempted to include it conditionally as part of a local development flow. However, the `#warning` (that strict concurrency was enabled) added to Package.swift was firing unconditionally because it was, itself, behind a _runtime_ conditional.

### Modifications

Now if `SWIFT_OPENAPI_STRICT_CONCURRENCY=true` then we will add a `define` to the `swiftSettings` and we will move the `#warning` to the existing `PlatformChecks.swift`.

### Result

Opening Xcode with `SWIFT_OPENAPI_STRICT_CONCURRENCY=true` will enable strict concurrency warnings locally. Opening Xcode without this will no longer produce a misleading warning.

### Test Plan

Tested locally.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -27,10 +27,9 @@ swiftSettings.append(
 
 // Strict concurrency is enabled in CI; use this environment variable to enable it locally.
 if ProcessInfo.processInfo.environment["SWIFT_OPENAPI_STRICT_CONCURRENCY"].flatMap(Bool.init) ?? false {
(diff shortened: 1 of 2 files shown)


czechboy0 (approved):
:shipit:

Issue #145, opened 2023-07-26
Wrap operation methods in specialized methods (1 success response, 1 content type)


MahdiBM:
Here's an example of some code i'm repeating again and again in [Penny](https://github.com/vapor/penny-bot):
```swift 
func getFilesChanges() async throws -> [FileDiff] {
    let response = try await githubClient.pulls_list_files(.init(
        path: .init(...),
        body: .init(...)
    ))

    guard case let .ok(ok) = response,
            case let .json(json) = ok.body else {
        throw Errors.httpRequestFailed(response: response)
    }

    return json
}
```
I think this function should roughly look like this. optimally:
```swift
func getFilesChanges() async throws -> [FileDiff] {
    try await githubClient.pulls_list_files(
        path1: ...,
        path2: ...,
        body: .init(...)
    ).decode()
}
```
So let's analyze how the second function is better:
- It's a single step so easier to use.
- Cleaner, so nicer to read. Doesn't need weird `case let` or nested `switch` statements.
- It is still what 99% of users want anyway.
- It also `decode()` the response only after requested by user.
  - I prefer this as sometimes you don't need the response at all, but it's lower on the priority list.
- The `decode()` function itself is supposed to throw a proper error if the response isn't a success response.
  - There can be another `decodeError()` function or the like to not check for success of the request considering one might be in fact trying to decode the error response. 

What i mentioned above is basically what i've tried and implemented in [DiscordBM](https://github.com/DiscordBM/DiscordBM), and i've really liked, which is achieved by:
- A base [HTTP-response type](https://github.com/DiscordBM/DiscordBM/blob/2ed51291f8c562b4ec09934edc006a76c2180482/Sources/DiscordHTTP/HTTP%20Types.swift#L8)
- Another [HTTP-response type that wraps the base one](https://github.com/DiscordBM/DiscordBM/blob/2ed51291f8c562b4ec09934edc006a76c2180482/Sources/DiscordHTTP/HTTP%20Types.swift#L44).
  - Only thing extra it has is that it takes a generic parameter.
    - So it can automatically `decode()` the response to that type when needed.
  - This is purely for ease of usage by users. So they don't need to specify the decode type or make one.
[...]

Issue #104, opened 2023-07-07
Improve ergonomics when switching over Outputs


clarkkent-superman:
Getting down to the actual response model object requires drilling down through few layers of structs and enums:

```swift
switch result {
case let .ok(reponse):
    case let .json(payload):
        print(payload.body, payload.headers)
    @unknown default:
        break
    }
@unknown default:
    break
} 
```

I understand there is a requirement to handle multiple response types. If the payload could be flattened into just a body and headers, instead of a struct containing both of these, it would be possible to use the following syntax which would still support multiple response types.

```swift
switch result {
case let .ok(.json(body, headers):
  print(body, headers)
@unknown default:
    break
} 
```

simonjbeaumont:
@clarkkent-superman thanks for filing this issue.

As you noted, the API of the generated code is designed to be as expressive as the OpenAPI specification allows. This includes multiple response bodies.

The [spec for the response object](https://swagger.io/specification/#response-object) specifies a response have headers and then a content map for various media types—i.e. the headers are shared across all possible content types for that response. 

You're correct: it does result in a verbose API for simple operations that have a single content type. However, our goal was to generate code that would evolve in a sensible way as the API evolves.

For example, when an operation that has a single content type grows the ability to provide an additional body content type. The experience today is that clients will have another enum case to handle, which feels like an idiomatic mapping to Swift. 

While we want to retain the expressive API layer, we are definitely interested in adding some conveniences/sugar APIs on top if/when they make sense, and we'd definitely welcome a contribution in this area. However, we'd need to have a well-thought-out design first. Issue #22 has some ideas, which might overlap with this issue.

Could you elaborate on the flattening you have in mind? Would we then need to duplicate the headers object into each of the response body type enum cases?

//cc @czechboy0

czechboy0:
Tagged as needing design – but feel free to add ideas in this issue until someone puts together a concrete proposal.

PadraigK:
❤️  Nice!

Issue #22, opened 2023-05-25
Generate convenience initializers that flatten some nested scopes


czechboy0:
Creating input types is quite verbose because they contain multiple nested enums. (e.g. `.ok(.init(headers: .init(), body: .json(...)))`). 

It may be useful to generate static helper functions on the output type so it's possible to write `.ok(body: .json(...))` instead, or even `.ok(...)` directly.

We have to be careful here, though, as it's possible to go very far and generate a ton of conveniences, that end up making the API more difficult to understand. So will require a well thought-out proposal first, describing which high-value conveniences we should generate, and how we evaluate further requests.

We also need to keep in mind that the generator complexity shouldn't be significantly increased with this work, as per https://github.com/apple/swift-openapi-generator/blob/main/Sources/swift-openapi-generator/Documentation.docc/Articles/Project-scope-and-goals.md#principle-reduce-complexity-of-the-generator-implementation

czechboy0:
For unwrapping, let's see if this could work: `case let .ok(.json(response)):`.

simonjbeaumont:
@czechboy0 this would work in a `guard`/`else`. Is that what you hand in mind? 

I think we still need some heuristics for what to name these and when to generate them.

I'm still not sure how much value that work will add when adopters can write the same in their code.

I think conveniences on for creating inputs and outputs will probably add more value. WDYT?

czechboy0:
> @czechboy0 this would work in a `guard`/`else`. Is that what you hand in mind?

Yeah, in a guard, if, or a switch. Just wanted to note here that once can unwrap more than one level in one go.

> I think we still need some heuristics for what to name these and when to generate them.

Agreed. Right now I'm just dumping thoughts in this issue, I don't yet have a good idea of how we approach improving these.

czechboy0:
Ah nevermind, that won't work as the response associate value us a struct, not the body enum itself.

PR #291, opened 2023-09-22
Add SOAR-0007 proposal: Shorthand APIs for inputs and outputs

simonjbeaumont:
### Motivation

We'd like to run a proposal for generating additional API to simplify providing operation inputs and handling
operation outputs for simple API calls with just one successful outcome.

### Modifications

- Add SOAR-0007: Shorthand APIs for operation inputs and outputs

(See the proposal itself for details)[^1]

### Result

n/a

### Test Plan

n/a

### Related Issues

#22, #104, #105, #145

[^1]: https://github.com/apple/swift-openapi-generator/pull/291/files


gjcairo:
This proposal is now **In Review**, which will run until September 29 - feel free to leave your feedback either on this PR or on the corresponding [Swift Forums post](https://forums.swift.org/t/proposal-soar-0007-shorthand-apis-for-inputs-and-outputs/67444).

Thanks @simonjbeaumont for the write-up!

yanniks:
Also from my side, thank you for this proposal! I think it will really simplify many use-cases where the client component of the generator is used.
However, I'm wondering whether the thrown errors can be improved somehow. The challenge I see is that one might want to abstract error handling from the actual API call and catch `Unauthorized` or `Forbidden` responses in a certain way. With the current proposal, I don't see that there is a generic way of identifying whether the `UnexpectedResponseError` was thrown because of a `404`, `403` or a `400` HTTP status code response. Could this be incorporated in this draft somehow?

MahdiBM:
Another related issue: https://github.com/apple/swift-openapi-generator/issues/145 🙂

simonjbeaumont:
> Another related issue: #145 🙂

Thanks. I'd inconsistently referenced the issues between the proposal and the PR description 🤦 #145 was already referenced in the proposal itself, at least 🙂 

- Proposal: 22, 104, 145
- PR description: 22, 104, 105

I'll clean both up to be the union.

simonjbeaumont:
> @simonjbeaumont did not really think of using a middleware. In that case, I really support this proposal. Thanks for detailed reply!

No problem!

simonjbeaumont:

czechboy0 (approved):

PR #295, opened 2023-09-25
YAML diagnostics collector should have stable output

### LINKED ISSUE #294: YAML diagnostics collector output does not seem to be stable
Possibly we need to add an explicit sort of diagnostics before writing it out.

czechboy0:
### Motivation

Fixes #294.

Since we're running generation in parallel now, the order of diagnostics depends on the racing of the cores, resulting in unstable output.

### Modifications

Sort the diagnostics before writing them to file, making the output stable again.

### Result

Stable YAML diagnostics file.

### Test Plan

N/A, the order isn't significant, so we're not testing the details of the sorting function, we just care about it being stable.


dnadoba (approved):
LGTM

PR #296, opened 2023-09-25
Enable all feature flags for 0.3.0

czechboy0:
### Motivation

Wait for SOAR-0004/5 to land first.

_[Explain here the context, and why you're making that change. What is the problem you're trying to solve.]_

### Modifications

_[Describe the modifications you've made.]_

### Result

_[After your change, what will change.]_

### Test Plan

_[Describe the steps you took, or will take, to qualify the change - such as adjusting tests and manual testing.]_


czechboy0:
Integrate test will keep failing until we tag generator 0.3.0, so this is expected.

simonjbeaumont (approved):

PR #297, opened 2023-09-25
Fixes #276 - Unknown configuration options should throw an error

### LINKED ISSUE #276: Unknown configuration options should throw an error
Example, if you use `featureFlag` instead of `featureFlags`:

```yaml
generate:
  - types
featureFlag:
  - nullableSchemas
```

This is easy to do because the CLI is `--feature-flag` (taking multiple).
 
We already print the following to stdout:

```none
- OpenAPI document path: ...
- Configuration path: ...
- Feature flags: ...
- Output directory: ...
- Current directory: ...
- Plugin source: <none>
- Is dry run: false
- Additional imports: <none>
```

Probably should print something here about unknown keys?

PARAIPAN9:
### Motivation

Fixes #276

### Modifications

Create a function that extracts the keys from the config file and compares them with the UserConfig Coding Keys

### Result

When the config file loads, it also checks for the keys to match

### Test Plan

Loaded different config files with different configurations to see if it works.

### DIFF
--- a/Sources/swift-openapi-generator/GenerateOptions.swift
+++ b/Sources/swift-openapi-generator/GenerateOptions.swift
@@ -89,28 +89,6 @@ extension _GenerateOptions {
         return config?.featureFlags ?? []
     }
 
-    /// Extracts the top-level keys from a YAML string.
-    ///
-    /// - Parameter yamlString: The YAML string from which to extract keys.
-    /// - Returns: An array of top-level keys as strings.
-    func extractTopLevelKeys(fromYAMLString yamlString: String) -> [String] {
-        var yamlKeys = [String]()
-
-        do {
-            let parser = try Parser(yaml: yamlString)
-            if let rootNode = try parser.singleRoot(),
-                case let .mapping(mapping) = rootNode
-            {
-                for (key, _) in mapping {
-                    yamlKeys.append(key.string ?? "")
-                }
-            }
-        } catch {
-            print("Error parsing YAML: \(error)")
-        }
-        return yamlKeys
-    }
-
     /// Validates a collection of keys against a predefined set of allowed keys.
     ///
     /// - Parameter keys: A collection of keys to be validated.
@@ -134,10 +112,14 @@ extension _GenerateOptions {
         }
         do {
             let data = try Data(contentsOf: config)
-
             let configAsString = String(decoding: data, as: UTF8.self)
+            var yamlKeys = [String]()
 
-            let yamlKeys = extractTopLevelKeys(fromYAMLString: configAsString)
+            do {
+                yamlKeys = try YamsParser.extractTopLevelKeys(fromYAMLString: configAsString)
+            } catch {
+                throw ValidationError("The config isn't valid. \(error)")
+            }
             try validateKeys(yamlKeys)
 
             let config = try YAMLDecoder().decode(_UserConfig.self, from: data)
--- a/Sources/swift-openapi-generator/GenerateOptions.swift
+++ b/Sources/swift-openapi-generator/GenerateOptions.swift
@@ -89,22 +89,39 @@ extension _GenerateOptions {
         return config?.featureFlags ?? []
     }
 
-    /// Extracts keys from a YAML string.
+    /// Extracts the top-level keys from a YAML string.
     ///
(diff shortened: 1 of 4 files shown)


Sources/swift-openapi-generator/GenerateOptions.swift
```diff
@@ -99,6 +117,16 @@ extension _GenerateOptions {
         }
         do {
             let data = try Data(contentsOf: config)
+            let configAsString = String(data: data, encoding: .utf8)
+
+            let yamlKeys = extractYamlKeys(yamlString: configAsString!)
```
czechboy0, line 122:
```suggestion
            let configAsString = String(decoding: data, as: UTF8.self)
            let yamlKeys = extractYamlKeys(yamlString: configAsString)
```

Sources/swift-openapi-generator/UserConfig.swift
```diff
@@ -29,4 +29,12 @@ struct _UserConfig: Codable {
 
     /// A set of features to explicitly enable.
     var featureFlags: FeatureFlags?
+
+    static let codingKeysRawValues = Set(CodingKeys.allCases.map({ $0.rawValue }))
```
czechboy0, line 33: Please add a comment for the symbol.

Sources/swift-openapi-generator/GenerateOptions.swift
```diff
@@ -86,7 +86,25 @@ extension _GenerateOptions {
...
+
+        var extractedKeys = [String]()
+
+        for line in lines {
+            if let range = line.range(of: ":") {
+                let key = line[line.startIndex..<range.lowerBound].trimmingCharacters(in: .whitespaces)
+                extractedKeys.append(key)
+            }
+        }
+        return extractedKeys
```
czechboy0, line 107: I agree we need to extract the keys somehow, but I don't think we should try to write our own YAML parser here. Can you see if you can use the lower level Yams API to do the same? https://www.jpsim.com/Yams/Classes/Parser.html

czechboy0 (changes requested):

Sources/swift-openapi-generator/GenerateOptions.swift
PARAIPAN9: @czechboy0, thank you for the comments. I wasn't sure about the YAML parsing method, which is why I opened the PR without writing tests for it. I will check the link provided by you to explore other possibilities.

Sources/swift-openapi-generator/GenerateOptions.swift
```diff
@@ -86,7 +86,42 @@ extension _GenerateOptions {
...
+            let parser = try Parser(yaml: yamlString)
+            if let rootNode = try parser.singleRoot(),
+                case let .mapping(mapping) = rootNode
+            {
+                for (key, _) in mapping {
+                    yamlKeys.append(key.string ?? "")
+                }
+            }
+        } catch {
+            print("Error parsing YAML: \(error)")
```
czechboy0, line 109: I think it's better to just let the error be thrown one level up, and there emit a ValidationError saying the config isn't valid and include the description of the error thrown by Yams. So you can remove the do/catch here.

czechboy0 (changes requested):
Just one more nit, and can you add a unit test please?

Sources/swift-openapi-generator/GenerateOptions.swift
PARAIPAN9: Sure thing!!

PARAIPAN9: Just a quick question, where would be a good place in the codebase, to write the unit test?

czechboy0: You could move the logic close to the YAML parser in the core module, and that one has unit tests that you can add to. We don't yet have a unit test target for the CLI module.

PARAIPAN9: perfect, TY!!

PARAIPAN9:
Please feel free to comment, suggest changes if still something is not right or could be improved. I really appreciate your input!!

czechboy0 (approved):
Thanks!

PR #300, opened 2023-09-27
Propagate nullability through references

### LINKED ISSUE #288: Propagate nullability through references
At the moment, when a referenced schema is nullable, the callers of that references don't correctly propagate that up and it's always deserialized as if it's required, leading to decoding failures.

Since we can't define a type itself as a nullable (e.g. how do you define a nullable struct?), we'd either have to go through an extra typealias everywhere (option 1) or have the callers of the reference peek into it to know (option 2).

Exploring option 2 for now, but it's possible we still do option 1 if it ends up being too messy.

czechboy0:
### Motivation

Fixes #288.

### Modifications

Refactors how we decide whether to mark a type as optional, and adds the ability to "peek" through references to decide the nullability of the reference schema.

Also added an article describing the overall logic as a starting point, will be expanded with a few more changes coming in this area in subsequent PRs.

### Result

We now correctly mark an object property as optional if it points at an optional schema through a reference. Previously, it was always marked as required, which caused decoding to fail e.g. in the GitHub OpenAPI doc.

### Test Plan

Added and adjusted unit tests.


Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
simonjbeaumont, line 1: Can we add a test here for this?

czechboy0: Added: https://github.com/apple/swift-openapi-generator/pull/300/commits/a5f268116849ccb7f5f2edc642f1f65bb178651f

simonjbeaumont (approved):

Issue #285, opened 2023-09-19
Generate API protocols based on tags


olejnjak:
Currently OpenAPI generator generates a single `APIProtocol` for all client calls.

It would be a good improvement if it would be possible to generate a protocol based on tags in the spec. So given the following spec, two protocols would be generated `PetsAPIProtocol` & `StoreAPIProtocol`.

```
paths:
  /pet/findByStatus:
    get:
      summary: Finds pets by Status
      tags:
        - pets
      ...
  /pet:
    post:
      summary: Adds a new pet to the store
      tags:
        - pets
      ...
  /store/inventory:
    get:
      summary: Returns pet inventories
      tags:
        - store
      ...
```

Based on https://x.com/czechboy0/status/1703869135682195838?s=20

czechboy0:
I think that's a good idea. Would help with large documents, and even when implementing servers, it could help break up the API handler a bit.

This should be doable even in a backwards compatible way, as we could continue to generate 

`protocol APIProtocol: PetsAPIProtocol, StoreAPIProtocol {}` so those who still want to just use a single object for everything could continue to do so.

Still, this should go through a proposal, but should be considered before 1.0 in case we choose to do it in an API-breaking way.

If anyone wants to take it on and go through the proposal and implementation process, that'd be great. We can help guide you through the process.

czechboy0:
Just to add to Si, you could then generate all the operations for one tag into one module, making the fully qualified protocol name `ModuleA.APIProtocol`, other tag into another module (`ModuleB.APIProtocol`), etc, allowing splitting a large doc into several smaller parts.

simonjbeaumont:
@olejnjak Just a heads up that SOAR-0008 proposal is up in #303, which might go some way to meet your requirements.

PR #303, opened 2023-09-28
Add proposal SOAR-0008: OpenAPI document filtering

simonjbeaumont:
### Motivation

We'd like to run a proposal for filtering the OpenAPI document for just the required parts prior to generating.

### Modifications

- Add SOAR-0008: OpenAPI document filtering

(See the proposal itself for details)[^1]

### Result

n/a

### Test Plan

n/a

### Related Issues

#285

[^1]: https://github.com/apple/swift-openapi-generator/pull/TODO/files

### DIFF
--- a/Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0008.md
+++ b/Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0008.md
@@ -1,0 +1,462 @@
+# SOAR-0008: OpenAPI document filtering
+
+Filtering the OpenAPI document for just the required parts prior to generating.
+
+## Overview
+
+- Proposal: SOAR-0008
+- Author(s): [Si Beaumont](https://github.com/simonjbeaumont)
+- Status: **Awaiting Review**
+    - Review period: YYYY-MM-DD – YYYY-MM-DD
+    - [Swift Forums post](TODO)
+- Issue:
+    - [apple/swift-openapi-generator#TODO](https://github.com/apple/swift-openapi-generator/issues/TODO)
+- Implementation:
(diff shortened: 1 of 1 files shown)


czechboy0:
Thanks, @simonjbeaumont!

This proposal is now In Review, which will run until **October 5** - feel free to leave your feedback either on this PR or on the corresponding [Swift Forums post](https://forums.swift.org/t/proposal-soar-0008-openapi-document-filtering/67574).

porglezomp:
Should it be possible to exclude tags as well as include them?
I'm thinking of cases like an API where some endpoints are tagged as higher privileged, and a low-privilege client might want to exclude those endpoints.

czechboy0:
I quite like the "!" prefix, and I agree this should all compose nicely, so folks are encouraged to propose further extensions of this feature.

porglezomp:
Another possible option that doesn't add nesting but keeps out-of-band signaling would be:
```yaml
filter:
  tags:
    - "t"
    - not: "rw"
```
Although I think `!` is safe because tags are required to be valid URI characters.

simonjbeaumont:
@porglezomp On reflection, I think that we want to include filtering at the operation level (cf. path) from the start. I have updated the proposal to bring that part forward from the "Future directions" section into the proposed changes and updated the PoC branch with an implementation.

At this time I haven't added support for excludes or complement filters.

As this feedback was incorporated pretty late in the review period, I'm proposing we extend it for a few days.

Please let me know your thoughts on the updates.

czechboy0:
Thanks @simonjbeaumont, looks great - okay, let's extend the review period into next week.

czechboy0 (approved):

czechboy0 (changes requested):
Actually wait - please add it to the Proposals.md page so that it's linkable.

czechboy0 (approved):

PR #308, opened 2023-10-02
Generate shorthand throwing APIs for providing inputs and handling outputs

### LINKED ISSUE #22: Generate convenience initializers that flatten some nested scopes

### LINKED ISSUE #104: Improve ergonomics when switching over Outputs

### LINKED ISSUE #145: Wrap operation methods in specialized methods (1 success response, 1 content type)

simonjbeaumont:
### Motivation

The review period for SOAR-0007 (Shorthand APIs for inputs and outputs) has now concluded. This pull request adds the required SPIs to the runtime library to throw a runtime error when the response and/or body does not match that of the shorthand API being used.

For further context, please review the proposal itself.[^1] 

[^1]: https://github.com/apple/swift-openapi-generator/pull/291

### Modifications

- Add support for computed properties with effects (e.g. throwing getters).
- Generate a protocol extension to `APIProtocol` with an overload for each operation that lifts each of the parameters of `Input.init` as function parameters.
- Generate a throwing computed property for each enum case related to a documented outcome, which will return the associated value for the expected case, or throw a runtime error if the value is a different enum case.

### Result

Code that used to be written like this 

```swift
// before
switch try await client.getGreeting(.init()) {
case .ok(let response):
    switch response.body {
    case .json(let body):
        print(body.message)
    }
case .undocumented(statusCode: _, _):
    throw UnexpectedResponseError()
}

// after
print(try await client.getGreeting().ok.body.json.message)
//                     ^          ^  ^       ^
//                     |          |  |       `- (New) Throws if body did not conform to documented JSON.
//                     |          |  |
//                     |          |  `- (New) Throws if HTTP response is not 200 (OK).
//                     |          |
//                     |          `- (New) No need to wrap parameters in input value.
//                     |
//                     `- (Existing) Throws if there is an error making the API call.
```

### Test Plan

This PR includes updates to the various tests:
- `SnippetBasedReferenceTests`
- `FileBasedReferenceTests`
- `PetstoreConsumerTests`

### Related Issues

- Resolves #22 
- Resolves #104 
- Resolves #145


Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
```diff
@@ -103,6 +103,61 @@ extension TypesFileTranslator {
...
                 bodyCases.append(bodyCase)
+
+                var throwingGetterSwitchCases = [
+                    SwitchCaseDescription(
+                        kind: .case(.identifier(".\(identifier)"), ["body"]),
+                        body: [.expression(.return(.identifier("body")))]
+                    )
+                ]
+                // We only generate the default branch if there is more than one case to prevent
+                // a warning when compiling the generated code.
```
czechboy0, line 114: That's a good idea.

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
```diff
@@ -103,6 +103,61 @@ extension TypesFileTranslator {
...
+                        SwitchCaseDescription(
+                            kind: .default,
+                            body: [
+                                .expression(
+                                    .try(
+                                        .identifier("throwUnexpectedResponseBody")
+                                            .call([
+                                                .init(
+                                                    label: "expectedContent",
+                                                    expression: .literal(.string(contentType.headerValueForSending))
```
czechboy0, line 126:
```suggestion
                                                    expression: .literal(.string(contentType.headerValueForValidation))
```

While this is just for printing, `headerValueForValidation` is more appropriate as it doesn't include `charset`, which isn't used to match content types.

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
```diff
@@ -103,6 +103,61 @@ extension TypesFileTranslator {
...
+                            )
+                        )
+                    ],
+                    getterEffects: [.throws]
+                )
+                let throwingGetterComment = Comment.doc(
+                    """
+                    The associated value of the enum case if `self` is `.\(identifier)`.
+
+                    - Throws: Runtime error if `self` is not `.\(identifier)`.
```
czechboy0, line 156:
```suggestion
                    - Throws: An error if `self` is not `.\(identifier)`.
```

Since "runtime error" isn't a public type, I wouldn't mention it here.

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponseOutcome.swift
```diff
@@ -73,7 +67,61 @@ extension TypesFileTranslator {
...
+            type: responseStructTypeName.fullyQualifiedSwiftName,
+            getter: [
+                .expression(
+                    .switch(
+                        switchedExpression: .identifier("self"),
+                        cases: [
+                            SwitchCaseDescription(
+                                kind: .case(
+                                    .identifier(".\(responseKind.identifier)"),
+                                    responseKind.wantsStatusCode ? ["_", "response"] : ["response"]
```
czechboy0, line 84: I thought about this a bit, and I think you made the right choice. For ranges, the alternative choice would have been to return `(Int, BODY_TYPE)`, but I prefer just returning `BODY_TYPE`, and if the adopter wants the code, they need to switch manually.

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponseOutcome.swift
```diff
@@ -73,7 +67,61 @@ extension TypesFileTranslator {
...
+                    )
+                )
+            ],
+            getterEffects: [.throws]
+        )
+        let throwingGetterComment = Comment.doc(
+            """
+            The associated value of the enum case if `self` is `.\(enumCaseName)`.
+
+            - Throws: Runtime error if `self` is not `.\(enumCaseName)`.
```
czechboy0, line 115:
```suggestion
            - Throws: An error if `self` is not `.\(enumCaseName)`.
```

Same as above.

czechboy0 (approved):
A few suggestions, but no blockers so I'll let you make the final call on these.

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
simonjbeaumont: OK, happy to change. FTR I didn't mean `RuntimeError`, I meant "a runtime error" (in the generic sense), but I guess the capital R led to confusion. Will update everywhere (manually, because it affects the reference code).

Resolved in 4c423d8.

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponseOutcome.swift
simonjbeaumont: Resolved in 4c423d8.

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
simonjbeaumont: After some further discussion, the right thing to do is probably use the `headerValueForValidation` here, because that's actually what was used to deserialize the body.

However, the fact that `headerValueForValidation` here does _not_ include the additional parameter, means we've probably uncovered an existing bug with how we handle responses when there are overlapping content types. We should deserialize as the most specific one, but we're probably at the mercy of the ordering of the deserialization logic, which, in turn, is probably derived from the order in which the contents appear in the OpenAPI document.

simonjbeaumont: Updated in 0905ce0.

simonjbeaumont:

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
czechboy0: Thanks. Please file an issue when you have a moment for improving the matching when content types overlap.

simonjbeaumont:

simonjbeaumont:

simonjbeaumont:

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponse.swift
simonjbeaumont: Done: https://github.com/apple/swift-openapi-generator/issues/315

PR #311, opened 2023-10-03
Split client and server examples into separate packages

simonjbeaumont:
### Motivation

The client and server examples currently live in a single Swift package. The majority of adopters will probably want to crib from an example of a client _or_ a server, and having them together runs the risk of them cargo-culting dependencies they don't need. The current setup also prevents us from updating the dependencies in the client example if the server transport releases are lagging.

### Modifications

- Split the client and server examples into separate Swift packages.
- Update the client example to the latest runtime and main generator.
- Drop dependencies in server example to 0.2.0 family for now.

### Result

Client example package is updated to latest dependencies and can build.

### Test Plan

Tested that both packages can build locally.


czechboy0 (approved):
That's a good idea, let's do it.

PR #313, opened 2023-10-03
Skip type: string + format: binary in object properties

### LINKED ISSUE #312: Skip type: string + format: binary in object properties
Until we support multipart (#36), we need to skip `type: string, format: binary` properties, as they get generated inside `Codable` structs and fail to compile.

Might require the `isSchemaSupported` family of functions to gain some "context" enum that tells the function where the schema is being used, and while we need to allow this schema as a root schema of `application/octet-stream` (and other binary) content types, we should disallow it (by skipping it) as object properties.

Once we support multipart, we'll likely generate non-Codable structs to hold the schemas used by multipart, but that's TBD and out of scope of this issue.

czechboy0:
### Motivation

Fixes #312.

### Modifications

Explicitly skip binary properties in codable structs.

### Result

Remove build failures for projects using multipart.

### Test Plan

Created a unit test.


simonjbeaumont (approved):
LGTM. Nice snippet test too!

czechboy0:

yim-lee:

PR #319, opened 2023-10-03
Add support for filtering OpenAPI document

simonjbeaumont:
### Motivation

When generating client code, Swift OpenAPI Generator generates code for the entire OpenAPI document, even if the user only makes use of a subset of its types and operations.

Generating code that is unused constitutes overhead for the adopter:
- The overhead of generating code for unused types and operations
- The overhead of compiling the generated code
- The overhead of unused code in the users codebase (AOT generation)

This is particularly noticeable when working with a small subset of a large API, which can result in O(100k) lines of unused code and long generation and compile times.

For a more detailed motivation and design, see the proposal in #303.

### Modifications

- Add document filter to the generator config.
- Run filter as a post-transition hook in the generator pipeline after parsing the document.
- Provide a CLI command that outputs the filtered document to stdout.

### Result

Users can filter a document before code-generation. For large APIs, this can result in >90% speedup (see proposal).

### Test Plan

- Unit tests.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
+++ b/Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
@@ -15,8 +15,6 @@
 @preconcurrency import OpenAPIKit
 
 /// Rules used to filter an OpenAPI document.
-///
-/// - Todo: Add endpoint support
 public struct DocumentFilter: Codable, Sendable {
 
     /// Operations with these operation IDs will be included in the filter.
--- a/Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
+++ b/Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
@@ -19,6 +19,9 @@
 /// - Todo: Add endpoint support
 public struct DocumentFilter: Codable, Sendable {
 
+    /// Operations with these operation IDs will be included in the filter.
+    public var operations: [String]?
+
     /// Operations tagged with these tags will be included in the filter.
     public var tags: [String]?
 
@@ -34,15 +37,17 @@ public struct DocumentFilter: Codable, Sendable {
     /// Create a new DocumentFilter.
     ///
     /// - Parameters:
-    ///   - operations: Paths that contain operations with IDs will be included in the filter.
+    ///   - operations: Operations with these IDs will be included in the filter.
     ///   - tags: Operations tagged with these tags will be included in the filter.
     ///   - paths: These paths will be included in the filter.
     ///   - schemas: These (additional) schemas will be included in the filter.
     public init(
+        operations: [String] = [],
         tags: [String] = [],
         paths: [OpenAPI.Path] = [],
         schemas: [String] = []
     ) {
+        self.operations = operations
         self.tags = tags
         self.paths = paths
         self.schemas = schemas
@@ -56,6 +61,9 @@ public struct DocumentFilter: Codable, Sendable {
         var builder = FilteredDocumentBuilder(document: document)
         for tag in tags ?? [] {
             try builder.requireOperations(tagged: tag)
+        }
+        for operationID in operations ?? [] {
+            try builder.requireOperation(operationID: operationID)
         }
         for path in paths ?? [] {
             try builder.requirePath(path)
--- a/Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
+++ b/Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
@@ -19,7 +19,7 @@
 /// - Todo: Add endpoint support
 public struct DocumentFilter: Codable, Sendable {
 
-    /// Paths that contain operations with these tags will be included in the filter.
+    /// Operations tagged with these tags will be included in the filter.
     public var tags: [String]?
 
     /// These paths will be included in the filter.
@@ -35,7 +35,7 @@ public struct DocumentFilter: Codable, Sendable {
     ///
     /// - Parameters:
     ///   - operations: Paths that contain operations with IDs will be included in the filter.
-    ///   - tags: Paths that contain operations with these tags will be included in the filter.
+    ///   - tags: Operations tagged with these tags will be included in the filter.
     ///   - paths: These paths will be included in the filter.
     ///   - schemas: These (additional) schemas will be included in the filter.
     public init(
@@ -55,7 +55,7 @@ public struct DocumentFilter: Codable, Sendable {
     public func filter(_ document: OpenAPI.Document) throws -> OpenAPI.Document {
         var builder = FilteredDocumentBuilder(document: document)
         for tag in tags ?? [] {
-            try builder.requirePaths(taggedWith: tag)
+            try builder.requireOperations(tagged: tag)
         }
         for path in paths ?? [] {
             try builder.requirePath(path)
--- a/Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
+++ b/Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
@@ -55,6 +55,8 @@ public struct DocumentFilter: Codable, Sendable {
     ///
     /// - Parameter document: The OpenAPI document to filter.
     /// - Returns: The filtered document.
+    /// - Throws: If any requested document components do not exist in the original document.
+    /// - Throws: If any dependencies of the requested document components cannot be resolved.
     public func filter(_ document: OpenAPI.Document) throws -> OpenAPI.Document {
         var builder = FilteredDocumentBuilder(document: document)
         for tag in tags ?? [] {
--- a/Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
+++ b/Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
@@ -1,0 +1,68 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftOpenAPIGenerator open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftOpenAPIGenerator project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+@preconcurrency import OpenAPIKit
+
+/// Rules used to filter an OpenAPI document.
+///
+/// - Todo: Add endpoint support
+public struct DocumentFilter: Codable, Sendable {
+
+    /// Paths that contain operations with these tags will be included in the filter.
+    public var tags: [String]?
+
+    /// These paths will be included in the filter.
+    public var paths: [OpenAPI.Path]?
+
+    /// These (additional) schemas will be included in the filter.
+    ///
+    /// These schemas are included in  addition to the transitive closure of schema dependencies of
+    /// the paths included in the filter.
+    public var schemas: [String]?
+
+    /// Create a new DocumentFilter.
+    ///
+    /// - Parameters:
+    ///   - operations: Paths that contain operations with IDs will be included in the filter.
+    ///   - tags: Paths that contain operations with these tags will be included in the filter.
+    ///   - paths: These paths will be included in the filter.
+    ///   - schemas: These (additional) schemas will be included in the filter.
+    public init(
+        tags: [String] = [],
+        paths: [OpenAPI.Path] = [],
+        schemas: [String] = []
+    ) {
+        self.tags = tags
+        self.paths = paths
+        self.schemas = schemas
+    }
+
+    /// Filter an OpenAPI document.
+    ///
+    /// - Parameter document: The OpenAPI document to filter.
+    /// - Returns: The filtered document.
+    public func filter(_ document: OpenAPI.Document) throws -> OpenAPI.Document {
+        var builder = FilteredDocumentBuilder(document: document)
+        for tag in tags ?? [] {
+            try builder.requirePaths(taggedWith: tag)
+        }
+        for path in paths ?? [] {
+            try builder.requirePath(path)
+        }
+        for schema in schemas ?? [] {
+            try builder.requireSchema(schema)
+        }
+        return try builder.filter()
+    }
+}
--- a/Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
+++ b/Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
@@ -60,16 +60,16 @@ public struct DocumentFilter: Codable, Sendable {
     public func filter(_ document: OpenAPI.Document) throws -> OpenAPI.Document {
         var builder = FilteredDocumentBuilder(document: document)
         for tag in tags ?? [] {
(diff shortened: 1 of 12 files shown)


Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
```diff
@@ -0,0 +1,76 @@
...
+public struct DocumentFilter: Codable, Sendable {
+
+    /// Operations with these operation IDs will be included in the filter.
+    public var operations: [String]?
+
+    /// Operations tagged with these tags will be included in the filter.
+    public var tags: [String]?
+
+    /// These paths will be included in the filter.
+    public var paths: [OpenAPI.Path]?
```
czechboy0, line 29: Is there extra validation in the decoder of OpenAPI.Path we want to take advantage of? Or should we just make this a String?

Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
```diff
@@ -0,0 +1,76 @@
...
+
+    /// Create a new DocumentFilter.
+    ///
+    /// - Parameters:
+    ///   - operations: Operations with these IDs will be included in the filter.
+    ///   - tags: Operations tagged with these tags will be included in the filter.
+    ///   - paths: These paths will be included in the filter.
+    ///   - schemas: These (additional) schemas will be included in the filter.
+    public init(
+        operations: [String] = [],
```
czechboy0, line 45: Is there a distinction between an empty array and a nil array? Naively, I'd expect a nil array to mean "don't filter using this axis, aka keep everything", versus an empty array, which I'd understand as "filter out everything".

Sources/_OpenAPIGeneratorCore/Hooks/FilteredDocument.swift
```diff
@@ -0,0 +1,485 @@
...
+        return filteredDocument
+    }
+
+    /// Include a path (and all its component dependencies).
+    ///
+    /// The path is added to the filter, along with the transitive closure of all components
+    /// referenced within the corresponding path item.
+    ///
+    /// - Parameter path: The path to be included in the filter.
+    public mutating func requirePath(_ path: OpenAPI.Path) throws {
```
czechboy0, line 119: Super small nit, I find the word "include" a bit clearer than "require" here. Especially if we might add "exclude" down the line.

Sources/_OpenAPIGeneratorCore/Hooks/FilteredDocument.swift
```diff
@@ -0,0 +1,485 @@
...
+
+    /// Include the operation with a given ID (and all its component dependencies).
+    ///
+    /// This may result in paths within filtered document with a subset of the operations defined
+    /// in the original document.
+    ///
+    /// - Parameter operationID: The operation to include (and its path).
+    public mutating func requireOperation(operationID: String) throws {
+        guard document.allOperationIds.contains(operationID) else {
+            throw FilteredDocumentBuilderError.operationDoesNotExist(operationID: operationID)
```
czechboy0, line 158: Very good call, will ensure adopters don't make a typo in their operation id and then spend time debugging why their operation isn't showing up.

I wonder if we should somehow "consume" all the other filters as well, and throw an error if you're trying to include something that doesn't exist in the document? Not crucial, but would make debugging a lot easier, especially since this is all stringly programming from the config file, without any autocompletion or validation.

Sources/_OpenAPIGeneratorCore/Hooks/FilteredDocument.swift
```diff
@@ -0,0 +1,485 @@
...
+    mutating func addComponentsReferencedBy(_ content: OpenAPI.Example) throws {}
+}
+
+fileprivate extension OpenAPI.Reference {
+    var internalComponentKey: OpenAPI.ComponentKey {
+        get throws {
+            guard case .internal(.component(name: let name)) = jsonReference else {
+                throw FilteredDocumentBuilderError.cannotResolveInternalReference(absoluteString)
+            }
+            return OpenAPI.ComponentKey(rawValue: name)!
```
czechboy0, line 470: Can we throw an error if it fails to init? Looking at the impl, there might be situations where the user's document could cause this to crash, which becomes difficult to debug.

Sources/swift-openapi-generator/FilterCommand.swift
```diff
@@ -0,0 +1,100 @@
...
+//
+//===----------------------------------------------------------------------===//
+import ArgumentParser
+import Foundation
+import _OpenAPIGeneratorCore
+import Yams
+import OpenAPIKit
+import OpenAPIKit30
+
+struct _FilterCommand: AsyncParsableCommand {
```
czechboy0, line 21: This will be useful for debugging. Could you also add a few words about it to the Configuring docc article, both about the new keys in the config file, and about using the new filter command?

Tests/OpenAPIGeneratorCoreTests/Hooks/FilteredDocumentTests.swift
```diff
@@ -0,0 +1,185 @@
...
+        XCTAssertUnsortedEqual(
+            filteredDocument.components.schemas.keys.map(\.rawValue),
+            schemas,
+            file: file,
+            line: line
+        )
+    }
+}
+
+private func XCTAssertUnsortedEqual<T>(
```
czechboy0, line 171: Can you move this to `TestUtilities.swift`? Could be useful elsewhere, and we already have some XCTAssert* utils there.

czechboy0 (changes requested):
Looks great overall, just a few small suggestions.

Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
simonjbeaumont: The semantics, which probably need better documenting, if you didn't catch them are as follows:

- If you do not specify a `filter` key in your config file, then `config.filter: DocumentFilter?` will be `nil` and no filtering will take place.
- If you _do_ specify a `filter` key in your config file, then filtering will take place.
- When filtering takes place, we start with empty paths and schemas and only include what you asked for.

Concretely this means that `schemas: []` and `schemas: nil` are equivalent, and mean you don't get any schemas, apart from those that may be required by other things you asked for (paths or operations).

The optionality at this level was for decoding from the YAML file, without having to duplicate a type that differs only in the optionality of each property.

Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
simonjbeaumont: Yes, an important one.

This allows you to only specify strings that can be used as keys in the `paths` object in the OpenAPI document. Specifically, it allows you to specify paths that take parameters, e.g. `/books/{id}`, but won't allow arbitrary strings.

Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
```diff
@@ -0,0 +1,76 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+@preconcurrency import OpenAPIKit
+
+/// Rules used to filter an OpenAPI document.
+///
+/// - Todo: Add endpoint support
+public struct DocumentFilter: Codable, Sendable {
```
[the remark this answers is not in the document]
simonjbeaumont: While it's true that it's only used for config. It is also `Encodable` so we can use the same trick you used for the `generate` command help message. That is, we encode an example filter to YAML, when rendering the help message[^1].

[^1]: https://github.com/apple/swift-openapi-generator/pull/319/files#diff-32fe5d86e35489cd6f0071eb479054741c5fc75b3d88ec27fa73ed9bea7d359bR31

Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
```diff
@@ -0,0 +1,76 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+@preconcurrency import OpenAPIKit
+
+/// Rules used to filter an OpenAPI document.
+///
+/// - Todo: Add endpoint support
```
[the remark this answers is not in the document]
simonjbeaumont: Nope, thanks for catching; removed in f782f3a.

Sources/_OpenAPIGeneratorCore/Hooks/DocumentFilter.swift
czechboy0: Ah right, I forgot we do the union of all the various filtering types. Makes sense, thanks for the detailed explanation.

Sources/_OpenAPIGeneratorCore/Hooks/FilteredDocument.swift
```diff
@@ -0,0 +1,485 @@
...
+            if let schema = array.items {
+                try addComponentsReferencedBy(schema)
+            }
+
+        case .not(let schema, _):
+            try addComponentsReferencedBy(schema)
+
+        case .all(of: let schemas, _): fallthrough
+        case .one(of: let schemas, _): fallthrough
+        case .any(of: let schemas, _):
```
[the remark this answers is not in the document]
czechboy0: Yeah for diffs it's nicer, but I don't expect this to change, as these are fundamental JSONSchema types. I've just been trained to avoid fallthrough if I can.

Sources/swift-openapi-generator/FilterCommand.swift
```diff
@@ -0,0 +1,100 @@
...
+        case "3.0.0", "3.0.1", "3.0.2", "3.0.3":
+            let document30x = try decoder.decode(OpenAPIKit30.OpenAPI.Document.self, from: documentData)
+            document = document30x.convert(to: .v3_1_0)
+        case "3.1.0":
+            document = try decoder.decode(OpenAPIKit.OpenAPI.Document.self, from: documentData)
+        default:
+            fatalError("Unsupported OpenAPI version found: \(versionedDocument.openapi)")
+        }
+        return document
+    }
```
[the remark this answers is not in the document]
czechboy0: Yes, it's fully internal to our needs, so no need to avoid improving it.

Sources/_OpenAPIGeneratorCore/Hooks/FilteredDocument.swift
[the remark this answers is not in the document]
simonjbeaumont: Done in 997e4dd.

Sources/swift-openapi-generator/FilterCommand.swift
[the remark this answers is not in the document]
simonjbeaumont: Done in 3833ef9.

Sources/swift-openapi-generator/FilterCommand.swift
simonjbeaumont: Done in 9df55ed.

Sources/swift-openapi-generator/Documentation.docc/Articles/Configuring-the-generator.md
```diff
@@ -66,3 +72,40 @@ generate:
 additionalImports:
   - APITypes
 ```
+
+### Document filtering
+
+The geneartor supports filtering the OpenAPI document prior to generation, which can be useful when
```
czechboy0, line 78:
```suggestion
The generator supports filtering the OpenAPI document prior to generation, which can be useful when
```

Sources/swift-openapi-generator/Documentation.docc/Articles/Configuring-the-generator.md
```diff
@@ -66,3 +72,40 @@ generate:
...
+generating client code for a subset of a large API.
```
czechboy0, line 79:
```suggestion
generating client code for a subset of a large API, or splitting an implementation of a server across multiple modules.
```

Sources/swift-openapi-generator/Documentation.docc/Articles/Configuring-the-generator.md
```diff
@@ -66,3 +72,40 @@ generate:
...
+
+For example, to generate client code for only the operations with a given tag, use the following config:
```
czechboy0, line 81:
```suggestion
For example, to generate code for only the operations with a given tag, use the following config:
```

Sources/swift-openapi-generator/FilterCommand.swift
```diff
@@ -0,0 +1,82 @@
...
+
+    func run() async throws {
+        let configData = try Data(contentsOf: config)
+        let config = try YAMLDecoder().decode(_UserConfig.self, from: configData)
+        let documentInput = try InMemoryInputFile(absolutePath: docPath, contents: Data(contentsOf: docPath))
+        let document = try timing(
+            "Parsing document",
+            YamsParser.parseOpenAPIDocument(documentInput, diagnostics: StdErrPrintingDiagnosticCollector())
+        )
+        try document.validate()
```
czechboy0, line 50: I think you actually want to use the function `validateDoc` from `validateDoc.swift`, which also takes the diagnostic collector and uses the same logic as the generation command. By default the `validate()` call contains a slightly different set of validations, and is overly strict about schema warnings.

Sources/swift-openapi-generator/FilterCommand.swift
```diff
@@ -0,0 +1,82 @@
...
+        guard let documentFilter = config.filter else {
+            FileHandle.standardError.write("warning: No filter config provided\n")
+            FileHandle.standardOutput.write(try YAMLEncoder().encode(document))
```
czechboy0, line 53: Should we generate JSON if the user gave us JSON here?

czechboy0 (commented):
A few more comments, otherwise looks great.

Sources/swift-openapi-generator/FilterCommand.swift
simonjbeaumont: I hadn't considered that, good point. Should we detect this or just base it on the input path extension for now?

Sources/swift-openapi-generator/FilterCommand.swift
simonjbeaumont: We could remove this altogether, I suppose, but I'll update to the function that takes the `DiagnosticCollector`.

Sources/swift-openapi-generator/Documentation.docc/Articles/Configuring-the-generator.md
simonjbeaumont: I don't think this one is right, given the YAML below is explicitly for a concrete example that has `client`, similar to examples above in the doc.

czechboy0: Sure, let's keep the "client" word here then.

Sources/swift-openapi-generator/FilterCommand.swift
simonjbeaumont: I decided to remove the call to validate here instead.

I wrote the patch to make use of `validateDoc` but I needed to either remove the config parameter it accepts (and doesn't use) or provide a config from here (which relies on providing a generator mode). Neither of these seemed worth it.

I then realised that I don't want to do this for two reasons:
1. Folks might actually want to be using the filter to remove parts of a doc that has validation errors.
2. The implementation of validateDoc fully dereferences the doc and then loops over all components again, and I spent some time making sure the filter command was one-pass.

czechboy0 (approved):
:shipit:

PR #322, opened 2023-10-04
Fix a bug around optional request bodies

czechboy0:
### Motivation

As part of #288, we improved handling of nullability. Unfortunately that caused a regression for documents that use a somewhat strange combination of a required request body that has a nullable schema.

### Modifications

Ensure that when generating the server and we have to spell out which type is being unwrapped, that we use the non-optional type name.

This means that we respect the requestBody's `required` field, and ignore the `nullable` field of the root schema. To me, that seems like the most reasonable interpretation of the potential conflict.

### Result

Unblocks one large OpenAPI document.

### Test Plan

Added snippet tests.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/RequestBody/translateRequestBody.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/RequestBody/translateRequestBody.swift
@@ -328,7 +328,11 @@ extension ServerFileTranslator {
                 .call([
                     .init(
                         label: nil,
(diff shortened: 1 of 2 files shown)


czechboy0:
@swift-server-bot compat-suite-full

czechboy0:

simonjbeaumont (commented):
LGTM, pending compat test.

simonjbeaumont (approved):

PR #325, opened 2023-10-05
Update all examples and integration tests to 0.3.0

czechboy0:
### Motivation

We're ready to release 0.3.0, so update all examples and integration tests, so that once tagged, the code uses the latest versions.

### Modifications

Updated integration tests and examples, also made a few minor improvements.

### Result

All code refers to 0.3.0.

### Test Plan

Verified manually, but some of these might not pass because 0.3.0 of the generator doesn't exist yet - chicken and egg, but that's understood as part of this process.

### DIFF
--- a/Examples/GreetingService/Package.swift
+++ b/Examples/GreetingService/Package.swift
@@ -20,9 +20,9 @@ let package = Package(
         .macOS(.v13)
     ],
     dependencies: [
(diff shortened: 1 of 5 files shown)


simonjbeaumont (approved):
LGTM. It's a bit sad that `swift package edit` cannot continue to work to override the `swift-openapi-generator` dependency.

Issue #11, opened 2023-05-25
Support base64 encoded data (type: string + format: byte)


czechboy0:
Currently, only `type: string` + `format: binary` is supported, and represents raw data (an octet stream).

OpenAPI also supports base64-encoded data, represented as the following JSON schema:

```yaml
type: string
format: byte
```

The generated type for it should also be `Foundation.Data` (same as for `format: binary`), but as part of serialization/deserialization, the extra step of base64 coding needs to be added.

Testing: verify that both a whole request/response body being `application/json` with the contents of a base64-encoded string (use case 1), but also as a nested field in a JSON object (use case 2), both are pretty common.

czechboy0:
Note that since this isn't a content type, but a way to encode and decode any schema, including a nested one, this'll likely justify creating a type looking like this in the runtime library:

```
struct Base64EncodedData: Sendable, Codable, Hashable {
    var data: Foundation.Data // or [UInt8] or ArraySlice<UInt8>
    init(data: Foundation.Data) {
        self.data = data
    }
    init(from decoder: Decoder) throws {
        let container = try decoder.singleValueContainer()
        let base64EncodedString = try container.decode(String.self)
        guard let data = Data(base64Encoded: base64EncodedString) else {
            throw RuntimeError.invalidBase64String(base64EncodedString)
        }
        self.init(data: data)
    }
    func encode(to encoder: Encoder) throws {
        var container = encoder.singleValueContainer()
        let base64String = data.base64EncodedString()
        try container.encode(base64String)
    }
}
```

And then for e.g. the following JSON Schema snippet

```yaml
MyObj:
  type: object
  properties:
    stuff:
      type: string
      format: byte
  required: [stuff]
```

generate a struct like:

```swift
struct MyObj: Sendable, Codable, Hashable {
  var stuff: OpenAPIRuntime.Base64EncodedData
  // ...
}
```

czechboy0:
Shipped in https://github.com/apple/swift-openapi-generator/releases/tag/0.3.1.

PR #326, opened 2023-10-06
Support base64-encoded data

rnro:
This change accompanies https://github.com/apple/swift-openapi-runtime/pull/55 and relies on it for the `OpenAPIRuntime.Base64EncodedData` type.

### Motivation

OpenAPI supports base64-encoded data but to this point Swift OpenAPI Generator has not (https://github.com/apple/swift-openapi-generator/issues/11).

### Modifications

A data type specified as `type: string, format: byte` will now result in a generated type which is `Codable` and backed by a `OpenAPIRuntime.Base64EncodedData` type which knows how to encode and decode base64 data.

### Result

Users will be able to specify request/response payloads as base64-encoded data which will be encoded and decoded transparently

### Test Plan

Unit tested locally.


czechboy0:

czechboy0:
@rnro As a last thing, can you add a base64 property to an object of your choice in the file-based reference tests? I'm a little paranoid that this PR is _passing_ even though we haven't released the update runtime yet, which means we're not testing that it all builds and works together. Add a property, and make sure PetstoreConsumerTests are updated to also populate and verify the value both on client and server side. Otherwise lgtm.

rnro:
Thanks for the review, I've added an endpoint to upload audio of your pet in base64-encoded binary format. I'm not sure if that's the best API design for an actual pet app but I think it's plausible to get the new code in there.

czechboy0 (changes requested):
Please also add tests to Test_Client and Test_Server that set and verify the value in `Pet.genome`, just to make sure it works end-to-end.

czechboy0:
Also, you can now update to 0.3.2 of the runtime library in Package.swift of swift-openapi-generator, all that should hopefully make the CI green on this PR.

PR #330, opened 2023-10-13
Support recursive types

czechboy0:
### Motivation

Fixes https://github.com/apple/swift-openapi-generator/issues/70.

### Modifications

To start, read the new docc [article](https://github.com/apple/swift-openapi-generator/blob/ceb51fa1a4f1f858590b22da75162c4bf999719b/Sources/swift-openapi-generator/Documentation.docc/Development/Supporting-recursive-types.md) about how reference types are implemented, then most of the PR should make sense.

As suggested by @simonjbeaumont, boxing of recursive types happens on the Swift representation, as opposed to my original approach, which tried to do this early in the translation layer. This massively simplified the problem and definitely seems like the better way to do it.

Highlights:
- In `validateDoc`, removed the dereferencing, which we previously used to catch cycles early and emit descriptive errors.
- Introduced an efficient stack type caller `ReferenceStack` that makes checking if an item is present in the stack fast, on top of being a stack (represented as an array).
- Helper methods like `isSchemaSupported` and `isKeyValuePair` gained an inout parameter of the stack, to allow it to break infinite recursion.
- The actual algorithm for walking the graph, detecting cycles, and deciding which types to box is implemented in `RecursionDetector`, which is a generic algorithm on top of nodes with edges.
- Then `DeclarationRecursionDetector` provides concrete types that glue it with our structured Swift representation's `Declaration`.
- The algorithm runs in `translateSchemas` where we're iterating over the items in `#/components/schemas`, as those are the only ones that can produce a cycle (as schemas in other parts of the document can refer to items in `#/components/schemas`, but not the other way around: items in `#/components/schemas` cannot refer to schemas outside of it.)

### Result

OpenAPI documents with recursive schemas are now supported.

### Test Plan

- Added unit tests for the recursion detector.
- Adapted other tests, of `isSchemaSupported` and `isKeyValuePair`.
- Added examples to `petstore.yaml`, as this one introduces quite a lot of new code that we want to make sure compiles without warnings.
- Also added examples to snippet tests, to allow us to expand those later with edge cases we haven't thought about yet.


Sources/swift-openapi-generator/Documentation.docc/Development/Supporting-recursive-types.md
```diff
@@ -0,0 +1,130 @@
...
+    }
+}
+```
+
+Would look like this when boxed:
+```swift
+public struct Person {
+    public var partner: Person? {
+        set { CopyOnWriteBox.write(to: &storage) { $0.partner = newValue }}
+        get { storage.read().partner }
```
dnadoba, line 90: You will very likely need a `_modify` accessor here to not always trigger copy on write if mutated.

dnadoba: The problem is discussed in detail in this [swift forums topic](https://forums.swift.org/t/modify-accessors/31872).

czechboy0: I'll read up more on `_modify`, but in the meantime I tried running one of the tests I added and it doesn't ever hit the codepath that requires the copy: https://github.com/apple/swift-openapi-generator/pull/330/files#diff-01dc1dd8a9f6c46ca976e8e4b85f2afbfd8e89502fd35ac0c502b1a296673ec6R353-R365

Do you have a scenario in mind where only a `_modify` would avoid a copy?

czechboy0: Actually, I did get it working with `get` + `_modify` and one of my tests (one with 3 levels deep) did remove a copy this way, so I'll update the PR tomorrow. Thanks for the tip, @dnadoba! 🙏

czechboy0: Updated.

Sources/_OpenAPIGeneratorCore/Translator/Recursion/DeclarationRecursionDetector.swift
```diff
@@ -0,0 +1,204 @@
...
+            default:
+                return []
+            }
+        case .extension, .protocol, .function:
+            return []
+        }
+    }
+}
+
+fileprivate extension Array where Element == String {
```
dnadoba, line 175: FWIW this can now also be written as `extension Array<String> { ... }` or even `extension [String] { ... }` but that's just a matter of style.

dnadoba (approved):
LGTM

PR #331, opened 2023-10-14
Relax dependency on SwiftSyntax and SwiftFormat to 508..<510

mbrandonw:
### Motivation

Currently this project pins to a specific major version of SwiftSyntax and SwiftFormat, which is 508. That means anyone depending on this project cannot also use macros, because that typically requires depending on SwiftSyntax 509.

### Modifications

I relaxed the dependency constraint to be `"508.0.1"..<"510.0.0"`.

### Result

There is only one small change that needs to be done to support SwiftSyntax 509, and it can easily be worked around by checking `#if canImport(SwiftSyntax509)`.

### Test Plan

This change does complicate testing a bit since there are subtle formatting changes in SwiftFormat 508 vs 509. In particular, the Client.swift file has a few lines that get formatted onto 2 lines, whereas previously they were allowed to be on a single line.

This means the test suite will pass when SwiftFormat 508 is used but fail when 509 is used. So, for now, I've decided to just omit the test when 509 is available, but maybe there is some way to support both Swift versions in the test. I am not very familiar with this library or how it's tests are set up.

### DIFF
--- a/Tests/OpenAPIGeneratorReferenceTests/FileBasedReferenceTests.swift
+++ b/Tests/OpenAPIGeneratorReferenceTests/FileBasedReferenceTests.swift
@@ -46,9 +46,9 @@ class FileBasedReferenceTests: XCTestCase {
 
     func testPetstore() throws {
         #if canImport(SwiftSyntax509)
-            try _test(referenceProject: .init(name: .petstore))
+        try _test(referenceProject: .init(name: .petstore))
         #else
-            XCTFail("Update SwiftFormat to at least 509 to run this test.")
+        XCTFail("Update SwiftFormat to at least 509 to run this test.")
         #endif
     }
 
--- a/Tests/OpenAPIGeneratorReferenceTests/FileBasedReferenceTests.swift
+++ b/Tests/OpenAPIGeneratorReferenceTests/FileBasedReferenceTests.swift
@@ -45,7 +45,11 @@ class FileBasedReferenceTests: XCTestCase {
     }
 
     func testPetstore() throws {
-        try _test(referenceProject: .init(name: .petstore))
+        #if canImport(SwiftSyntax509)
+            try _test(referenceProject: .init(name: .petstore))
+        #else
+            XCTFail("Update SwiftFormat to at least 509 to run this test.")
+        #endif
     }
 
     // MARK: - Private
--- a/Package.swift
+++ b/Package.swift
@@ -55,13 +55,13 @@ let package = Package(
         // Generate Swift code
         .package(
             url: "https://github.com/apple/swift-syntax.git",
-            from: "508.0.1"
+            "508.0.1"..<"510.0.0"
         ),
 
         // Format Swift code
         .package(
             url: "https://github.com/apple/swift-format.git",
-            from: "508.0.1"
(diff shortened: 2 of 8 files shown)


czechboy0:
Hi @mbrandonw - thanks so much! Regarding the tests, let's keep with the times and make the tests pass with 5.9 instead, as that's what happens for a new project. If someone gets 5.8 resolved, that's okay, we know it's worked till now.

Otherwise looks great!

mbrandonw:
@czechboy0 No problem, pushed fixes to the tests for SwiftFormat 509.

Tests/OpenAPIGeneratorReferenceTests/FileBasedReferenceTests.swift
```diff
@@ -44,9 +44,11 @@ class FileBasedReferenceTests: XCTestCase {
         #endif
     }
 
-    func testPetstore() throws {
-        try _test(referenceProject: .init(name: .petstore))
-    }
+    #if canImport(SwiftSyntax509)
+        func testPetstore() throws {
+            try _test(referenceProject: .init(name: .petstore))
+        }
```
czechboy0, line 50: One more thing - to avoid silently not running the test, can you put the #if into the function and in the #else just fail with a descriptive message? Once this lands, nobody should be running tests of this package with the 5.8 version of the dependencies.

czechboy0:

Tests/OpenAPIGeneratorReferenceTests/FileBasedReferenceTests.swift
mbrandonw: Sure thing, I took a stab at the messaging. Feel free to tweak it in anyway you see fit.

czechboy0 (approved):
Lgtm, thanks! Please just check the CI failures, we'll need to get them to green before landing.

mbrandonw:
Hi @czechboy0, I don't know why the build is failing. The Jenkins logs show that all tests pass, yet it then says:

> Build step 'Execute shell' marked build as failure

Maybe someone more familiar with y'alls CI should take this over. I don't think there's much I can do to help.

czechboy0:
I can take a look Monday, but first thing I search for is "error:".

mbrandonw:
> error: Target SwiftSyntaxMacroExpansion imports another target (SwiftOperators) in the package without declaring it a dependency.

Looks like something for SwiftSyntax to address, or for OpenAPIGenerator to ignore:

https://github.com/apple/swift-syntax/blob/74203046135342e4a4a627476dd6caf8b28fe11b/Package.swift#L190-L194

czechboy0:
Okay, unfortunately it seems that swift-syntax has some missing dependencies, filed here: https://github.com/apple/swift-syntax/issues/2287

To make this PR pass, can you comment out the line `- IMPORT_CHECK_ARG=--explicit-target-dependency-import-check error` in all 4 docker-compose files in the docker directory? I filed another issue for us to re-enable it later: https://github.com/apple/swift-openapi-generator/issues/332

mbrandonw:
Looks like it's now passing. 🙂

czechboy0:
Thank you @mbrandonw and @groue!

PR #335, opened 2023-10-20
[Bugfix] Detect more schema type reference cycles.

czechboy0:
### Motivation

The newly introduced recursive type support had a slight optimization to minimize the number of boxed types, but it contained a bug that only surfaced on larger real-world documents. It's not really necessary, so removed it and the number of boxed types is still reasonable, plus it doesn't lead to non-compiling code anymore.

### Modifications

Removed an optimization that tried to further reduce the number of boxed types, because it actually missed some cycles and produced non-compiling code for some larger OpenAPI documents.

### Result

Code now compiles even for more complex cycles.

### Test Plan

Added more unit tests, plus tested on some real-world docs with cycles that failed before this change.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/Recursion/RecursionDetector.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Recursion/RecursionDetector.swift
@@ -85,12 +85,7 @@ struct RecursionDetector {
         //   document.
         // - Walk all references and keep track of names already visited.
         // - If visiting a schema that is already in the stack, we found a cycle.
-        // - In the cycle, first identify the set of types involved in it, and
-        //   check if any of the types is already recorded as a recursive type.
-        //   If so, no action needed and terminate this branch and continue with
-        //   the next one.
-        // - If no type in the cycle is already included in the set of recursive
-        //   types, find the first boxable type starting from the current one
+        // - Find the first boxable type starting from the current one
         //   ("causing" the recursion) following the cycle, and add it to this
         //   set, and then terminate this branch and continue.
         // - At the end, return the set of recursive types.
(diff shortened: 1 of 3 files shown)


Sources/_OpenAPIGeneratorCore/Translator/Recursion/RecursionDetector.swift
```diff
@@ -125,42 +122,38 @@ struct RecursionDetector {
...
 
             // We now choose which node will be marked as recursive.
             // Only consider boxable nodes, trying from the start of the cycle.
             guard let firstBoxable = cycleNodes.first(where: \.isBoxable) else {
                 throw RecursionError.invalidRecursion(name.description)
             }
 
+            let nameToAdd = firstBoxable.name
+
+            // Check if we're already going to box this type, if so, we're done.
```
glbrntt, line 145: Is it worth checking if any of the boxable `cycleNodes` are already in `boxed`?

glbrntt (approved):

Sources/_OpenAPIGeneratorCore/Translator/Recursion/RecursionDetector.swift
czechboy0: Hah, that's what I just removed in this PR. There are cases when multiple cycles that go through some common nodes where we ended up _not_ boxing enough nodes. My graph theory knowledge isn't deep and current enough to actually spend more time on it, so I'm leaving further optimization to future contributors. The important thing here is that we're going from an incorrect algorithm (that happened to work in simple cases) to a (more) correct algorithm that doesn't miss these nodes, unblocking some real-world docs. I'd love to see someone optimize this, we just have to be super careful not to miss any nodes that need boxing again.

PR #337, opened 2023-10-23
Respect the input order when filtering

czechboy0:
### Motivation

The original implementation of document filtering iterated over a Set and a Dictionary, which do not guarantee a stable order, meaning that every time the filter command was invoked with identical inptus, it could produce different outputs, potentially leading to needless rebuilds when used as a plugin, and needless commits when using ahead-of-time generation. It also made debugging of the generator itself more difficult.

### Modifications

Respect the input order of operations and components.

### Result

Invoking the generator multiple times with the same input produces the same output every time, improving cachability.

### Test Plan

Renamed the filter test to be consistent with the rest and manually tested on a larger document.

### DIFF
--- a/Tests/OpenAPIGeneratorCoreTests/Hooks/Test_FilteredDocument.swift
+++ b/Tests/OpenAPIGeneratorCoreTests/Hooks/Test_FilteredDocument.swift
@@ -16,7 +16,7 @@ import XCTest
 import Yams
 @testable import _OpenAPIGeneratorCore
 
-final class FilteredDocumentTests: XCTestCase {
+final class Test_FilteredDocument: XCTestCase {
 
     func testDocumentFilter() throws {
         let documentYAML = """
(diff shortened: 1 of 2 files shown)


Tests/OpenAPIGeneratorCoreTests/Hooks/Test_FilteredDocument.swift
```diff
@@ -16,7 +16,7 @@ import XCTest
 import Yams
 @testable import _OpenAPIGeneratorCore
 
-final class FilteredDocumentTests: XCTestCase {
+final class Test_FilteredDocument: XCTestCase {
```
glbrntt, line 19: Is it possible to add a test that filters a document twice and validates the output is identical?

glbrntt (approved):

Tests/OpenAPIGeneratorCoreTests/Hooks/Test_FilteredDocument.swift
czechboy0: IIRC I was seeing the order to be different between process launches, so we'd have to create a new kind of test where we spawn the CLI. But even then it could coincidentally work, and make the test itself flaky, which I'd like to avoid. We just need to keep an eye on any Set/Dictionary iteration in the future and always make it deterministic (e.g. using a `sort()`).

Issue #175, opened 2023-08-07
Update swift-format version to 5.9


takeshi-1000:
It seems that the following source code of Petstore was not properly formatted.
https://github.com/apple/swift-openapi-generator/blob/d22ddfeeef9d767386506ff6ec052e640839f1f8/Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Server.swift#L86

The source code inside swift-openapi-generator seems fine and I've confirmed that if I raise the swift-format version to the current commit revision, it's formatted properly. There seems to be no release version currently available on the swift-format side, so I open this issue.

If you want to check, change the code as below,

```diff
--- a/Package.swift
+++ b/Package.swift
@@ -52,7 +52,7 @@ let package = Package(
         // Format Swift code
         .package(
             url: "https://github.com/apple/swift-format.git",
-            from: "508.0.1"
+            revision: "8337523533bbd153cdac74c8fff4a1ac79a0b805"
         ),
 
         // General algorithms
```

and when you build it, an error will occur due to the change in the swift-format interface. (around [here](https://github.com/apple/swift-openapi-generator/blob/d22ddfeeef9d767386506ff6ec052e640839f1f8/Sources/_OpenAPIGeneratorCore/Extensions/SwiftFormat.swift#L44))

```
/path/to/swift-openapi-generator/Sources/_OpenAPIGeneratorCore/Extensions/SwiftFormat.swift:44:100: error: value of type 'SourceLocation' has no member 'debugDescription'
                    Formatting the following code produced diagnostic at location \(sourceLocation.debugDescription) (see end):
```

If you change the above code appropriately so that the build passes, and run the test, the output will be as follows.

```diff
diff --git a/ReferenceSources/Petstore/Server.swift b/var/folders/nn/80g9wmmd37lc63c9wqtnspj40000gp/T/5705E10F-715B-4E33-8295-2090E9E2FA85/Server.swift
index cd60e0c..fc5d004 100644
--- a/ReferenceSources/Petstore/Server.swift
+++ b/var/folders/nn/80g9wmmd37lc63c9wqtnspj40000gp/T/5705E10F-715B-4E33-8295-2090E9E2FA85/Server.swift
@@ -81,11 +81,12 @@ fileprivate extension UniversalServer where APIHandler: APIProtocol {
         try await handle(
             request: request,
             with: metadata,
             forOperation: Operations.listPets.id,
             using: { APIHandler.listPets($0) },
[...]

PR #340, opened 2023-10-23
Bump swift-format to 5.9

czechboy0:
### Motivation

As per https://github.com/apple/swift-openapi-generator/issues/175.

### Modifications

Updated to swift-format 5.9, updated code to pass soundness.

### Result

Using swift-format 5.9.

### Test Plan

Soundness passed.

### DIFF
--- a/Tests/PetstoreConsumerTests/Test_Types.swift
+++ b/Tests/PetstoreConsumerTests/Test_Types.swift
@@ -17,6 +17,7 @@ import PetstoreConsumerTestCore
 
 final class Test_Types: XCTestCase {
 
+    /// Setup method called before the invocation of each test method in the class.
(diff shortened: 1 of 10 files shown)


Tests/PetstoreConsumerTests/Test_Types.swift
```diff
@@ -17,6 +17,7 @@ import PetstoreConsumerTestCore
 
 final class Test_Types: XCTestCase {
 
+    /// Setup method called before the invocation of each test method in the class.
```
dnadoba, line 20: hm... we might want to have a different swift-format docs policy for test targets but this can be in a separate PR if we decided to do it

dnadoba (approved):

PR #345, opened 2023-10-25
Add a soundness --fix flag

czechboy0:
### Motivation

When running `./scripts/soundness.sh` produces swift-format warnings, we ask adopters to manually copy/paste a call to swift format to fix the warnings up. This is tedious and unnecessary.

### Modifications

Add a `--fix` option on the `soundness.sh` script to actually apply the fixes as well, avoiding the need to copy/paste long commands.

### Result

Easier fixing up of formatting warnings.

### Test Plan

Manually tested the workflow locally.

### DIFF
--- a/scripts/soundness.sh
+++ b/scripts/soundness.sh
@@ -21,11 +21,17 @@ fatal() { error "$@"; exit 1; }
 CURRENT_SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 NUM_CHECKS_FAILED=0
 
+FIX_FORMAT=""
+for arg in "$@"; do
+  if [ "$arg" == "--fix" ]; then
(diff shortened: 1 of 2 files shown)


scripts/soundness.sh
```diff
@@ -35,6 +41,13 @@ for SCRIPT_PATH in "${SCRIPT_PATHS[@]}"; do
   fi
 done
 
+log "Running swift-format..."
+bash ${CURRENT_SCRIPT_DIR}/run-swift-format.sh $FIX_FORMAT > /dev/null
```
glbrntt, line 45:
```suggestion
bash "${CURRENT_SCRIPT_DIR}"/run-swift-format.sh $FIX_FORMAT > /dev/null
```

glbrntt (approved):

PR #346, opened 2023-10-25
Disable "respectsExistingLineBreaks" in .swift-format for more consistent styling

### LINKED ISSUE #230: Disable "respectsExistingLineBreaks" in .swift-format for more consistent styling
Disable "respectsExistingLineBreaks" in .swift-format for more consistent styling.

Today, `respectsExistingLineBreaks` is enabled, which means that it depends on how you write the code before running the formatter, and it affects the output. That's not ideal, and is different than the setting we have for formatted code, so affects #40.

This should be done as a single PR, as it'll cause some churn.

PARAIPAN9:
### Motivation

- Fixes #230

### Modifications

- Disable respectsExistingLineBreaks .swift-format rule and address changes requested

### Result

- One of the .swift-format rules will be disabled

### Test Plan

- Run Tests


PARAIPAN9:
I had to revert the changes from the previous PR because there were too many conflicts. I've been waiting for the PR section to be empty.

czechboy0 (changes requested):
Just two comments, otherwise looks great, thank you @PARAIPAN9!

If you have more time, it'd be great to apply the same change to the other 3 repos, we want to keep all 4 consistent.

czechboy0:
Still seems the symlink got changed, unless github is slow to refresh the diff.

PARAIPAN9:
Yes, I see...

PARAIPAN9:
I was on -v 5.9, I've just updated the latest version and addressed complaints

czechboy0 (approved):

czechboy0:
Great, thank you @PARAIPAN9! I think the other 3 repos will be easier 🙂

PARAIPAN9:
You're welcome, I'm pretty sure too 😄.

PR #353, opened 2023-10-30
Disable warnings-as-errors for nightlies

czechboy0:
### Motivation

In Swift 5.11/nigtlies Swift upgraded preconcurrency diagnostics from a remark to a warning, and since we run CI with warnings-as-errors, that broke it. I tried removing the preconcurrency attribute on imports in #352, but we still need it.

Since main builds of Swift are still in-development, we shouldn't be so sensitive to changes there, so disabling warnings-as-errors for nightlies only (keeping for 5.8, 5.9, and 5.10).

### Modifications

Disable warnings as errors on CI for nightlies.

### Result

CI should pass again.

### Test Plan

CI.


PeterAdams-A (approved):
looks good

PR #359, opened 2023-11-01
Use granular RetroactiveAttribute feature guard

rnro:
### Motivation

To protect against corner-cases where the more coarse compiler guard is insufficient.

### Modifications

Use granular `RetroactiveAttribute` feature guard

### Result

Safer code building.

### Test Plan

Tested code building with 5.8 on Linux


czechboy0 (approved):
Thank you!

PR #363, opened 2023-11-06
Fix spelling of retroactive guard

rnro:
### Motivation:

`hasFeature(RetroactiveAttribute)` doesn't work as expected, but `$RetroactiveAttribute` does.

### Modifications:

Switch from `hasFeature` to `$RetroactiveAttribute`.

### Result:

- `@retroactive` is applied appropriately


czechboy0:

czechboy0 (approved):

PR #380, opened 2023-11-14
Fix a bug in path parameter ordering

### LINKED ISSUE #378: Path Parameter Ordering
I can't tell if this is an issue with how the OpenAPI is setup, but I cannot find anywhere it explicitly states that the `parameters` is required to be ordered.  

Here is the OpenAPI snippet:
```
"/config/settings/models/{model}/brands/{brand}/versions/{version}": {
    "get": {
        "parameters": [
            {
                "in": "path",
                "name": "version",
                "required": true,
                "schema": {
                    "type": "string"
                }
            },
            {
                "in": "path",
                "name": "model",
                "required": true,
                "schema": {
                    "type": "string"
                }
            },
            {
                "in": "path",
                "name": "brand",
                "required": true,
                "schema": {
                    "type": "string"
                }
            }
        ],
```

It contains three path parameters `model`, `brand`, and `version` ordered correctly within the path, but the `parameters` array field has them unordered with `version`, `model`, and `brand` with correct `name` fields matching those within the path. 

The generated Client code results is a an array which retains the order of the parameters list, but make to attempt to align based on the `name` key.
```
let path = try converter.renderedPath(
  template: "/config/settings/models/{}/brands/{}/versions/{}",
  parameters: [input.path.version, input.path.model, input.path.brand]
)
```

Is there a way I can generate the swift code to use a Dictionary instead of an array, or is the initial OpenAPI not to spec?

czechboy0:
### Motivation

Fixes #378.

Basically, if the path parameter order was different between the URL template `/a/{a}/b/{b}` and the list of path parameters, for example if `b` came before `a` there, we used the parameter order, not the URL template order, which was _incorrect_. I suspect this bug has been hiding for this long is because most documents use the same order in both.

### Modifications

Change the algorithm that composes the URL template and the list of parameters to respect the URL template's order.

### Result

The URL template order is now respected, regardless of the path parameter order.

### Test Plan

Added a snippet test that deliberately uses a different order between the URL template and the path paramater list. It failed before this fix, passes with it.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/Operations/OperationDescription.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Operations/OperationDescription.swift
@@ -251,17 +251,32 @@ extension OperationDescription {
     /// For example, `/cats/{}` and `[input.catId]`.
     var templatedPathForClient: (String, Expression) {
         get throws {
-            let path = self.path.rawValue
-            let pathParameters = try allResolvedParameters.filter { $0.location == .path }
-            // replace "{foo}" with "{}" for each parameter
-            let template = pathParameters.reduce(into: path) { partialResult, parameter in
-                partialResult = partialResult.replacingOccurrences(of: "{\(parameter.name)}", with: "{}")
+            let pathParameterNames = try Set(allResolvedParameters.filter { $0.location == .path }.map(\.name))
+            var orderedPathParameters: [String] = []
+            // Replace "{foo}" with "{}" for each parameter and record the order
+            // in which the parameters are used.
+            var newComponents: [String] = []
+            for component in path.components {
+                guard component.hasPrefix("{") && component.hasSuffix("}") else {
+                    newComponents.append(component)
+                    continue
+                }
+                let componentName = String(component.dropFirst().dropLast())
+                guard pathParameterNames.contains(componentName) else {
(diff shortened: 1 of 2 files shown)


czechboy0:

simonjbeaumont (approved):
Does this work for repeated path parameters? e.g. `/a/{a}/b/{b}/{a}`?

czechboy0:
> Does this work for repeated path parameters? e.g. `/a/{a}/b/{b}/{a}`?

I don't think that's even spell-able in OpenAPI? If you use an array path parameter, it'd render as e.g. `/a/1,2,3`. I don't know of a way to get `/a/1/2/3`. Maybe it's possible, but this PR doesn't change that, we didn't support it before, and still don't support it now.

czechboy0:

simonjbeaumont:
> I don't think that's even spell-able in OpenAPI? If you use an array path parameter, it'd render as e.g. `/a/1,2,3`. I don't know of a way to get `/a/1/2/3`. Maybe it's possible, but this PR doesn't change that, we didn't support it before, and still don't support it now.

I was expecting it to be `/a/1/2/1` if `a = 1` and `b=2`. But if it's not a thing you can express in OpenAPI, then that's fine.

czechboy0:

PR #381, opened 2023-11-14
Skip inferred properties

### LINKED ISSUE #379: Filter out fragment properties only appearing in the required list
Adopt OpenAPIKit 3.1.0, which allows us to differentiate now.

https://github.com/mattpolzin/OpenAPIKit/releases/tag/3.1.0

We should emit a warning diagnostic when an inferred property is encountered and filtered out.

czechboy0:
### Motivation

Fixes #379.

Context:

If you define an object schema as follows:

```yaml
Foo:
  type: object
  properties:
    a:
      type: string
  required:
    - b
```

OpenAPIKit will parse it as having two properties: `a` of type `string` and `b` being a fragment, with `a` optional and `b` required.

Unfortunately, often property names appear _only_ in the `required` list as a result of a typo, leading to non-compiling code and a difficult debugging story.

It comes down to the fact that Swift OpenAPI Generator never officially supported defining properties this way, and continues not to.

If adopters want to do this intentionally, they should do this: 

```yaml
Foo:
  type: object
  properties:
    a:
      type: string
    b: {}
  required:
    - b
```

### Modifications

This PR tightens the rules around inferred properties:

1. Inferred properties are skipped, so the result of the first example will only have the property `a` generated, and the second example both `a` and `b`.
2. The generator emits a warning diagnostic when encountering an inferred property, which should help adopters who made a typo catch it much earlier in the process.
3. Bumps OpenAPIKit to 3.1.0, as the `inferred` property was added just yesterday.

### Result

Less likely typos and mismatch between `properties` and `required`, and a diagnostic to help debug this case.

### Test Plan

Added a snippet test.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -55,7 +55,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-algorithms", from: "1.0.0"),
 
         // Read OpenAPI documents
-        .package(url: "https://github.com/mattpolzin/OpenAPIKit.git", from: "3.0.1"),
+        .package(url: "https://github.com/mattpolzin/OpenAPIKit.git", from: "3.1.0"),
         .package(url: "https://github.com/jpsim/Yams.git", "4.0.0"..<"6.0.0"),
 
         // CLI Tool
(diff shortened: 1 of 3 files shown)


czechboy0:

simonjbeaumont (approved):

PR #384, opened 2023-11-17
Add deprecated annotations on shorthand functions for deprecated operations

### LINKED ISSUE #367: Operation deprecations don't seem to be added to convenience methods on the APIProtocol extension
Operation deprecations don't seem to be added to convenience methods on the APIProtocol extension.

This means that when an operation is marked as deprecated, we still get warnings from the Types.swift file in the place where the convenience method calls into the Input -> Output operation method.

The deprecation attribute should be added on the extension method as well, if it's on the original operation method.

simonjbeaumont:
### Motivation

If an operation is marked as deprecated in the OpenAPI document then the generated protocol requirement is annotated as deprecated. However, the shorthand overload for this function, in the protocol extension, was missing this annotation.

### Modifications

Add deprecated annotation to the generated function in the protocol extension for deprecated API operations.

### Result

If an operation is marked as deprecated in the OpenAPI document then both variants of the generated function (the protocol requirement _and_ the protocol extension) are annotated as deprecated.

### Test Plan

Updated the snippet test to include a path with one deprecated operation and one non-deprecated operation.

### Related Issues

- Fixes #367.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateAPIProtocol.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateAPIProtocol.swift
@@ -78,7 +78,11 @@ extension TypesFileTranslator {
                         )
                     ]
                 )
-                return .commentable(operation.comment, .function(function))
+                if operation.operation.deprecated {
+                    return .commentable(operation.comment, .deprecated(.init(), .function(function)))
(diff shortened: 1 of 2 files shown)


czechboy0 (approved):

groue:
I just met this one today :-)

czechboy0:
@groue we published a release today with the fix

PR #385, opened 2023-11-17
Fix compilation of tests on older macOS

simonjbeaumont:
### Motivation

The tests fail to compile on anything older than macOS 13, because they use `Collection.split(separator:maxSplits:omittingEmptySubsequences:)`, which is only available in macOS 13+.

### Modifications

Rather than use `->` in the stringified tests, use just `>`, which means the function resolves to `String.split(separator:maxSplits:omittingEmptySubsequences:)`, which is available on older platforms.

### Result

Tests can be built on older platforms.

### Test Plan

Manual verification.


czechboy0 (approved):

czechboy0:
Whoops, thanks

PR #393, opened 2023-11-27
Use the package access modifier for generated code

### LINKED ISSUE #270: Use the `package` access modifier for generated code
Swift 5.9 adds the new `package` access modifier (see the [evolution proposal](https://github.com/apple/swift-evolution/blob/main/proposals/0386-package-access-modifier.md)). 

This is a very natural fit for the generated code as:
- Users are encouraged to wrap the generated types rather than vending them directly as part of their package.
- Users are encouraged to generate code for into separate modules per document.

However, it's likely that some users will still want to vend generated code.

I think the generator should provide a configuration option for visibility so that the user can decide. The default should likely be set to the new `package` access level.

czechboy0:
### Motivation

Fixes #270.

The `package` access modifier seems like the right default access modifier for generated code, as it forces adopters to explicitly opt into generated APIs leaking outside of their package. `internal` was too limiting, as often you want to isolate generated code into a separate module. Fortunately, Swift 5.9 provided that feature and since 1.0 will ship with 5.9 as minimum Swift version, we can switch to `package` as default.

### Modifications

- (Re)added the ability to customize the access modifier of the generated code. You can do so from the config file and with a CLI option.
- The default has changed from `public` to `package` (a breaking change).
- Updated docs (some drive-by docs improvements as well for which OpenAPI features are supported.)

### Result

Generated code uses the `package` access modifier by default, but the user can change it to `public` or `internal`, if needed.

### Test Plan

Manually tested the CLI path, the config file path, updated tests, but didn't actually change our fixtures, as there's no need. We just explicitly override to `public`.

I also tried this branch on a real-world document using the generator plugin, and tried all 3 access modifiers and verified all 3 worked (the project built successfully).


Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
simonjbeaumont, line 1: Could we have one snippet test that shows that changing it actually results in a change in the rendered code?

simonjbeaumont (changes requested):
Overall looks good.

I think we should park this behind dropping support for 5.8, which will involve updating the Swift version used to run the integration test, which, I take it, is why its currently failing?

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
czechboy0: Added

simonjbeaumont (approved):

PR #396, opened 2023-11-27
Remove no-longer-needed `@preconcurrency`s 

MahdiBM:
### Motivation

The closed-source Darwin Foundation had marked `Date`, `Data` marked as Sendable for a long time.
Now both swift-foundation and the swift-corelibs-foundation have them as Sendable as well on Swift 5.9 (5.9.1+ ?):
The PRs for swift corelibs-foundation can be found [here](https://github.com/apple/swift-corelibs-foundation/pulls?q=is:pr+author:franzbusch+is:closed).
We no longer need those `@preconcurrency`s.

### Modifications

Remove `@preconcurrency`s for `Foundation.Date`, `.Data`.

### Result

Less warnings in the logs.

### Test Plan

Tested on Swift 5.9.1 on Linux.


MahdiBM:
apparently URL is still not Sendable. I thought it was. I'll modify the PR.

Sources/_OpenAPIGeneratorCore/Translator/CommonTypes/Constants.swift
```diff
@@ -28,7 +28,11 @@ enum Constants {
...
             ImportDescription(
                 moduleName: "Foundation",
-                moduleTypes: ["struct Foundation.URL", "struct Foundation.Data", "struct Foundation.Date"],
+                moduleTypes: ["struct Foundation.Data", "struct Foundation.Date"]
+            ),
+            ImportDescription(
+                moduleName: "Foundation",
+                moduleTypes: ["struct Foundation.URL"],
                 preconcurrency: .onOS(["Linux"])
             ),
```
MahdiBM, line 37: not sure if these 2 Foundation imports will cause problems, but pointing out just incase.

czechboy0:

czechboy0:

czechboy0:
@MahdiBM This one will be tricky, as the generated code doesn't know which version it'll run in, so it needs to work on 5.9.0, even if it produces remarks/warnings on newer Linux toolchains.

simonjbeaumont:
> @MahdiBM This one will be tricky, as the generated code doesn't know which version it'll run in, so it needs to work on 5.9.0, even if it produces remarks/warnings on newer Linux toolchains.

I'm not sure I understand this comment. The generated code will still be compiled, just not by us. At the time of compilation we can still check `#if swift(...)` and if we don't produce warnings on any of our CIs then it shouldn't produce warnings for adopters. 

Not counting the case where they fix more issues in Linux toolchains that aren't yet released; for prerelease toolchains we shouldn't expect to be warning-free 100% of the time. 

Maybe you were saying we cannot generate code that expects a certain version, the, yes, I agree: the generated code needs to handle all supported versions, but that doesn't mean we can't make it warning free for _both_ Swift 5.9.0 and 5.9.1, or am I misunderstanding something?

czechboy0:
You're right, if we're willing to _generate_ the conditional compilation code, which I guess there isn't a reason not to.

simonjbeaumont:
> You're right, if we're willing to _generate_ the conditional compilation code, which I guess there isn't a reason not to.

Not only is there no reason not to, this is the only place where it matters. If adopters are compiling with warnings-as-errors then warnings in dependency libraries do not cause errors in the downstream build, but warnings in the generated code will. 

Appreciate in this case that these may be remarks or notes (cf. warnings), but in principle the generated code needs to be as clean as we can make it.

MahdiBM:
Tests Please 🙂

MahdiBM:
@czechboy0 ok that had some obvious problems. Another try please. Hopefully this one or next one should be good.

simonjbeaumont:

simonjbeaumont:
@MahdiBM you should be able to test this locally using:

```
docker-compose -f docker/docker-compose.yaml -f docker/docker-compose.2204.590.yaml run test
```

MahdiBM:
I'm kind of lost why the 5.9.0 tests aren't passing:
```swift
```

MahdiBM:
To be honest it could be one of those cases where I'm getting something obvious wrong, but as far as I can tell, yes, I'm saying this is a compiler bug.

I'm not sure how easy it'll be to catch it In a vaccum, but I can try.

I'll also be happy if you could checkout this branch and run the Swift 5.9.0 tests locally and confirm it does seem suspicious to you. It should as simple as you verifying that the compiler is throwing Date-not-Sendable warnings when @preconcurrency is already there for the Date import 🙂

czechboy0:
We still have a long list of things we need to do for 1.0, so I won't be able to dig into this before then. Feel free to leave this PR open and move it to draft until post-1.0, then we can investigate further.

czechboy0:
Marking this as draft for now.

czechboy0:
Briefly tried to play with this today, but it's just too many changes at once to be able to test this between different versions.

I wonder if you could actually break this up into individual PRs, each PR fixing a single type's remarks/warnings? That should allow us to take better advantage of CI instead of having to spend the time rerunning things locally.

MahdiBM:
@czechboy0 Sure. I can break it down to 2 PRs:
1- Use `canImport(Darwin)` instead of the current `os(Linux)`. Will perhaps be slightly more correct and also inline with other OpenAPI libraries.
2- Try to resolve the warnings.

PR #399, opened 2023-11-28
Switch default visibility from package to public

czechboy0:
### Motivation

The `package` access modifier seems to have issues for files produced by plugins, so let's switch back to `public`.

### Modifications

Switch the default access modifier to `public`.

### Result

The default modifier is `public` again.

### Test Plan

Updated tests.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Config.swift
+++ b/Sources/_OpenAPIGeneratorCore/Config.swift
@@ -27,7 +27,7 @@ public struct Config: Sendable {
     public var access: AccessModifier
 
     /// The default access modifier.
-    public static let defaultAccessModifier: AccessModifier = .package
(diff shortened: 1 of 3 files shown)


simonjbeaumont:
I think we should take a beat and consider whether `public` is the right default here. I think there's a strong case to be made for `internal` to make `public` an opt-in and/or making the `accessModifier` a required config and command line option to make it explicit. The former, at least, would be more consistent with other workflows for factoring handwritten code into modules and with other generators (e.g. grpc-swift).

czechboy0:
Yup, discussed offline, I'll switch it to internal tomorrow.

czechboy0:
It was user error, all is well if you use the 5.9 Swift tools version.

PR #400, opened 2023-11-29
Bump to 1.0.0-alpha.1

czechboy0:
### Motivation

Prep for 1.0.0-alpha.1.

### Modifications

Bumped documentation and upstream SOAR dependencies to 1.0.0-alpha.1.

Note that because the generator itself is not yet released, the integration test and examples will fail to build, but that's the unfortunate nature of having examples projects in the same repo as the project itself. But it's important that once this repo is tagged, that the examples/integration tests included use the right tags, so I don't think there's another way to stage this in.

### Result

Ready to tag 1.0.0-alpha.1.

### Test Plan

Temporarily modified the generator dependency in examples/integration tests to point to `main` and verified all works. Then changed it back.

All other tests pass.

### DIFF
--- a/Examples/GreetingService/Package.swift
+++ b/Examples/GreetingService/Package.swift
@@ -18,9 +18,9 @@ let package = Package(
     name: "GreetingService",
     platforms: [.macOS(.v13)],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-openapi-generator", .upToNextMinor(from: "0.3.0")),
-        .package(url: "https://github.com/apple/swift-openapi-runtime", .upToNextMinor(from: "0.3.0")),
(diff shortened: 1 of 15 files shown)


simonjbeaumont:
OK, the interdependent integration test and examples is a bother, I agree. Let's hook up and test this out locally before we merge.

czechboy0:
> OK, the interdependent integration test and examples is a bother, I agree. Let's hook up and test this out locally before we merge.

Verified locally and made one more fix: https://github.com/apple/swift-openapi-generator/pull/400/commits/f1376000d774ae4e75aeab4206c30678c44a81ef

simonjbeaumont (approved):

PR #401, opened 2023-11-29
Remove the base64 feature flag

czechboy0:
### Motivation

In preparation for 1.0.0-alpha.1, we enable and remove all feature flags.

### Modifications

Remove the feature flag.

Now, the fun thing I just discovered is that the feature flag wasn't correctly wired up, so we accidentally broke some users in 0.3.x when base64 landed, as it changed `Swift.String` to `OpenAPIRuntime.Base64EncodedData` _unconditionally_. This wasn't caught in PR review and unfortunately didn't have a negative test that would have caught this.

### Result

Removed a stale feature flag.

### Test Plan

All tests pass.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/FeatureFlags.swift
+++ b/Sources/_OpenAPIGeneratorCore/FeatureFlags.swift
@@ -28,11 +28,6 @@
 public enum FeatureFlag: String, Hashable, Codable, CaseIterable, Sendable {
     // needs to be here for the enum to compile
     case empty
-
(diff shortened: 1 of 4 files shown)


simonjbeaumont (commented):
~What am I missing here. I don't see any conditional being removed. Was this being ignored already and we're just removing a vestigial boolean property (along with the feature flag)?~

Read your description, seems my assessment was correct from the diff. Approving.

simonjbeaumont (approved):

PR #411, opened 2023-11-30
Add Docker Compose file for Swift 5.9.0

simonjbeaumont:
### Motivation

As we approach 1.0 we want to make sure that we compile without warnings (or remarks) on all the Swift versions we claim to support. As discussed in https://github.com/apple/swift-openapi-urlsession/pull/35, this turns out to be subtly different between 5.9.0 and 5.9.1. We cannot drop support for 5.9.0 because the latest released (non-seed) Xcode still uses 5.9.0. However 5.9.1 is the latest release for Linux.

While we don't expect to do this for all patch versions of Swift, in this instance we'd like to add a pipeline for Swift 5.9.0 until such a time that we can expect everyone to be using 5.9.1, which will be some time after an Xcode release with this version.

This will allow us to progress squashing the remarks on 5.9.1 with confidence that we aren't introducing warnings on 5.9.0, which could result in an error for adopters if they compile our generated code with warnings-as-errors.

### Modifications

This PR adds a new Docker Compose file to explicitly use `5.9.0`.

It leaves the existing `5.9` pipeline to pull `5.9-jammy` which resolves to `5.9.1`, and will continue to resolve to the latest 5.9.x version.

### Result

We can now stand up a new CI pipeline for Swift 5.9.0 explicitly, which we can use this PR to validate.

### DIFF
--- a/docker/docker-compose.2204.590.yaml
+++ b/docker/docker-compose.2204.590.yaml
@@ -1,0 +1,18 @@
+version: "3"
+
+services:
+  runtime-setup:
+    image: &image swift-openapi-generator:22.04-5.9.0
+    build:
+      args:
+        ubuntu_version: "jammy"
+        swift_version: "5.9.0"
+
+  test:
+    image: *image
+    environment:
+      - WARN_AS_ERROR_ARG=-Xswiftc -warnings-as-errors
+      - IMPORT_CHECK_ARG=--explicit-target-dependency-import-check error
+      - STRICT_CONCURRENCY_ARG=-Xswiftc -strict-concurrency=complete
+  shell:
+    image: *image


czechboy0 (approved):

simonjbeaumont:
Hi @yim-lee would you be able to help us set up another pipeline (see the description for the reason) to test 5.9.0 explicitly? 

I've added the compose file and it should look just like the 5.9 one, I guess; so, `pull request validation (5.9.0)`.

Thanks in advance 🙏 

There'll be equivalent PRs in the other repos:
- swift-openapi-urlsession
- swift-openapi-async-http-client
- swift-openapi-runtime

simonjbeaumont:
Just to show the results:

```
❯ docker-compose -f docker/docker-compose.yaml -f docker/docker-compose.2204.59.yaml run --build shell

root@47ba3d0065e1:/code# swift -version
Swift version 5.9.1 (swift-5.9.1-RELEASE)
Target: aarch64-unknown-linux-gnu

❯ docker-compose -f docker/docker-compose.yaml -f docker/docker-compose.2204.590.yaml run shell

root@a0be1db899d9:/code# swift -version
Swift version 5.9 (swift-5.9-RELEASE)
Target: aarch64-unknown-linux-gnu
```

PR #415, opened 2023-12-01
[Examples] Add HelloWorldClient and HelloWorldServer

czechboy0:
### Motivation

Adding examples, there'll be a bunch of these.

### Modifications

See title.

### Result

N/A

### Test Plan

Tested locally, the `examples` CI will also build and test them.

Verified that if I run the server in one Terminal tab with `swift run`, and then run the client in another with `swift run`, they successfully talk to each other.

### DIFF
--- a/Examples/HelloWorldClient/Sources/GreetingServiceClient/Tool.swift
+++ b/Examples/HelloWorldClient/Sources/GreetingServiceClient/Tool.swift
@@ -1,0 +1,27 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftOpenAPIGenerator open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftOpenAPIGenerator project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import OpenAPIRuntime
+import OpenAPIURLSession
+import Foundation
+
+@main struct Tool {
+    static func main() async throws {
+        let client = Client(
+            serverURL: URL(string: "http://localhost:8080/api")!,
+            transport: URLSessionTransport()
+        )
+        let response = try await client.getGreeting()
+        print(try response.ok.body.json.message)
+    }
+}
--- a/Examples/HelloWorldClient/Sources/GreetingServiceClient/openapi.yaml
+++ b/Examples/HelloWorldClient/Sources/GreetingServiceClient/openapi.yaml
@@ -1,0 +1,36 @@
+openapi: '3.1.0'
+info:
+  title: GreetingService
+  version: 1.0.0
+servers:
+  - url: https://example.com/api
+    description: Example service deployment.
+paths:
+  /greet:
+    get:
+      operationId: getGreeting
+      parameters:
+      - name: name
+        required: false
+        in: query
+        description: The name used in the returned greeting.
+        schema:
+          type: string
+      responses:
+        '200':
+          description: A success response with a greeting.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Greeting'
+components:
+  schemas:
+    Greeting:
+      type: object
+      description: A value with the greeting contents.
+      properties:
+        message:
+          type: string
+          description: The string representation of the greeting.
+      required:
+        - message
--- a/Examples/HelloWorldServer/.gitignore
+++ b/Examples/HelloWorldServer/.gitignore
@@ -1,0 +1,11 @@
+.DS_Store
+.build
+/Packages
+/*.xcodeproj
+xcuserdata/
+DerivedData/
+.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
+.vscode
+/Package.resolved
+.ci/
+.docc-build/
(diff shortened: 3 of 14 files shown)


Examples/HelloWorldServer/Package.swift
```diff
@@ -0,0 +1,45 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import PackageDescription
+
+let package = Package(
+    name: "HelloWorldServer",
+    platforms: [.macOS(.v13)],
```
simonjbeaumont, line 19: I think this is overly restrictive these days and dates back to before we relaxed things, pre 0.1.

Examples/HelloWorldServer/Package.swift
```diff
@@ -0,0 +1,45 @@
...
+    dependencies: [
+        .package(url: "https://github.com/apple/swift-openapi-generator", exact: "1.0.0-alpha.1"),
+        .package(url: "https://github.com/apple/swift-openapi-runtime", exact: "1.0.0-alpha.1"),
+        .package(url: "https://github.com/swift-server/swift-openapi-vapor", exact: "1.0.0-alpha.1"),
+        .package(url: "https://github.com/vapor/vapor", from: "4.76.0"),
```
simonjbeaumont, line 24: Weren't there some bug fixes related to streaming in Vapor? I realise we'd probably get them by resolution, but let's set this to something more recent, that we know has the fixes we need.

Examples/HelloWorldServer/Sources/HelloWorldServer/openapi.yaml
simonjbeaumont, line 1: For the examples where we have a client-server pair, should these just be symlinks?

EDIT: I take it this is intentional based on the text in `Examples/README`: 

> All the examples are self-contained, so you can copy the package directory of your chosen example and use it as a starter template for your project.

Examples/HelloWorldServer/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Run tests using:
+
+```
+swift test
+```
+
+The testing strategy is to call the `Handler` directly from tests, as it conforms the `APIProtocol` and implements the business logic.
+
+This allows you to provide any input to the handler methods and verify that the correct outputs are returned, such as error responses when the input data is invalid, and so on.
```
simonjbeaumont, line 40: Could we emphasise here that there's no transport involved in these tests so you preserve the ability to swap to another one at any time and no networking types or connectivity is required to run the tests.

simonjbeaumont (commented):
Thanks for starting this. As you suggested we should probably do some bikeshedding at this point, before we start copying the pattern across to all the others so I've added some comments for discussion.

Examples/HelloWorldClient/Sources/GreetingServiceClient/Tool.swift
simonjbeaumont, line 1: Do we have to use `Tool.swift`? What's wrong with `HelloWorldClient.swift`?

Examples/HelloWorldServer/Sources/HelloWorldServer/Tool.swift
simonjbeaumont, line 1: Do we have to use `Tool.swift`? What's wrong with `HelloWorldServer.swift`?

czechboy0: I generally like to avoid naming all the layers the same name, as it can then be confusing for someone going to rename to match their project which things should be named what. I think the package name == module name are an okay tradeoff, but otherwise I prefer unique names. Or even better, names that inform the role of the file/type, rather than a long prefix + short suffix that is still pretty general.

Examples/HelloWorldServer/Sources/HelloWorldServer/openapi.yaml
czechboy0: Yup, exactly.

Examples/HelloWorldServer/Package.swift
czechboy0: Sure, I can update to the latest released one.

Examples/HelloWorldServer/Package.swift
czechboy0: Oh right, yeah let me try to decrease it.

Examples/HelloWorldServer/.gitignore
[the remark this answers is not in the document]
czechboy0: This is the default gitignore you get when you run `swift package init`, so I kept it in there so that people can copy this directory and use it as their project. It'd be a shame if they then had to go copy this .gitignore from somewhere else. I think this is part of the "self-contained project" constraint.

Examples/HelloWorldClient/Sources/GreetingServiceClient/openapi.yaml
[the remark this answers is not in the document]
czechboy0: Sure, good idea.

Examples/HelloWorldClient/Sources/GreetingServiceClient/Tool.swift
czechboy0: See my answer on the server one below.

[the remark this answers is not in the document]
czechboy0: Whoops meant to rename that 🤦‍♂️ Will fix.

Examples/HelloWorldServer/Sources/HelloWorldServer/Tool.swift
czechboy0: Outcome: use the same name of the sample everywhere.

Examples/HelloWorldServer/README.md
czechboy0: Updated.

simonjbeaumont (approved):

PR #418, opened 2023-12-03
Update compat suite OpenAPI documents to their latest versions

czechboy0:
### Motivation

Update the docs used by the compat suite, to be representative for 1.0.

### Modifications

Updated to the latest versions of all docs.

### Result

Using the latest versions.

### Test Plan

CI passed (both generation and building), and locally I ran both GitHubs, as they don't build in CI even in the full pipeline.


czechboy0:
> Looks like even the non-build compat test failed. 🥴

Yup, I didn't actually run GitHub and GitHub Enterprise locally, I'll make the necessary fixes. Seems it's just the expected diagnostics changed.

czechboy0:

czechboy0:
Ok I also ran and built both GitHub's locally (they're disabled in CI). All good.

czechboy0:
> We could potentially filter GitHub before building and/or increase the timeout, given this job doesn't run on every PR.

I don't think we build either GitHub in the full pipeline, so it's strange that it still timed out. Should I disable another large doc from building to get it through?

simonjbeaumont:
Potentially we regressed our generation and/or compile times since we introduced this pipeline. Be curious to understand by how much, the numbers posted by the adopter on the other issue seem to suggest quite a bit.

czechboy0:
Didn't we also last test this on 5.8, and switched to 5.9 since?

czechboy0:
Yeah, last time this passed was in early October still on 5.8: https://ci.swiftserver.group/job/swift-openapi-generator-compatibility-test-full-prb/12/

simonjbeaumont:
Probably worth getting this to run periodically. 

So there's a few variables here:
- the generator version
- the document versions
- the toolchain version

simonjbeaumont (approved):

PR #420, opened 2023-12-04
[Docs] Add an article about spec-driven development

czechboy0:
### Motivation

Swift OpenAPI Genearator was designed to be used with spec-driven development, so we should have an article of what it is and how to practice it in our docs.

### Modifications

Add an article about spec-driven development, compare it with code-driven development, and describe how to migrate to it.

### Result

More design decisions of Swift OpenAPI Generator should make sense with this extra context.

### Test Plan

Edited it myself a bit, but more polish needed.

### DIFF
--- a/Sources/swift-openapi-generator/Documentation.docc/Articles/Practicing-spec-driven-API-development.md
+++ b/Sources/swift-openapi-generator/Documentation.docc/Articles/Practicing-spec-driven-API-development.md
@@ -1,0 +1,259 @@
+# Practicing spec-driven API development
+
+Design, iterate on, and generate both client and server code from your hand-written OpenAPI document.
+
+## Overview
+
+An OpenAPI document represents a machine-readable _contract_ between a server and its clients.
+
+There are two high-level workflows of creating and using the OpenAPI document:
+
+- **Spec-driven development**:
+    - OpenAPI document is hand-written.
+    - Client and server code is generated.
+- **Code-driven development**:
+    - Server code is hand-written.
+    - OpenAPI document and client code are generated.
+
+This guide explains why spec-driven development is the recommended practice for use with Swift OpenAPI Generator.
+
+### (Recommended) Spec-driven development
+
+Starting with an API contract, before any server or client code is written, allows both the server and client teams to be involved in the API design process from the start.
+
+The goal is to draft the initial OpenAPI document, then let both server and client teams work in parallel as they write the code necessary to implement their side, gather feedback, and propose the next version. This cycle never ends, however after a stable version is released, more care needs to be taken to avoid breaking changes until the next major version (learn more about API stability in <doc:API-stability-of-generated-code>.)
+
+Having the OpenAPI document written before the server team writes any code allows the client team to spin up a mock server that follows the proposed API, and start work on the client.
+
+Unblocking client teams allows for quicker iteration than a code-driven workflow, as there is less waiting of teams for one another.
+
+```
+                                   ┌────────────────────────────┐
+                             ┌────▶│    Server Implementation   │
+                             │     └────────────────────────────┘
+                             │     ┌────────────────────────────┐
+  ┌────────────────────┐     ├────▶│         App Client         │
+  │  OpenAPI Document  │─────┤     └────────────────────────────┘
+  └────────────────────┘     │     ┌────────────────────────────┐
+                             ├────▶│         Web Client         │
+                             │     └────────────────────────────┘
+                             │     ┌────────────────────────────┐
+                             └────▶│            ...             │
+                                   └────────────────────────────┘
+```
+
+Swift OpenAPI Generator generates both the client and server stub code, helping ensure that the implementation of both sides follow the agreed-upon OpenAPI document. 
+
+Collaborating on the API definition, as opposed to code, allows inexpensive iteration on the OpenAPI document. It is "just" a YAML or JSON file, after all, made easy to browse and edit using tools available in the broader OpenAPI community, offering syntax highlighting, autocompletion, and validation in many popular text editors.
+
+### (Discouraged) Code-driven development
+
+Code-driven development, or in other words, writing server code first, and generating the OpenAPI document from it, is discouraged for use with Swift OpenAPI Generator. 
+
+While initially it can appear convenient, as server developers start writing code without having an agreed API contract first, it quickly slows down the iteration cycle as soon as clients get involed and start proposing changes.
+
+```
+                                                                  ┌────────────────────────────┐
+                                                              ┌──▶│         App Client         │
+                                                              │   └────────────────────────────┘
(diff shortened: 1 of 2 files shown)


Sources/swift-openapi-generator/Documentation.docc/Articles/Practicing-spec-driven-API-development.md
```diff
@@ -0,0 +1,259 @@
...
+
+struct Handler: APIProtocol {} // <<< this is where you now get a build error
+
+let transport = VaporTransport(routesBuilder: app)
+try handler.registerHandlers(on: transport, serverURL: ...)
+
+try app.run()
+```
+
+If you try to compile the above, you'll get a build error. Because now, the `APIProtocol` contains the requirement to implement the `getFoo` operation, but you're not implementing it, yet. Xcode will offer a Fix-it, and drop in a function stub that you just fill in:
```
heckj, line 198:
```suggestion
When you compile the example above, you'll get a build error because the `APIProtocol` contains the requirement to implement the `getFoo` operation, but it isn't yet implemented.
Xcode will offer a Fix-it, and drop in a function stub that you can fill in:
```

Sources/swift-openapi-generator/Documentation.docc/Articles/Practicing-spec-driven-API-development.md
```diff
@@ -0,0 +1,259 @@
...
+try app.run()
+```
+
+Now, build run, and test!
+
+Only one operation was moved over, but you can get confidence that it works and even deploy the service.
+
+#### Repeat for the remaining operations
+
+At this point, `POST /foo` and `GET /bar` are still manually implemented, but `GET /foo` is coming from your OpenAPI document and you only had to move the business logic over.
```
heckj, line 225:
```suggestion
At this point, `POST /foo` and `GET /bar` are still manually implemented, but `GET /foo` is coming from the OpenAPI document and you only had to move the business logic over.
```

Sources/swift-openapi-generator/Documentation.docc/Articles/Practicing-spec-driven-API-development.md
```diff
@@ -0,0 +1,259 @@
...
+
+At your convenience, repeat this for the remaining two operations, until you have no manual operations (or feel free to keep some manual operations there, for example for serving static css/js files, those endpoints usually don't go into the OpenAPI document, which is meant mainly for the REST API).
+
+#### Final state
+
+You should end up with something like this:
```
heckj, line 231:
```suggestion
The end result should be something like this:
```

heckj (commented):
@czechboy0 Feel free to ignore some of all of these feedbacks - I tried to make them easy for you to consider and implement, if you like the wording.

I tried to collect and coalesce the arguments in the early sections, and after having re-read it - I think this really might be two articles slapped together, and better separated. One that describes the possible workflows - with it's pros and cons - and the other that talks about a process to migration to a spec-driven workflow when you started with a code-driven one (or a code-based experiment you want to formalize).

I also tried to minimize the "you, your" and shift more to direct language (older author training), and tried to step on as much passive voice as I could to make the actors and actions clear in all sentences. I probably missed a bit, but that was my goal.

Thanks for encouraging me provide feedback for this. I really appreciate the flow and draft of the arguments to work with.

czechboy0:
Thanks for taking the time, @heckj, I'll work on incorporating your feedback.

This is indeed two sections, "why" use spec-driven development and "how" to move to it. I don't think they make much sense separated out, so I'll keep them in one article. Developers are expected to read them in this order.

Swift OpenAPI Generator was designed for a workflow where you have complete control over your OpenAPI document, which isn't the case for code-driven (see the Project Goals and Scope article in our docs). So developers might successfully use code-driven elsewhere, with a different tooling stack. But not with a code generator like ours, that aims to faithfully represent the OpenAPI document, rather than applying a bunch of opinionated "fixups" on it (when it comes to name formatting, and so on).

czechboy0:
Also, the active voice is intentional and is part of our docs standard, but you made many other good suggestions that I'll work through 🙂 Thanks again.

Sources/swift-openapi-generator/Documentation.docc/Articles/Practicing-spec-driven-API-development.md
```diff
@@ -0,0 +1,259 @@
...
+| Workflow | Source of truth | Published | Transcoding |
+|----------|-----------------|-----------|-------------|
+| Spec-driven | OpenAPI doc | OpenAPI doc | None required ✅ |
+| Code-driven | Code | OpenAPI doc | Can be lossy and unpredictable ❌ |
+
+Notice from the table above that the recommended spec-driven workflow publishes the same document that the developers of the server use to define the server's API (source of truth).
+
+Publishing the source of truth is preferable to relying on transcoding from code (or other representations) to an OpenAPI document, which is often lossy and difficult to predict how a given language-specific annotation affects the OpenAPI output. It also means that any feature unsupported by the transcoder cannot be represented in the generated OpenAPI document.
+
+> Tip: Publish the source of truth, not a representation transcoded from the source of truth. That way, your clients can open pull requests to your OpenAPI document without having to learn how your server is implemented, nor do they need access to the source code of your server.
```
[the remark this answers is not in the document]
czechboy0: I broke it out as you suggested, but kept the first part in a tip - visually, when rendered, it looks better to break up too many plain text paragraphs with a highlight like this.

Sources/swift-openapi-generator/Documentation.docc/Articles/Practicing-spec-driven-API-development.md
```diff
@@ -0,0 +1,259 @@
...
+
+### Migrate from code-driven to spec-driven development
```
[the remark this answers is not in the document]
czechboy0: Yeah that's also be reasonable and I've considered it, but I still believe the "why" and "how" are best kept in a single article. We can also always link to sections with fragments in URLs.

czechboy0:
I applied the vast majority of your feedback, @heckj, thanks again for providing it!

I think the article is now in a decent shape to land, but of course, PRs improving our docs are always welcome, even after 1.0 next week 🙂

heckj:
glad it was useful, thanks for the updates!

simonjbeaumont (approved):
I'm fine with this as-is, but some scope for further polish could include:

- More fine grained breakdown of the path to spec-driven development. Specifically moving from no docs -> handwritten docs -> handwritten, machine-readable docs -> generated machine-readable docs -> spec-driven development.
- Also pointing out that this doesn't _stop_ you making breaking changes, but makes it _better_ for a number of reasons: detection, explicitly documented, explicit deprecation, explicitly versioned.

If you feel these aren't suited to _this_ doc, maybe we can find another home for these points.

czechboy0:
Let's discuss offline, I do think the content belongs in our docs, let's just find a good place for it.

PR #421, opened 2023-12-04
Log the access modifier for clarity

czechboy0:
### Motivation

The access modifier is now customizable, but adopters might not be aware that we also changed the default from `public` to `package`. We should log that fact in the summary we print by the CLI, hopefully making it more discoverable during debugging.

### Modifications

Log the resolved access modifier when the CLI runs (also printed when the plugin invokes it).

### Result

Should be clearer what the chosen access modifier is and why some API might be either leaking out, or not visible.

### Test Plan

N/A, it's a log line.


simonjbeaumont (approved):
LGTM.

PR #426, opened 2023-12-06
Add an example using a Postgres database

simonjbeaumont:
### Motivation

As we approach 1.0, we want more examples of how to use this package and integrate with other packages in the ecosystem.

### Modifications

Add an example that uses PostgresNIO to persist state in a local Postgres database.
 
### Result

More examples.

### Test Plan

- Locally verified.
- CI will build the package.

### DIFF
--- a/Examples/PostgresDatabaseServer/README.md
+++ b/Examples/PostgresDatabaseServer/README.md
@@ -1,0 +1,94 @@
+# Use a database for persistent storage
+
+In this example we'll integrate persistent state into the server by adding an
+integration to a database.
+
+## Overview
+
+This example extends the [HelloWorldVaporServer](../HelloWorldVaporServer/)
+with a trivial API to return the number of messages it has received.
+
+For persistent state, it makes use of a local Postgres database to store the
+messages and uses [PostgresNIO](https://github.com/vapor/postgres-nio) for the
+database interaction.
+
+The type that provides the API handlers has been converted to an actor and
+a database connection is established in the initializer. The Postgres API
+requires a logger for all queries so a logger property has also been added.
+                
+The server uses a table called `messages` to record the messages the server
+responds with. This table is created if it does not already exist in the
+initializer.
+
+In the `getGreeting` handler, a row is inserted into the `messages` table
+containing the message before returning a response.
+
+This example also includes two other API operations: `getCount`, which returns
+the number of messages the server has provided; and `reset`, which resets the
+database.
+
+## Testing
+
+### Running a local database container
+
+We need a database running locally to use in this exercise. We'll use
+[Compose](https://docs.docker.com/compose) to run Postgres locally:
+
+```console
+% docker compose start
+...
+[+] Running 1/1
+ ⠿ Container postgresdatabaseserver-postgres-1  Started            0.4s
+```
+
+### Testing the server
+
+Run the server locally using the following command:
+
+```console
+% swift run
+```
+
+Then, in another terminal, make requests and see the database in action.
+
+```console
+% curl "localhost:8080/api/greet?name=Jane"
+{
+  "message" : "Hello, Jane!"
+}
+
+
+% curl "localhost:8080/api/greet?name=Jane"
+{
+  "message" : "Hello, Jane!"
+}
+
+
+% curl "localhost:8080/api/count"
+{
+  "count" : 2
+}
+
+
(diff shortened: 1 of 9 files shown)


Examples/PostgresDatabaseServer/README.md
```diff
@@ -0,0 +1,95 @@
...
+## Overview
+
+This example extends the [HelloWorldVaporServer](../HelloWorldVaporServer/)
+with a trivial API to return the number of messages it has received.
+
+For persistent state, it makes use of a local Postgres database to store the
+messages and uses [PostgresNIO](https://github.com/vapor/postgres-nio) for the
+database interaction.
+
+The type that provides the API handlers has been converted an actor and
```
czechboy0, line 15:
```suggestion
The type that provides the API handlers has been converted to an actor and
```

Examples/PostgresDatabaseServer/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+This example also includes two other API operations: `getCount`, which returns
+the number of messages the server has provided; and `reset`, which resets the
+database.
+
+## Testing
+
+### Running a local database container
+
+We need a database running locally to use in this exercise. We'll use Compose
```
czechboy0, line 34:
```suggestion
We need a database running locally to use in this exercise. We'll use [Compose](https://docs.docker.com/compose/)
```

Examples/PostgresDatabaseServer/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Testing the server
+
+Run the server locally using the following command:
+
+```console
+% swift run
+```
+
+Then, in another terminal, make requests and see the database in action.
```
czechboy0, line 52:
```suggestion
Then, in another Terminal window, make requests and see the database in action.
```

Examples/PostgresDatabaseServer/Sources/PostgresDatabaseServer/openapi.yaml
```diff
@@ -0,0 +1,59 @@
...
+            application/json:
+              schema:
+                type: object
+                properties:
+                  count:
+                    type: integer
+                required:
+                  - count
+  /reset:
+    get:
```
czechboy0, line 42: Should be POST/PUT?

Examples/PostgresDatabaseServer/Sources/PostgresDatabaseServer/openapi.yaml
```diff
@@ -0,0 +1,59 @@
...
+      operationId: reset
+      responses:
+        '200':
+          description: Clear out table
+          content:
+            application/json:
+              schema:
+                type: object
```
czechboy0, line 50: Can return 204 No Content instead.

czechboy0 (approved):
lgtm, a few suggestions.

Examples/PostgresDatabaseServer/README.md
simonjbeaumont: Less keen on this one.

Examples/PostgresDatabaseServer/Sources/PostgresDatabaseServer/openapi.yaml
simonjbeaumont: Ha, I had the following at some point:

> Note: There may be more appropriate HTTP method for reset, but we're just using GET for illustrative purposes.

I think it was so we didn't need to use `-X POST` with `curl`. But I changed it to `POST` and 204 return.

Examples/PostgresDatabaseServer/Sources/PostgresDatabaseServer/openapi.yaml
simonjbeaumont: Seems reasonable.

Examples/PostgresDatabaseServer/README.md
simonjbeaumont: Done

Examples/PostgresDatabaseServer/README.md
simonjbeaumont: Done.

Aside: I considered _not_ using Compose at all, but think it will be necessary for the upcoming telemetry ones.

simonjbeaumont (commented):
Applied most of the feedback, thanks!

PR #427, opened 2023-12-06
Remove unneeded target dependencies from HelloWorldVaporServerTests

simonjbeaumont:
### Motivation

The HelloWorld example had a test target that had the same target dependencies as the source target. These weren't necessary and, if these are going to be examples, they should be cleaned up to be as minimal as possible.

### Modifications

Remove unneeded dependencies from HelloWorldVaporServerTests target.

### Result

Simpler Package.swift

(And, annoyingly formatted Package.swift, due to swift-format)

### Test Plan

CI should still be fine.

### DIFF
--- a/Examples/HelloWorldVaporServer/Package.swift
+++ b/Examples/HelloWorldVaporServer/Package.swift
@@ -32,14 +32,6 @@ let package = Package(
                 .product(name: "Vapor", package: "vapor"),
             ],
             plugins: [.plugin(name: "OpenAPIGenerator", package: "swift-openapi-generator")]
-        ),
-        .testTarget(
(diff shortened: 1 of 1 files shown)


czechboy0:
I was struggling to get this working without the extra dependencies, was getting a linker issue. Let's debug this together.

simonjbeaumont:
> I was struggling to get this working without the extra dependencies, was getting a linker issue. Let's debug this together.

Sure. Odd: I see the CI failed too. But locally, this builds fine, from clean.

czechboy0:
Superseded by https://github.com/apple/swift-openapi-generator/pull/487.

PR #433, opened 2023-12-07
Add RetryingClientMiddleware example

czechboy0:
### Motivation

Adds a retrying middleware example.

### Modifications

Ditto.

### Result

Now we can point to how to do retries in a middleware, the part about potentially calling `next` multiple times might not be immediately obvious, so it's an important example to show.

### Test Plan

Tested manually by tweaking a local server to return 500 randomly and saw the retry work.


simonjbeaumont:
@czechboy0 why does this depend on the linked PR?

Examples/RetryingClientMiddleware/Package.swift
```diff
@@ -0,0 +1,43 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import PackageDescription
+
+let package = Package(
+    name: "RetryingClientMiddleware",
+    platforms: [.macOS(.v11), .iOS(.v14), .tvOS(.v14), .watchOS(.v7), .visionOS(.v1)],
```
simonjbeaumont, line 19: Any reason for the platform bump?

simonjbeaumont (approved):

Examples/RetryingClientMiddleware/Package.swift
czechboy0: Fixed, it was copied from the logging examples that required it.

czechboy0:
Passed in docker locally, landing without waiting for examples to pass.

Integration test is unaffected, this diff only changes the Examples directory.

PR #439, opened 2023-12-07
Update minimum version of OpenAPIKit to 3.1.2

simonjbeaumont:
### Motivation

This picks up a bug fix with inferred properties. While Swift package resolution should pick this up for adopters automatically, we want to go into 1.0 with the minimum version of the dependency bumped to one that includes the fix.

### Modifications

Update minimum version of OpenAPIKit to 3.1.2

### Result

Requires newer OpenAPIKit with fix.

### Test Plan

CI.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -51,7 +51,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-algorithms", from: "1.0.0"),
 
         // Read OpenAPI documents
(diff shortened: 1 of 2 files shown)


czechboy0 (approved):

czechboy0:
Landing this to unblock other PRs. Compat suite passed, which is the only thing changed here.

PR #441, opened 2023-12-08
Default the access modifier to internal

czechboy0:
### Motivation

After further consideration, `internal` seems the be the right default for access modifier of the generated, mirroring Swift itself.

### Modifications

Switch the default to `internal` and update docs.

### Result

`accessModifier: internal` is the default now.

### Test Plan

Updated tests.


simonjbeaumont (changes requested):
Integration test relied on `public` or `package`. Probably needs making `package` explicitly.

simonjbeaumont (approved):
LGTM.

PR #442, opened 2023-12-08
Delete example temp data on CI

czechboy0:
### Motivation

We're not properly cleaning up temporary files in our scripts for examples, leading to CI running out of disk space.

### Modifications

Delete each example copy, and the scratch/cache space at the end of the script.

### Result

Hopefully will recover CI.

### Test Plan

N/A, PR CI will show if this fixes things.

### DIFF
--- a/scripts/test-examples.sh
+++ b/scripts/test-examples.sh
@@ -74,4 +74,11 @@ for EXAMPLE_PACKAGE_PATH in $(find "${EXAMPLES_PACKAGE_PATH}" -maxdepth 2 -name 
         --scratch-path "${SHARED_SCRATCH_PATH}" \
         --cache-path "${SHARED_CACHE_PATH}" \
         unedit swift-openapi-generator
+
+    log "Deleting example ${EXAMPLE_PACKAGE_NAME} at ${EXAMPLE_COPY_DIR}"
+    rm -rf "${EXAMPLE_COPY_DIR}"
 done
+
+log "Deleting cache directories"
+rm -rf "${SHARED_SCRATCH_PATH}"
+rm -rf "${SHARED_CACHE_PATH}"


scripts/test-examples.sh
```diff
@@ -74,4 +74,11 @@ for EXAMPLE_PACKAGE_PATH in $(find "${EXAMPLES_PACKAGE_PATH}" -maxdepth 2 -name
...
         --cache-path "${SHARED_CACHE_PATH}" \
         unedit swift-openapi-generator
+
+    log "Deleting example ${EXAMPLE_PACKAGE_NAME} at ${EXAMPLE_COPY_DIR}"
+    rm -rf "${EXAMPLE_COPY_DIR}"
 done
+
+log "Deleting cache directories"
+rm -rf "${SHARED_SCRATCH_PATH}"
+rm -rf "${SHARED_CACHE_PATH}"
```
simonjbeaumont, line 84:
```suggestion
log "Deleting temporary directory"
rm -rf "${TMP_DIR}"
```

czechboy0: Ah cool, I didn't realize we own `$TMP_DIR` here and it's not a shared one.

simonjbeaumont:
OK, let's merge this over the compat-test failure, which is fixed by #439 but that's blocked behind this.

scripts/test-examples.sh
```diff
@@ -47,31 +45,24 @@ for EXAMPLE_PACKAGE_PATH in $(find "${EXAMPLES_PACKAGE_PATH}" -maxdepth 2 -name
...
-        --package-path "${EXAMPLE_COPY_DIR}" \
-        --scratch-path "${SHARED_SCRATCH_PATH}" \
-        --cache-path "${SHARED_CACHE_PATH}" \
-        unedit swift-openapi-generator
+    log "Deleting example ${EXAMPLE_PACKAGE_NAME} at ${EXAMPLE_COPY_DIR}"
+    rm -rf "${EXAMPLE_COPY_DIR}"
 done
+
+log "Deleting temporary directory"
+rm -rf "${TMP_DIR}"
```
simonjbeaumont, line 68: I don't think this is necessary any more as we can see in the logs that the pipeline is run with `--rm`
```suggestion
```

simonjbeaumont (changes requested):

scripts/test-examples.sh
czechboy0: Whoops will fix in the next PR

PR #458, opened 2023-12-08
Use conventional naming for example projects

simonjbeaumont:
### Motivation

The example packages were named using `CamelCase`, rather than `kebab-case` and also could use with some renaming to group them better and make it clear they are examples.

### Modifications

- Rename all the example Swift packages to use `kebab-case`.
- Include `-example` suffix in all example package names.
- Rename the Xcode project too to align with the Hello World examples.

### Result

Easier to browse the examples.

### Test Plan

CI.


Examples/curated-client-library-example/README.md
```diff
@@ -1,4 +1,4 @@
-# Curated Library Client
+# Curated library client module
```
czechboy0, line 1: Have we used the word "module" in any other example?

Examples/manual-generation-package-plugin-example/README.md
```diff
@@ -1,4 +1,4 @@
-# A command plugin invocation Client
+# Manual code generation using the Swift package plugin
```
czechboy0, line 1:
```suggestion
# Manual code generation using the Swift package command plugin
```

Examples/swagger-ui-endpoint-example/README.md
```diff
@@ -1,4 +1,4 @@
-# Swagger UI Endpoints Server
+# Server with Swagger UI Endpoint
```
czechboy0, line 1:
```suggestion
# Server with Swagger UI Endpoints
```

There are 3 endpoints, right? The raw file and the HTML page, plus the redirect.

czechboy0 (approved):
Just a few nits, otherwise looks great! We can update the Readme separately.

Examples/swagger-ui-endpoint-example/README.md
simonjbeaumont: Went with "Server with OpenAPI document and Swagger UI endpoints" for now.

Examples/manual-generation-package-plugin-example/README.md
simonjbeaumont: If it's using a manual workflow with a package plugin, it implies command plugin, right?

Examples/curated-client-library-example/README.md
simonjbeaumont: Updated to "Curated client library" for now.

simonjbeaumont:
@czechboy0 OK, I updated the parent README and applied most of your suggestions. Given you approved already, I'll land this, and we can always follow up with more refinements before release.

PR #463, opened 2023-12-10
Update NOTICE.txt with projects used in examples

simonjbeaumont:
### Motivation

We've recently added a bunch of example projects that show how to integrate with other packages and projects in the Swift and OpenAPI ecosystems. We should make sure that these projects and their associated licenses are mentioned in `NOTICE.txt`.

### Modifications

Update NOTICE.txt with projects used in examples.

### Result

Updated NOTICE.txt with projects used in examples.

### Test Plan

None.

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -120,10 +120,71 @@ This product contains OpenAPI document(s) for compatibility testing from:
 
 -------------------------------------------------------------------------------
 
-This product contains an example Swift package that shows how to use swagger-ui.
+This product contains samples of how to use Swift OpenAPI Generator with the
+following other projects:
 
-  * swagger-ui
+  * AsyncHTTPClient
+    * LICENSE (Apache License 2.0):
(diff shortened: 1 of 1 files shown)


NOTICE.txt
```diff
@@ -120,10 +120,71 @@ This product contains OpenAPI document(s) for compatibility testing from:
...
+    * HOMEPAGE:
+      * https://github.com/apple/swift-log
+
+  * Swift Metrics
+    * LICENSE (Apache License 2.0):
+      * https://www.apache.org/licenses/LICENSE-2.0
+    * HOMEPAGE:
+      * https://github.com/apple/swift-metrics
+
+  * SwiftNIO
```
simonjbeaumont, line 168: The projects that are styled as `SwiftThing` (cf. `Swift Thing`) appear here like this because this is how they are styled in their own project's license headers.

czechboy0 (approved):

PR #471, opened 2023-12-11
Validate content type strings in validateDoc

### LINKED ISSUE #342: Validate content-type strings in validateDoc
Validate content-type strings in validateDoc, similar to https://github.com/apple/swift-openapi-generator/issues/341, to catch invalid inputs at the very start.

Would mean trying to parse the content type of all places that can have content in the doc, and throwing a descriptive error if it cannot be parsed correctly. (This means a malformed string, like empty, one without `foo/bar` two components, etc.)

There are many docs with malformed content types and it'd be good to emit good errors early in the process to allow the adopter to fix it up.

PARAIPAN9:
### Motivation

- Fixes #342.

### Modifications

- Create extracting and validation functions for content type strings.

### Result

- This will help catch invalid content types before code generation.

### Test Plan

- Test it on an openapi document to make sure the errors are thrown as intended and wrote additional unit tests.


simonjbeaumont:

PARAIPAN9:
You're very welcome! Thank you for considering the changes, and I completely understand the priority of shipping 1.0.0 first. I'm thrilled to contribute, and eagerly looking forward to the exciting milestone of the first release.

czechboy0:
Hi @PARAIPAN9 - the issue might not have been super clear, but the issue was about doing the work of walking the OpenAPI document, and validating all the content type strings that exist there, in the `validateDoc` function.

That way, we'd catch invalid content types at the start, instead of doing so later when we're trying to generate code.

The errors emitted today, when an invalid content type is encountered, are good enough, we just don't do the validation at the start, so it can lead to a more confusing debugging experience.

Does that make sense? If that task is not something you want to pick up, no worries at all, just let me know and I'd unassign you. 🙂

czechboy0 (changes requested):
See comment above, the task is slightly different.

PARAIPAN9:
Ah, thank you for the clarification. It's ok, I'll continue working on it.

Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
```diff
@@ -13,6 +13,58 @@
...
+
+    return contentTypes
+}
+
+/// Validates an array of content types.
+///
+/// - Parameter contentTypes: An array of strings representing content types.
+/// - Throws: A Diagnostic error if any content type is invalid.
+func validateContentTypes(_ contentTypes: [String]) throws {
+    let mediaTypePattern = "^[a-zA-Z]+/[a-zA-Z][a-zA-Z-]*$"
```
czechboy0, line 51: To avoid duplicating similar logic, you can validate whether the content type is valid by trying to create a `_OpenAPIGeneratorCore.ContentType` value from the string, and if it fails, it's an invalid content type. This regex doesn't fully represent the requirements, and we don't want to have to keep multiple pieces of similar logic in sync manually.

Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
```diff
@@ -13,6 +13,58 @@
 //===----------------------------------------------------------------------===//
 
 import OpenAPIKit
+import Foundation
+
+/// Extracts content types from a ParsedOpenAPIRepresentation.
+///
+/// - Parameter doc: The OpenAPI document representation.
+/// - Returns: An array of strings representing content types extracted from requests and responses.
+func extractContentTypes(from doc: ParsedOpenAPIRepresentation) -> [String] {
```
czechboy0, line 22: You could change this to a visitor pattern, where you provide a closure with the signature `(String) -> Bool`, that returns `true` if the content type is valid. And this method would be responsible for walking all the places that have content types.

That way, you can then emit a good diagnostic about _which_ content type was invalid, and _where_ it was found. That's important for debugging.

Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
```diff
@@ -13,6 +13,58 @@
...
+            .flatMap { (response: [Either<OpenAPI.Reference<OpenAPI.Response>, OpenAPI.Response>]) in
+                response.compactMap { (eitherResponse: Either<OpenAPI.Reference<OpenAPI.Response>, OpenAPI.Response>) in
+                    guard case .b(let actualResponse) = eitherResponse else { return nil }
+                    return actualResponse.content.keys.first?.rawValue
+                }
+            }
+
+        return requestBodyContentTypes + responseContentTypes
+    }
+
```
czechboy0, line 42: We should also validate the content types in `#/components/requestBodies` and `#/components/responses` here.

czechboy0 (changes requested):
Thanks - just a few more comments about how we can validate all the content types and emit more information if an invalid content type is found.

Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
```diff
@@ -14,6 +14,71 @@
...
+        guard case .b(let pathItem) = pathValue else { continue }
+        for endpoint in pathItem.endpoints {
+
+            if let eitherRequest = endpoint.operation.requestBody {
+                if case .b(let actualRequest) = eitherRequest {
+                    for contentType in actualRequest.content.keys {
+                        if !validate(contentType.rawValue) {
+                            throw Diagnostic.error(
+                                message:
+                                    "Invalid content type string: '\(contentType.rawValue)' found in requestBody at path '\(path.rawValue)'. Must have 2 components separated by a slash '<type>/<subtype>'.\n"
```
czechboy0, line 37:
```suggestion
                                    "Invalid content type string: '\(contentType.rawValue)' found in requestBody at path '\(path.rawValue), method: \(endpoint.method.rawValue)'. Must have 2 components separated by a slash '<type>/<subtype>'.\n"
```

Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
```diff
@@ -14,6 +14,71 @@
...
+                }
+            }
+
+            for eitherResponse in endpoint.operation.responses.values {
+                if case .b(let actualResponse) = eitherResponse {
+                    for contentType in actualResponse.content.keys {
+                        if !validate(contentType.rawValue) {
+                            throw Diagnostic.error(
+                                message:
+                                    "Invalid content type string: '\(contentType.rawValue)' found in responses at path '\(path.rawValue)'. Must have 2 components separated by a slash '<type>/<subtype>'.\n"
```
czechboy0, line 50:
```suggestion
                                    "Invalid content type string: '\(contentType.rawValue)' found in responses at path '\(path.rawValue), method: \(endpoint.method.rawValue)'. Must have 2 components separated by a slash '<type>/<subtype>'.\n"
```

Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
```diff
@@ -14,6 +14,71 @@
...
+                            )
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    for component in doc.components.requestBodies.values {
```
czechboy0, line 59: Let's iterate both keys and values, to allow us to print the key later down.
```suggestion
    for (key, component) in doc.components.requestBodies {
```

Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
```diff
@@ -14,6 +14,71 @@
...
+            if !validate(contentType.rawValue) {
+                throw Diagnostic.error(
+                    message:
+                        "Invalid content type string: '\(contentType.rawValue)' found in #/components/requestBodies. Must have 2 components separated by a slash '<type>/<subtype>'.\n"
+                )
+            }
+        }
+    }
+
+    for component in doc.components.responses.values {
```
czechboy0, line 70:
```suggestion
    for (key, component) in doc.components.responses {
```

Same here. And please include the key in the diagnostic.

Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
```diff
@@ -14,6 +14,71 @@
...
+        for contentType in component.content.keys {
+            if !validate(contentType.rawValue) {
+                throw Diagnostic.error(
+                    message:
+                        "Invalid content type string: '\(contentType.rawValue)' found in #/components/responses. Must have 2 components separated by a slash '<type>/<subtype>'.\n"
```
czechboy0, line 75:
```suggestion
                    message:
                        "Invalid content type string."
                    context: ["contentType": contentType.rawValue, "location": "#/components/responses/\(key.rawValue)", "recoverySuggestion": "Must have 2 components separated by a slash '<type>/<subtype>'."]
```

Please keep the message a static string, so that we can group errors easily, and put the dynamic data into the context dictionary, for example like this. Please do this for the rest of the diagnostics you added.

czechboy0 (changes requested):
Great, we're basically there - just added a few more suggestions, but this is the right shape!

czechboy0 (approved):
Great, thank you! Let's just get the CI green and I'll merge it.

czechboy0:

PARAIPAN9:
Perfect, You’re welcome!!

PR #481, opened 2023-12-12
Use a shared cache when testing examples, but better

simonjbeaumont:
### Motivation

The examples pipeline currently takes around one hour to run. The CI timeout is 60 minutes so we sometimes don't even finish it before the CI job gets cancelled. We previously experimented with enabling a shared build cache but this wasn't working as desired because the paths to the source files were different for each of the examples we build. However, by using the same package harness path, we should be able to get better build cache performance (and keep the cache smaller during the run).

### Modifications

- Use a shared `--cache-path` and `--scratch-path`.
- Use a stable `--package-path` for each example (we delete it entirely and recreate to be sure it's clean).

I experimented briefly with `ccache` which I thought might help with some of the builds (e.g. the BoringSSL build, which comes in a few times as a dependency) but didn't get very far.

### Result

CI pipeline should run much quicker.

It's possible that we will not catch failures where we depend on something that we haven't declared properly, because these object files will still be in the cache. I can't say with confidence that this is how the build system works though; potentially it's a non-issue.

### Test Plan

CI.

### DIFF
--- a/scripts/test-examples.sh
+++ b/scripts/test-examples.sh
@@ -27,42 +27,60 @@ TMP_DIR=$(/usr/bin/mktemp -d -p "${TMPDIR-/tmp}" "$(basename "$0").XXXXXXXXXX")
 
 PACKAGE_PATH=${PACKAGE_PATH:-${REPO_ROOT}}
 EXAMPLES_PACKAGE_PATH="${PACKAGE_PATH}/Examples"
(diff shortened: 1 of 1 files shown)


simonjbeaumont:
@czechboy0 OK, this seems to get it down to 21 minutes.

czechboy0 (approved):

Issue #299, opened 2023-09-26
Improve access to undocumented HTTP data


groue:
Hello,

OpenAPIRuntime v0.2.5 comes with an [`UndocumentedPayload`](https://swiftpackageindex.com/apple/swift-openapi-runtime/0.2.5/documentation/openapiruntime/undocumentedpayload) type that is returned for undocumented status codes.

This is all and good, but this type is empty and does not provide any information about the actual response returned by the server. This means that the client is 100% unable to process the response (and maybe look for information present in the raw body response, who knows).

The [tutorial](https://swiftpackageindex.com/apple/swift-openapi-generator/0.2.3/tutorials/swift-openapi-generator/clientswiftpm) gives this example:

```swift
let client = Client(...)
let response = try await client.getGreeting(...)

switch response {
case .ok(let okResponse):
    print(okResponse)
case .undocumented(statusCode: let statusCode, _):
    // ALL INFORMATION IS LOST HERE (but the status code)
    print("🥺 undocumented response: \(statusCode)")
}
```

I don't find this very convenient:

- Server errors (5xx) are uneasy to specify - after all the server has crashed. All bets are off. I wish I could read the raw response (body and headers), but I can not.
- Some pass-through endpoints are also uneasy to specify (4xx wrapper errors, for example). I also wish I could read the raw response (body and headers), but I can not.

Sure, 100% accurate and complete openapi.yaml files are ideal. But what about people who did not reach 100% yet, or are just developing or prototyping a server?

Is there another technique I should know about, that would allow the generated code to expose raw information? Maybe some technique at the level of `openapi.yaml`?

czechboy0:
Hi @groue 🙂 

It really depends on what you'd like to do with the undocumented payload.

If you just want to print it for debugging, I'd recommend implementing a custom `ClientMiddleware` that logs the response information, including the headers and body however you need. There, you can decide whether/how to decode the payload.

As an alternative, if you know the payload will be JSON, in the OpenAPI document, add a `default` response like this:

```yaml
responses:
  '200':
    ... successful response
  default:
    content:
      application/json: {}
```

That way, a `default` case will be generated in the response enum that holds a freeform parsed JSON object for you, allowing you to inspect it.

---

That all said, I'd only recommend using this as a temporary workaround, and document the payloads you want to extract in the OpenAPI document.

Hope this helps!

czechboy0:
[...]

PR #488, opened 2023-12-14
[Generator] Add headerFields and body to UndocumentedPayload

czechboy0:
### Motivation

Fixes https://github.com/apple/swift-openapi-generator/issues/299.

Depends on https://github.com/apple/swift-openapi-runtime/pull/90 landing first and getting released, and the version dependency being bumped here.

### Modifications

Adapted to the runtime changes, added `headerFields` and `body` properties to `UndocumentedPayload`.

Now, the generated code actually also forwards the values to it.

### Result

Easier access to the raw response data if an undocumented response is returned.

### Test Plan

Adapted snippet and file-based reference tests, and also the unit tests of the generated code (PetstoreConsumerTests).

### DIFF
--- a/Tests/PetstoreConsumerTests/Test_Client.swift
+++ b/Tests/PetstoreConsumerTests/Test_Client.swift
@@ -613,13 +613,16 @@ final class Test_Client: XCTestCase {
     }
 
     func testProbe_undocumented() async throws {
-        transport = .init { request, requestBody, baseURL, operationID in (.init(status: .serviceUnavailable), nil) }
+        transport = .init { request, requestBody, baseURL, operationID in (.init(status: .serviceUnavailable), "oh no")
+        }
         let response = try await client.probe(.init())
-        guard case let .undocumented(statusCode, _) = response else {
+        guard case let .undocumented(statusCode, payload) = response else {
             XCTFail("Unexpected response: \(response)")
             return
         }
         XCTAssertEqual(statusCode, 503)
+        XCTAssertEqual(payload.headerFields, [:])
+        try await XCTAssertEqualStringifiedData(payload.body, "oh no")
     }
 
     func testUploadAvatarForPet_200() async throws {
--- a/Sources/_OpenAPIGeneratorCore/Translator/ClientTranslator/translateClientMethod.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/ClientTranslator/translateClientMethod.swift
@@ -112,7 +112,17 @@ extension ClientFileTranslator {
                         .init(
                             label: "statusCode",
                             expression: .identifierPattern("response").dot("status").dot("code")
-                        ), .init(label: nil, expression: .dot("init").call([])),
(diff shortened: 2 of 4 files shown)


Tests/PetstoreConsumerTests/Test_Client.swift
```diff
@@ -613,13 +613,16 @@ final class Test_Client: XCTestCase {
...
+        }
         let response = try await client.probe(.init())
-        guard case let .undocumented(statusCode, _) = response else {
+        guard case let .undocumented(statusCode, payload) = response else {
             XCTFail("Unexpected response: \(response)")
             return
         }
         XCTAssertEqual(statusCode, 503)
+        XCTAssertEqual(payload.headerFields, [:])
+        try await XCTAssertEqualStringifiedData(payload.body, "oh no")
```
groue, line 625: 🤣❤️

groue:
Thank you @czechboy0, this is a much welcome addition :-)

I guess it is the change to `Translator/ClientTranslator/translateClientMethod.swift` that fills the undocumented payload with the response headers and body, although I don't quite understand how it works 😅

I'm not sure I understand what's in the response body of the undocumented payload, though. But that's only because I don't understand exactly when `.undocumented` is returned.

I assume that the response body was not fully buffered, since this would create a risk of high memory consumption.

I thus assume that not a single byte of response body was consumed, otherwise it would throw `TooManyIterationsError` when touched, and the undocumented payload would be unusable.

Hence the undocumented payload is only returned based on the headers of the response, and this will never change in the future. All unexpected data in the response body itself will never return `.undocumented`, but some kind of decoding error.

Do I understand well? If so, I'm very happy with this PR!

czechboy0:
> I don't understand exactly when `.undocumented` is returned.

For example, if you OpenAPI document defines responses for 200, 400, and 500, and you receive 401, you'll get an  `.undocumented` case, with the associated status value saying `401`. And now, the payload will contain the header fields and the body stream.

> I assume that the response body was not fully buffered, since this would create a risk of high memory consumption.

Correct, `HTTPBody` is an async sequence of byte chunks, and it's up to the ultimate consumer of this data to start pulling from there. So this change still doesn't lead to any buffering.

> Hence the undocumented payload is only returned based on the headers of the response, and this will never change in the future. All unexpected data in the response body itself will never return `.undocumented`, but some kind of decoding error, and this is a strong contract of the library.

Yes, exactly. The response HTTP status code is how we decide the handling. If the status code is undocumented, you get the raw data. If it's documented, we unwrap the way that the OpenAPI document tells us. If you receive a documented response, but with body data that don't fit the documented structure, an error will be thrown.

> Do I understand well? If so, I'm very happy with this PR!

I believe so - please just confirm again that the above fits well with your understanding 🙂

groue:
Perfect! This was a quick win, enabled by a good initial design (the smoothest butter developers can dream of) 🙂👍

simonjbeaumont (approved):

PR #489, opened 2023-12-14
examples: Flesh out Makefile-based example

simonjbeaumont:
### Motivation

We have an example project showing ahead-of-time generation using the `swift-openapi-generator` CLI directly, driven by a Makefile. This PR polishes it up a bit and makes it more flexible. Additionally, the generated code had not been updated since we switched the default access modifier from package to internal.

### Modifications

- Allow overriding a bunch of things when calling `make`.
- Prompt for confirmation before blowing away the output directory.
- Regenerate sources to pick up the new access modifier.
- Add a `help` target.

### Result

More extensive example.

E.g. running the following command:

```console
% make help
Run make with one of the following targets:

  help:          Display this help.
  generate:      Generate the sources using swift-openapi-generator.
  build:         Runs swift build.
  clean:         Delete the output directory used for generated sources.
  clean-all:     Clean everything, including the checkout of swift-openapi-generator.
  dump:          Dump all derived values used by the Makefile.

The following options can be overriden on the command line:

  SWIFT_OPENAPI_GENERATOR_GIT_URL (e.g. https://github.com/apple/swift-openapi-generator)
  SWIFT_OPENAPI_GENERATOR_GIT_TAG (e.g. 1.0.0)
  SWIFT_OPENAPI_GENERATOR_CLONE_DIR (e.g. .swift-openapi-generator)
  SWIFT_OPENAPI_GENERATOR_BUILD_CONFIGURATION (e.g. release)
  OPENAPI_YAML_PATH (e.g. openapi.yaml)
  OPENAPI_GENERATOR_CONFIG_PATH (e.g. openapi-generator-config.yaml)
  OUTPUT_DIRECTORY (e.g. Sources/ManualGeneratorInvocationClient/Generated)
``` 

### Test Plan

CI should still build the package.

### DIFF
--- a/Examples/manual-generation-generator-cli-example/Makefile
+++ b/Examples/manual-generation-generator-cli-example/Makefile
@@ -1,21 +1,86 @@
-REPO_URL = https://github.com/apple/swift-openapi-generator
-VERSION = 1.0.0
-TMP_DIR = /tmp
-REPO_DIR = $(TMP_DIR)/swift-openapi-generator
-CLI_EXECUTABLE = $(REPO_DIR)/.build/release/swift-openapi-generator
+# To see how to drive this makefile use:
+#
+#   % make help
 
-$(REPO_DIR):
-	git clone --branch $(VERSION) --depth 1 $(REPO_URL) $(REPO_DIR)
+# The following values can be changed here, or passed on the command line.
+SWIFT_OPENAPI_GENERATOR_GIT_URL ?= https://github.com/apple/swift-openapi-generator
+SWIFT_OPENAPI_GENERATOR_GIT_TAG ?= 1.0.0
+SWIFT_OPENAPI_GENERATOR_CLONE_DIR ?= $(CURRENT_MAKEFILE_DIR)/.swift-openapi-generator
+SWIFT_OPENAPI_GENERATOR_BUILD_CONFIGURATION ?= debug
+OPENAPI_YAML_PATH ?= $(CURRENT_MAKEFILE_DIR)/openapi.yaml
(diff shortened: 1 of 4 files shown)


Examples/manual-generation-generator-cli-example/Makefile
```diff
@@ -1,21 +1,86 @@
...
+#
+#   % make help
 
-$(REPO_DIR):
-	git clone --branch $(VERSION) --depth 1 $(REPO_URL) $(REPO_DIR)
+# The following values can be changed here, or passed on the command line.
+SWIFT_OPENAPI_GENERATOR_GIT_URL ?= https://github.com/apple/swift-openapi-generator
+SWIFT_OPENAPI_GENERATOR_GIT_TAG ?= 1.0.0
+SWIFT_OPENAPI_GENERATOR_CLONE_DIR ?= $(CURRENT_MAKEFILE_DIR)/.swift-openapi-generator
+SWIFT_OPENAPI_GENERATOR_BUILD_CONFIGURATION ?= debug
```
czechboy0, line 9: The default was release before, should we retain that?

czechboy0 (approved):
Just one suggestion

Examples/manual-generation-generator-cli-example/Makefile
simonjbeaumont: I don't think so. The trade off is for the user to pick. I tried with both. For small documents, the release build time dominates. On my machine it's 20s for debug, 44s for release. Generation time is 1s in both cases.

czechboy0: Ok, either is fine with me, just wanted to point out that this PR changes the default.

simonjbeaumont: Gotcha; yes, you're right.

PR #492, opened 2023-12-15
Generate arrays with an optional element type for arrays with nullable items

### LINKED ISSUE #444: Support generating nullable array elements
Error: 
```
Client error - cause description: 'Unknown', underlying error: DecodingError: valueNotFound Double
```
```
Cannot get unkeyed decoding container -- found null value instead (underlying error: <nil>)
```

openapi.yaml
```
openapi: 3.0.3
info:
  title: MyService
  version: 1.0.0
servers:
  - url: 'https://api.myservice.com'
    description: myservice
paths:
  '/data/{id}':
    get:
      operationId: data
      parameters:
        - name: id
          required: true
          in: path
          description: The id.
          schema:
            type: string
      responses:
        '200':
          description: A success response with a MyData.
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/MyData'
components:
  schemas:
    MyData:
      type: object
      required: []
      properties:
        points:
          type: object
          required: []
          properties:
            result:
              type: array
              items:
                type: object
                required: []
                properties:
                  one:
                    type: array
                    items:
                      type: number
                      nullable: true
```
```
% sw_vers                                                       
ProductName:        macOS
ProductVersion:     14.1.2
BuildVersion:       23B92
```
My schema requires that the type of array's items (numbers) need to be nullable (or `nil`).

simonjbeaumont:
### Motivation

Array items can be nullable. When this is the case, the element type of the generated array should be optional. Today it is not. For example, consider the following OpenAPI snippet:

```yaml
  StringArrayNullableItems:
    type: array
    items:
      type: [string, null]
```

This currently generates a type alias for `[String]` but it should be `[String?]`

### Modifications

Update the translator to take into account `arrayContext.items.nullable` when generating the types for array values.

### Result

Arrays with nullable items 

### Test Plan

- Snippet test for standalone schema.
- Snippet test for use within an object.
- A test that shows lossless conversion to JSON with Foundation.

### Related issues

- Fixes #444.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeMatcher.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeMatcher.swift
@@ -44,7 +44,9 @@ struct TypeMatcher {
         Self._tryMatchRecursive(
             for: schema,
             test: { schema in Self._tryMatchBuiltinNonRecursive(for: schema) },
-            matchedArrayHandler: { elementType in elementType.asArray },
+            matchedArrayHandler: { elementType, nullableItems in
+                nullableItems ? elementType.asOptional.asArray : elementType.asArray
+            },
             genericArrayHandler: { TypeName.arrayContainer.asUsage }
         )
     }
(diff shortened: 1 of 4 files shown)


Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateArray.swift
```diff
@@ -48,6 +48,7 @@ extension TypesFileTranslator {
                 let nestedDecls = try translateSchema(typeName: elementType.typeName, schema: items, overrides: .none)
                 inline.append(contentsOf: nestedDecls)
             }
+            if items.nullable { elementType = elementType.asOptional }
```
czechboy0, line 51: Please remove this one.

czechboy0 (changes requested):
Just the one change can be undone, otherwise looks great

czechboy0 (approved):

PR #500, opened 2024-01-04
Validate references in validateDoc

### LINKED ISSUE #341: Validate references in validateDoc
Validate references in validateDoc to reject unsupported/dead references at the very start, to produce a good error.

Would likely mean walking all references and checking that they're internal and can be looked up in components. Throw a descriptive error if one is found that fails that.

Apply this _after_ filtering, to avoid throwing an error about a part of the document that the adopter doesn't want to generate anyway.

PARAIPAN9:
### Motivation

- Fixes #341.
 
### Modifications

- Develop a function responsible for iterating through and validating the references within the OpenAPI document. If external references are present or if they are not found in the 'components' section, the function should throw errors.

### Result

- This will help catch invalid references before code generation.

### Test Plan

- Make sure that tests pass and wrote additional unit tests.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
+++ b/Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
@@ -259,6 +259,11 @@ func validateReferences(in doc: ParsedOpenAPIRepresentation) throws {
 /// - Returns: An array of diagnostic messages representing validation warnings.
 /// - Throws: An error if a fatal issue is found.
 func validateDoc(_ doc: ParsedOpenAPIRepresentation, config: Config) throws -> [Diagnostic] {
+    try validateReferences(in: doc)
+    try validateContentTypes(in: doc) { contentType in
+        (try? _OpenAPIGeneratorCore.ContentType(string: contentType)) != nil
+    }
+
     // Run OpenAPIKit's built-in validation.
     // Pass `false` to `strict`, however, because we don't
     // want to turn schema loading warnings into errors.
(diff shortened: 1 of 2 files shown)


Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
```diff
@@ -115,6 +272,9 @@ func validateDoc(_ doc: ParsedOpenAPIRepresentation, config: Config) throws -> [
         (try? _OpenAPIGeneratorCore.ContentType(string: contentType)) != nil
     }
 
+    let filteredDoc = try FilteredDocumentBuilder(document: doc).filter()
```
PARAIPAN9, line 275: The issue mentioned that the documentation should be filtered before validation. I'm not entirely sure if this is the intended approach.

simonjbeaumont: This filter doesn't do the right thing as the filter hasn't been configured to include anything.

However, it isn't necessary to do it here anyway because this function, `validateDoc(_:config:)`, is called as a post-transition hook after parsing the OpenAPI document _after_ the existing post-transition hook for filtering the doc[^1].

[^1]: https://github.com/apple/swift-openapi-generator/blob/85fec92ea3/Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift#L120

simonjbeaumont (changes requested):

simonjbeaumont:

simonjbeaumont (approved):

simonjbeaumont:
Thanks @PARAIPAN9!

PARAIPAN9:
You're welcome, always a pleasure to contribute!!

czechboy0 (approved):

PR #506, opened 2024-01-16
Fix compiler warnings in example projects

simonjbeaumont:
### Motivation

A recent example project was added that had a compiler warning which wasn't caught by CI. We can have our example code free of warnings and ensure they remain so by compiling them with `-warnings-as-errors` in CI.

### Modifications

- Fix compiler warnings in example.
- Build with `-warnings-as-errors` in examples CI pipeline.

### Result

Example projects compile without warnings.

### Test Plan

CI updated.

### DIFF
--- a/Examples/event-streams-server-example/Sources/EventStreamsServer/GreetingStream.swift
+++ b/Examples/event-streams-server-example/Sources/EventStreamsServer/GreetingStream.swift
@@ -25,7 +25,7 @@ final class StreamStorage: @unchecked Sendable {
     private func finishedStream(id: String) {
         lock.lock()
         defer { lock.unlock() }
-        guard let task = locked_streams[id] else { return }
+        guard locked_streams[id] != nil else { return }
         locked_streams.removeValue(forKey: id)
         print("Finished stream \(id)")
     }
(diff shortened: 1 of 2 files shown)


simonjbeaumont:
Oh, this is annoying:

```
[16/19] Compiling HelloWorldURLSessionClient Client.swift
/tmp/test-examples.sh.mM8MYEM7LH/swift-openapi-example-cache/x86_64-unknown-linux-gnu/debug/CBacktrace.build/module.modulemap:2:14: error: umbrella directory '/tmp/test-examples.sh.mM8MYEM7LH/swift-openapi-example-cache/checkouts/swift-backtrace/Sources/CBacktrace/include' not found
    umbrella "/tmp/test-examples.sh.mM8MYEM7LH/swift-openapi-example-cache/checkouts/swift-backtrace/Sources/CBacktrace/include"
```

This is leftover because we're using the shared cache/build dirs, which produces a warning. I guess we cannot enable `-warnings-as-errors` in CI then.

Will update PR to just fix the code to remove the compiler warning instead.

czechboy0 (approved):

PR #509, opened 2024-01-19
Update Project-scope-and-goals.md

czechboy0:
### Motivation

One of the examples no longer applies, we now do support customizing the access modifier.

### Modifications

Updated the docs.

### Result

More accurate docs.

### Test Plan

N/A

### DIFF
--- a/Sources/swift-openapi-generator/Documentation.docc/Articles/Project-scope-and-goals.md
+++ b/Sources/swift-openapi-generator/Documentation.docc/Articles/Project-scope-and-goals.md
@@ -38,9 +38,7 @@ Another example is the generation of empty structs within the input or output ty
 
 Some generators offer lots of options that affect the code generation process. In order to keep the project streamlined and maintainable, Swift OpenAPI Generator offers very few options.
 
-One concrete example of this is that users cannot configure the access modifier for generated code, nor does the generated code make any provision for namespace collisions in the target into which it is generated.
-
-Instead, users are advised to generate code into its own target, and use Swift's module system to separate the generated code from code that uses it.
+One concrete example of this is the lack of ability to configure the names of generated types, such as `APIProtocol`, or to customize how Swift names are computed from strings provided in the OpenAPI document.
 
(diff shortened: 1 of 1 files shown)


Sources/swift-openapi-generator/Documentation.docc/Articles/Project-scope-and-goals.md
```diff
@@ -38,9 +38,7 @@ Another example is the generation of empty structs within the input or output ty
 
 Some generators offer lots of options that affect the code generation process. In order to keep the project streamlined and maintainable, Swift OpenAPI Generator offers very few options.
 
-One concrete example of this is that users cannot configure the access modifier for generated code, nor does the generated code make any provision for namespace collisions in the target into which it is generated.
-
-Instead, users are advised to generate code into its own target, and use Swift's module system to separate the generated code from code that uses it.
+One concrete example of this is the lack of ability to configure the names of generated types, such as `APIProtocol`, or to customize how Swift names are computed from strings provided in the OpenAPI document.
```
simonjbeaumont, line 41: Although we _do_ now allow configuration of the access modifier, one salient point above, that has been deleted in this patch, is still very relevant: we do not make any provision for namespace collisions in the target into which the code is generated. The advise of using a dedicated Swift module still applies, no?

czechboy0: Yes - can you provide a suggestion I could just apply? 🙏

simonjbeaumont, line 41:
```suggestion
One concrete example of this is that users cannot configure the names of generated types, such as `Client` and `APIProtocol`, and there is no attempt to prevent namespace collisions in the target into which it is generated.

Instead, users are advised to generate code into a dedicated target, and use Swift's module system to separate the generated code from code that depends on it.

Another example is the lack of ability to customize how Swift names are computed from strings provided in the OpenAPI document.
```

simonjbeaumont (approved):

PR #510, opened 2024-01-19
Fix tests to include content-length header, now included by runtime library

czechboy0:
### Motivation

The runtime library now always includes content-length with bodies, and our tests verified the exact set of header fields present.

### Modifications

Update the expectations to include the content-length header.

### Result

Passing tests again.

### Test Plan

Tests are passing again.


simonjbeaumont:
I think this should come with an update of the minimum version of runtime.

czechboy0:
> I think this should come with an update of the minimum version of runtime.

Good point, let me update

Issue #246, opened 2023-09-05
Is it compatible with Xcode Cloud


leecaa:
Can it be used normally on Xcode Cloud? Because it uses a plugin to generate code, I've encountered issues with other repositories where I didn't have write permissions on Xcode Cloud, which prevented the code from being generated normally. [[Link to GitHub issue]](https://github.com/mac-cain13/R.swift/issues/777)

Thanks for any reply

simonjbeaumont:
@leecaa yes, it’s possible. But as you noted, it requires permissions. This post on the developer forums explains how to get the plugin working in Xcode Cloud: https://developer.apple.com/forums/thread/732893. 

Hope this helps.

PR #519, opened 2024-02-02
Disable use of auxiliary file to prevent permission error on XcodeCloud

novr:
### Motivation

- RELATED #246

### Modifications

The implementation of swift-openapi-generator is [try encoder.encode(container).write(to: url, atomically: true, encoding: .utf8)](https://github.com/apple/swift-openapi-generator/blob/6a98026199720b3e23b238248d4444bc065061b2/Sources/swift-openapi-generator/YamlFileDiagnosticsCollector.swift#L54C54-L54C64). if atomically is true, the content is written to an intermediate file and then moved to the outputURL. However, XocdeCloud does not seem to allow writing outside of the pluginWorkDirectory, even for temporary files created by the Foundation framework.

- https://developer.apple.com/documentation/foundation/nsdata/1408033-write

The easiest way to get Build to succeed on Xcode Cloud seems to be to change "atomically" to false.

What do you think about this change?

### Result

Swift Package Manager's Build Tools plug-in works on XcodeCloud

### Test Plan

N/A

### DIFF
--- a/Sources/swift-openapi-generator/YamlFileDiagnosticsCollector.swift
+++ b/Sources/swift-openapi-generator/YamlFileDiagnosticsCollector.swift
@@ -51,6 +51,6 @@ final class _YamlFileDiagnosticsCollector: DiagnosticCollector, @unchecked Senda
         let encoder = YAMLEncoder()
         encoder.options.sortKeys = true
         let container = _DiagnosticsYamlFileContent(uniqueMessages: uniqueMessages, diagnostics: sortedDiagnostics)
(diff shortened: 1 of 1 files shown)


simonjbeaumont:
Interesting. Are you saying that, with this patch, you no longer need the defaults-write workaround detailed here: https://developer.apple.com/forums/thread/732893?

novr:
@czechboy0 Sorry, this method didn't solve the problem.

simonjbeaumont:
No problem, thanks for taking the time to investigate @novr! Hopefully the defaults-write workaround detailed in the other issue will allow you to make use of Xcode Cloud, if that's your goal.

PR #524, opened 2024-02-05
Add macCatalyst to the PlatformChecks

### LINKED ISSUE #523: PlatformChecks emits error when compiling for MacCatalyst
### Description

When attempting to compile the Swift OpenAPI Generator on MacCatalyst, code that compiles perfectly on iOS will not run as it will fail on the `PlatformChecks`. This is unexpected. 

### Reproduction

Use the same instructions as used in https://developer.apple.com/wwdc23/10171 but enable the MacCatalyst target in the project settings. The project will compile fine for iOS but when targeting mac, the following error will be emitted:

`OpenAPIGeneratorCore is only to be used by swift-openapi-generator itself—your target should not link this library or the command line tool directly.`

### Package version(s)

1.2.0

### Expected behavior

I would have expected this to work fine without extra adjusments.

### Environment

macOS Sonoma
Xcode 15.2

### Additional information

I suspect that it has to do with the behaviour of MacCatalyst. When compiling the app normally `targetEnvironment(macCatalyst)` will be true while `os(macOS)` will be false which is already confusing by itself but with good reason. Perhaps the build tool plug-in is ran in a similar environment? 

It looked like from my limited testing that when disabling this check it will not include the runtime, so the setup should be correct.

lennartkerkvliet:
### Motivation

When attempting to compile the Swift OpenAPI Generator on MacCatalyst, code that compiles on iOS will _not_ compile as it will fail on the `PlatformChecks`. 

### Modifications

Added an extra condition in `PlatformChecks` to allow running on MacCatalyst.

### Result

fixes #523

### Test Plan

I've tried to familiarize myself with the project, but could not come up with a good way to add a test for running the tool on MacCatalyst. In my manual testing giving the instructions in the issue, it now works as expected.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/PlatformChecks.swift
+++ b/Sources/_OpenAPIGeneratorCore/PlatformChecks.swift
@@ -14,7 +14,7 @@
 
 // Emit a compiler error if this library is linked with a target in an adopter
 // project.
-#if !(os(macOS) || os(Linux))
+#if !(os(macOS) || targetEnvironment(macCatalyst) || os(Linux))
 #error(
     "_OpenAPIGeneratorCore is only to be used by swift-openapi-generator itself—your target should not link this library or the command line tool directly."
 )
--- a/Sources/_OpenAPIGeneratorCore/PlatformChecks.swift
+++ b/Sources/_OpenAPIGeneratorCore/PlatformChecks.swift
@@ -14,7 +14,9 @@
 
 // Emit a compiler error if this library is linked with a target in an adopter
 // project.
-#if !(os(macOS) || targetEnvironment(macCatalyst) || os(Linux))
+//
+// When compiling for MacCatalyst, the plugin is (erroneously?) compiled with os(iOS).
+#if !(os(macOS) || os(Linux) || (os(iOS) && targetEnvironment(macCatalyst)))
 #error(
     "_OpenAPIGeneratorCore is only to be used by swift-openapi-generator itself—your target should not link this library or the command line tool directly."
 )


simonjbeaumont:
> When attempting to compile the Swift OpenAPI Generator on MacCatalyst...

When compiling an iOS app, you run the generator on your macOS developer environment (usually by using the Swift package plugin), but you only link the _runtime library_ to your app. The platform checks here are to stop you linking the _generator_ itself in applications by mistake. 

Are you trying to compile _for_ MacCatalyst or _on_ MacCatalyst?

lennartkerkvliet:
Here's an Xcode project reproducing the issue. The same configuration compiles on iOS while failing to compile on Catalyst.
[OpenAPI-Catalyst.zip](https://github.com/apple/swift-openapi-generator/files/14173577/OpenAPI-Catalyst.zip)

simonjbeaumont:
Thanks for the response and for the reproducer. I can see locally with a dummy package that, when compiling the _application code_:

1. When compiling for destination "My Mac", I see `os(macOS)` is true and `targetEnvironment(macCatalyst)` is false.
2. When compiling for destination "My Mac (Mac Catalyst), I see `os(iOS)` is true and `targetEnvironment(macCatalyst)` is true.
3. When compiling for a simulator, e.g. "iPhone 15", I see `os(iOS)` is true and `targetEnvironment(simulator)` is true.

This is all expected in the code for the application. What's unexpected is the compiler conditions that are used to build the package plugin:

- In scenario (3), things work fine because when compiling the _plugin_ it uses `os(macOS)`.
- In scenario (2), things do not work because when compiling the _plugin_ it uses `os(iOS)`.

I think this is a bug in the build system. For now, we can workaround it with this patch—although I've suggested an edit to document it clearer. Would you mind filing an issue using Feedback Assistant and let us know the issue number?

Sources/_OpenAPIGeneratorCore/PlatformChecks.swift
```diff
@@ -14,7 +14,7 @@
 
 // Emit a compiler error if this library is linked with a target in an adopter
 // project.
-#if !(os(macOS) || os(Linux))
+#if !(os(macOS) || targetEnvironment(macCatalyst) || os(Linux))
```
simonjbeaumont, line 17:
```suggestion
// When compiling for Mac Catalyst, the plugin is (erroneously?) compiled with os(iOS).
#if !(os(macOS) || os(Linux) || (os(iOS) && targetEnvironment(macCatalyst)))
```

lennartkerkvliet: I think the comment is good to have, but maybe adding `os(iOS)` is making things more complicated than it has to. When `targetEnvironment(macCatalyst)` is true it's implicit that `os(iOS)` is also true.

lennartkerkvliet:
@simonjbeaumont Here's the feedback identifier: FB13589581

Sources/_OpenAPIGeneratorCore/PlatformChecks.swift
simonjbeaumont: > When `targetEnvironment(macCatalyst)` is true it's implicit that `os(iOS)` is also true.

IIUC the fact this implication currently holds is the bug we're working around. I'd rather make that explicit and, when this bug is addressed, we shouldn't require even `targetEnvironment(macCatalyst)`, which was why I suggested keeping it explicit in the condition itself.

lennartkerkvliet: Updated!

simonjbeaumont (approved):

simonjbeaumont:

simonjbeaumont:
Nightly failure is a compiler crash:

```
error: compile command failed due to signal 6 (use -v to see invocation)
SIL verification failed: keypath value type should match value type of keypath pattern
  any CodingKey
  URICoderCodingKey
Verifying instruction:
->   %8 = keypath $KeyPath<URIValueFromNodeDecoder.CodingStackEntry, any CodingKey>, (root $URIValueFromNodeDecoder.CodingStackEntry; stored_property #URIValueFromNodeDecoder.CodingStackEntry.key : $URICoderCodingKey) // user: %9
     %9 = move_value [lexical] [var_decl] %8 : $KeyPath<URIValueFromNodeDecoder.CodingStackEntry, any CodingKey> // users: %15, %11
In function:
// URIValueFromNodeDecoder.codingPath.getter
```

Merging over this.

simonjbeaumont:
Thanks @lennartkerkvliet!

PR #536, opened 2024-03-04
Extremely small: adds in a missing newline.

everyplace:
Most minimal of PRs, but the detail of the extra long line seems wrong… so here we are.

### Motivation
I noticed a very long line on the Package definition. 

### Modifications
One additional new line. 

### Result
Easier to read file.


simonjbeaumont:
Thank you for the attention to detail but unfortunately this is a bug in Swift-format. Our CI pipeline will fail this PR.

everyplace:
@simonjbeaumont Thank you so much for the response, and that makes a lot of sense.

Issue #259, opened 2023-09-09
Support deepObject style for query params


czechboy0:
Support deepObject style for query params.

Used by the Stripe API, for example.

czechboy0:
Not just for query items, but also URL encoded bodies, once we support them.

tib:
I just encountered this issue and I'd love to see better support for query params, let me know if you need help with it. :)

czechboy0:
Yes support for at least deepObject is planned for 1.0, and it would be awesome if you could contribute it. I suspect most of the changes would be in the runtime library's URIEncoder and URIDecoder, please take a look there and see if adding deepObject support would be a project you want to take on. Note that while it's being improved, it might also make sense to just do https://github.com/apple/swift-openapi-generator/issues/258 as well, as the latter might be a natural extension of the existing logic. deepObject will be slightly more involved, but shouldn't be too difficult.

Take a look at how the encoder/decoder works today and see if you can extend it, and I'm here if you have any questions 🙂

czechboy0:
I'll assign you to the issue for now, but don't feel obligated if you decide not to do it, just let me know and I'll unassign it. It's to avoid multiple people working on the same issue 🙂

czechboy0:
A related ask for this: https://github.com/apple/swift-openapi-generator/issues/292

czechboy0:
Ok just let us know if you start working on this @tib, but no pressure. I unassigned it for now to leave it available for others to pick up.

kstefanou52:
Hi @czechboy0, I encountered the same issue so I took the initiative to address it. 
Excited to hear your thoughts when you get a chance to check my PRs out. Cheers! 🚀

- https://github.com/apple/swift-openapi-runtime/pull/100
- https://github.com/apple/swift-openapi-generator/pull/538

czechboy0:
Runtime changes landed in swift-openapi-runtime 1.4.0.

czechboy0:
Landed in #538

PR #538, opened 2024-03-07
[Generator] Add support of deepObject style in query params

kstefanou52:
### Motivation

The generator changes for: 
[apple/swift-openapi-generator](https://github.com/apple/swift-openapi-generator/issues/259)

Depends on https://github.com/apple/swift-openapi-runtime/pull/100 
landing first and getting released, and the version dependency being bumped here.

### Modifications

Added `deepObject` style to serializer & parser in order to support nested keys on query parameters.

### Result

Support nested keys on query parameters.

### Test Plan

Adapted snippet tests (SnippetBasedReferenceTests)

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
@@ -131,7 +131,9 @@ extension FileTranslator {
             switch location {
             case .query:
                 switch style {
-                case .form, .deepObject:
+                case .form:
+                    break
+                case .deepObject where explode:
                     break
                 default:
                     try diagnostics.emitUnsupported(
--- a/Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
@@ -131,10 +131,8 @@ extension FileTranslator {
             switch location {
             case .query:
                 switch style {
-                case .form:
-                    break
-                case .deepObject where explode:
-                    break
+                case .form: break
+                case .deepObject where explode: break
                 default:
                     try diagnostics.emitUnsupported(
                         "Query params of style \(style.rawValue), explode: \(explode)",
--- a/Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
@@ -130,7 +130,10 @@ extension FileTranslator {
             let location = parameter.location
             switch location {
             case .query:
-                guard case .form = style else {
+                switch style {
+                case .form, .deepObject:
+                    break
+                default:
                     try diagnostics.emitUnsupported(
                         "Query params of style \(style.rawValue), explode: \(explode)",
                         foundIn: foundIn
@@ -243,6 +246,7 @@ extension OpenAPI.Parameter.SchemaContext.Style {
     var runtimeName: String {
         switch self {
         case .form: return Constants.Components.Parameters.Style.form
+        case .deepObject: return Constants.Components.Parameters.Style.deepObject
         default: preconditionFailure("Unsupported style")
         }
     }
--- a/Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
+++ b/Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
@@ -2600,6 +2600,18 @@ final class SnippetBasedReferenceTests: XCTestCase {
                       type: array
                       items:
                         type: string
+                  - name: sort
+                    in: query
+                    required: true
+                    style: deepObject
+                    explode: true
+                    schema:
+                      type: object
+                      properties:
+                        option1:
+                          type: string
+                        option2:
+                          type: string
                 responses:
                   default:
                     description: Response
@@ -2610,18 +2622,36 @@ final class SnippetBasedReferenceTests: XCTestCase {
                         public var single: Swift.String?
                         public var manyExploded: [Swift.String]?
                         public var manyUnexploded: [Swift.String]?
+                        public struct sortPayload: Codable, Hashable, Sendable {
+                            public var option1: Swift.String?
+                            public var option2: Swift.String?
+                            public init(
+                                option1: Swift.String? = nil,
+                                option2: Swift.String? = nil
+                            ) {
+                                self.option1 = option1
+                                self.option2 = option2
+                            }
+                            public enum CodingKeys: String, CodingKey {
+                                case option1
+                                case option2
+                            }
+                        }
+                        public var sort: Operations.get_sol_foo.Input.Query.sortPayload
                         public init(
                             single: Swift.String? = nil,
                             manyExploded: [Swift.String]? = nil,
-                            manyUnexploded: [Swift.String]? = nil
+                            manyUnexploded: [Swift.String]? = nil,
+                            sort: Operations.get_sol_foo.Input.Query.sortPayload
                         ) {
                             self.single = single
                             self.manyExploded = manyExploded
                             self.manyUnexploded = manyUnexploded
+                            self.sort = sort
                         }
                     }
                     public var query: Operations.get_sol_foo.Input.Query
-                    public init(query: Operations.get_sol_foo.Input.Query = .init()) {
+                    public init(query: Operations.get_sol_foo.Input.Query) {
                         self.query = query
                     }
                 }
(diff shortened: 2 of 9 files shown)


simonjbeaumont:
Thanks @kstefanou52 for taking the time to tackle this. We'll park this PR review until we land the runtime PR.

czechboy0:
@kstefanou52 Ok please bump the runtime dependency to 1.4.0 in Package.swift and you should be able to finish the generator changes.

kstefanou52:
@czechboy0 Done! I have created some tests, but I'm not sure if you would like to add more. Could you please suggest any additional tests you think might be necessary? Thank you!

Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
```diff
@@ -139,12 +139,14 @@ extension FileTranslator {
...
-                guard case .form = style else {
+                switch style {
+                case .form, .deepObject:
+                    break
+                default:
                     diagnostics.emitUnsupported(
                         "Query params of style \(style.rawValue), explode: \(explode)",
                         foundIn: foundIn
                     )
-                    return nil
```
czechboy0, line 147: I think the `return nil` needs to stay, just should live in the `default` case now.

czechboy0 (changes requested):
One code change suggested, and regarding the extra tests - you're right 🙂 


Start by adding a similar parameter like you did in the snippet tests, but this time to `petstore.yaml`, which makes it available in generated form in `PetstoreConsumerTests`. There, please extend one client and one server unit test to verify that the serialization/deserialization is correctly piped through all the way. The relevant test classes are `Test_Client` and `Test_Server`.

Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
kstefanou52: That's correct, I must have deleted it by mistake.

simonjbeaumont:
Just checking in on this one. Is it ready for review again? (Thanks for the effort here!)

kstefanou52:
Hi @simonjbeaumont! Unfortunately some tests are failing, I'll work on this on weekend and I'll let you know. Cheers!

czechboy0:
Hi @kstefanou52 - do you expect to be able to come back to finish this? 🙂 It's also okay if you can't and someone else will pick up where you left off.

kstefanou52:
@czechboy0 I'm apologising for my late response. I'll try to catch up and complete the PR by the end of the week.

czechboy0:
Hi @kstefanou52 - let us know if you need any help with this? I think it just needs to be updated from the main branch and a conflict resolved, but otherwise might be good to go?

Tests/PetstoreConsumerTests/Test_Server.swift
```diff
@@ -33,6 +33,7 @@ final class Test_Server: XCTestCase {
             XCTAssertEqual(input.query.habitat, .water)
             XCTAssertEqual(input.query.since, .test)
             XCTAssertEqual(input.query.feeds, [.carnivore, .herbivore])
+            XCTAssertEqual(input.query.sort, .init(id: "ascending", name: "descending"))
```
kstefanou52, line 36: This test fails because input.query.sort is nil. It appears there’s an issue in the runtime library method [decodeRootIfPresent(_ type:) throws -> T](https://github.com/apple/swift-openapi-runtime/blob/daa2fb54fe4a7f5187d7286047d5144c8cb97477/Sources/OpenAPIRuntime/URICoder/Decoding/URIValueFromNodeDecoder.swift#L83).
The node being returned is [id: ascending, name: descending], and, as expected, there's no `node[rootKey]`, so `currentElementAsArray().isEmpty` returns true and the decoding fails.

I think we may need to update the runtime library with something like this:
```
func decodeRootIfPresent<T: Decodable>(_ type: T.Type = T.self) throws -> T? {
    // The root is only nil if the node is empty.
    if try currentElementAsArray().isEmpty { return nil }
    if style != .deepObject, try currentElementAsArray().isEmpty {
        return nil
    }
    return try decodeRoot(type)
}
```

kstefanou52:
Hello @czechboy0 and @arthurcro, I've made some progress, but I'm facing an issue with one of the unit tests. The code appears to work as expected, but I suspect there may be a problem with the runtime library. I’d appreciate your thoughts on this; more details can be found [here](https://github.com/apple/swift-openapi-generator/pull/538/commits/adfa77efbe029a745ce7c26297bc9427bcda5e9d#r1817848222).

czechboy0:
Can you elaborate? The link doesn't really say what the issue is - what are you seeing and what are you expecting?

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
```diff
@@ -2514,6 +2514,18 @@ final class SnippetBasedReferenceTests: XCTestCase {
                       type: array
                       items:
                         type: string
+                  - name: sort
+                    in: query
+                    required: true
```
czechboy0, line 2519:
```suggestion
                    required: false
```

Let's use `required: false`, or if you'd like to test it being required, make one of the properties required.

Tests/PetstoreConsumerTests/Test_Server.swift
czechboy0: Yeah that's possible, can you open a PR to the runtime library with a test first so we agree on what the expected behavior should be? If you try that, does that make this test pass?

kstefanou52: Yes it passed!

czechboy0: Ok let's get that change into the runtime library then!

czechboy0: @kstefanou52 Will you be able to open that PR to the runtime library?

kstefanou52:
Hey @czechboy0, I just opened the [runtime PR](https://github.com/apple/swift-openapi-runtime/pull/125). Please let me know if there are any additional tests we should add. Thanks!

Tests/PetstoreConsumerTests/Test_Server.swift
kstefanou52: Done!

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
```diff
@@ -2514,6 +2514,18 @@ final class SnippetBasedReferenceTests: XCTestCase {
...
+                    required: false
+                    style: deepObject
+                    explode: true
+                    schema:
+                      type: object
+                      properties:
+                        option1:
+                          type: string
+                        option2:
+                          type: string
```
czechboy0, line 2528: Please add one more parameter here, one that is required and has one of the object properties required as well. I want to see that the default parameter is working correctly below.

czechboy0 (changes requested):
Sorry it took longer than expected, the deepObject improvements landed in runtime 1.7.0, so you should be able to bring this branch up to speed again. Added a few suggestions

kstefanou52:
Hello @czechboy0 , I hope you are doing well. I found some time yesterday to rebase my PR and add the unit tests you requested.

czechboy0 (approved):
Looks great, thank you!

czechboy0:
@kstefanou52 Seems some of the changed tests are failing, and the formatter needs to be rerun (details: https://github.com/apple/swift-openapi-generator/blob/main/CONTRIBUTING.md#run-ci-checks-locally)

czechboy0:
@kstefanou52 Still seems to be failing

alishba-bloomandwild:
Hi @czechboy0, just checking in to see if you're able to merge these changes yet or if the tests are still failing?

czechboy0:
Hi @alishba-bloomandwild - there now seems to be a conflict. I'll let @kstefanou52 take a look when they have time, but this is close to landing.

kstefanou52:
@czechboy0 Hey! I'll do a rebase and run the tests locally during the week. I'll let you know when it will be ready.

alishba-bloomandwild:
I appreciate you taking the time to look into this, as it's currently a blocker for us. @kstefanou52

mheh:
Excited about this one too.

kstefanou52:
@czechboy0 @alishba-bloomandwild I plan to work on this during the weekend. I hope by Monday it will be ready. Thanks for your patience.

kstefanou52:
@czechboy0 I ran the tests locally and I'm getting a lot of `trailing spaces` errors in the Soundness/YAML lint check, even on files I haven't touched. Is that normal? Should I fix them? Thanks!

czechboy0:
The formatter in CI is happy, so it might be a local issue. I'll merge this once the rest of the CI passes - thanks! 🙏

Issue #513, opened 2024-01-24
Advice regarding nullable refs?


groue:
### Question

Hello,

I'm writing a client with an openapi.yml that frequently uses nullable ref properties, as below:

```yaml
// Technique 1
my_field:
  nullable: true
  allof:
    - $ref: '#/components/schemas/MyField'
```

An alternative technique, AFAIK, is:

```yaml
// Technique 2
my_field:
  oneOf:
    - $ref: '#/components/schemas/MyField'
    - type: 'null'
```

With both techniques, the generated code is not quite easy to deal with, both when decoding responses or encoding requests. The generated payload is a struct that has a non-semantic `value1` property, or an enum that does not expose any accessor for extracting its non-null value.

```swift
// Technique 1
public struct my_fieldPayload: Codable, Hashable, Sendable {
    public var value1: Components.Schemas.MyField
    public init(value1: Components.Schemas.MyField) {
        self.value1 = value1
    }
}

// Technique 2
@frozen public enum my_fieldPayload: Codable, Hashable, Sendable {
    case MyField(Components.Schemas.MyField)
    case case2(OpenAPIRuntime.OpenAPIValueContainer)
}
```

We're far from the expected Swift optionals. The struct generated by the technique 1 is surprisingly shallow.

Did any member in the community meet the same problem? Is there any workaround? Or a better YAML technique for specifying a nullable ref, while preserving good Swift ergonomics? Maybe I'm just holding it wrong?

czechboy0:
This looks a bit related to a question from just yesterday: https://github.com/apple/swift-openapi-generator/issues/512

What is the use case? What are the HTTP semantics you're trying to express?

Note that `nullable` is only OpenAPI 3.0 syntax, and adding a `null` to types is the equivalent in OpenAPI 3.1. So which you use depends on the version of the OpenAPI specification you're using.

groue:
> This looks a bit related to a question from just yesterday: #512

Yes: in both issues the OpenAPI spec describes optionality with a _pattern_, and the generated code creates the desire to share some remarks.

Two such patterns are described in this issue, and another one in #512. I guess there are others.

> What is the use case? What are the HTTP semantics you're trying to express?

[...]

PR #557, opened 2024-03-31
Support `type: 'null'` as OpenAPIValueContainer

brandonbloom:
### Motivation

OpenAPIKit's handling of `type: 'null'` was problematic with older Yams versions. See https://github.com/mattpolzin/OpenAPIKit/pull/366/files for a test demonstrating that & dependency version bump.

Similarly, swift-openapi-generator explicitly does not "support" `type: 'null'`. As part of searching for something to address https://github.com/apple/swift-openapi-generator/issues/513 I have added such support for explicit nulls. This gives one more sort of input schema where swift-openapi-generator produces some useful output instead of skipping a schema.

### Modifications

- Bumped minimum supported Yams version.
- Return`true` from `isSchemaSupported` for `type: 'null'` schemas.
- Defined `.null` to have a builtint ype of `OpenAPIValueContainer`

### Result

Nothing should change for existing users because `type: null` schemas were previously unsupported. Going forward, usages of these schemas is supported.

### Test Plan

New test included.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeMatcher.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeMatcher.swift
@@ -289,6 +289,7 @@ struct TypeMatcher {
     private static func _tryMatchBuiltinNonRecursive(for schema: JSONSchema.Schema) -> TypeUsage? {
         let typeName: TypeName
         switch schema {
+        case .null(_): typeName = TypeName.valueContainer
         case .boolean(_): typeName = .swift("Bool")
         case .number(let core, _):
             switch core.format {
@@ -331,7 +332,7 @@ struct TypeMatcher {
             // arrays are already recursed-into by _tryMatchTypeRecursive
             // so just return nil here
             return nil
-        case .reference, .not, .all, .any, .one, .null:
+        case .reference, .not, .all, .any, .one:
             // never built-in
             return nil
         }
--- a/Package.swift
+++ b/Package.swift
@@ -52,7 +52,7 @@ let package = Package(
 
         // Read OpenAPI documents
         .package(url: "https://github.com/mattpolzin/OpenAPIKit", from: "3.1.2"),
-        .package(url: "https://github.com/jpsim/Yams", "4.0.0"..<"6.0.0"),
+        .package(url: "https://github.com/jpsim/Yams", "5.1.0"..<"6.0.0"),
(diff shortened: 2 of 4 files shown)


Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeMatcher.swift
```diff
@@ -289,6 +289,7 @@ struct TypeMatcher {
     private static func _tryMatchBuiltinNonRecursive(for schema: JSONSchema.Schema) -> TypeUsage? {
         let typeName: TypeName
         switch schema {
+        case .null(_): typeName = TypeName.valueContainer
```
czechboy0, line 292: Hmm I'm still unsure whether we want to promote null to a JSON value here. Let's discuss in the other PR first, then we can come back to this.

brandonbloom: I also considered the singleton instance of NSNull, or some new enum type with one case, or `()` or something. It wasn't obvious if there was a more appropriate type for this in Swift.

czechboy0: Hmm maybe we should introduce an `typealias OpenAPINull = Void` in the runtime library, and use that? What do you think? cc @simonjbeaumont 

`OpenAPIValueContainer` isn't appropriate, I think, as it erases our knowledge of it being null, forcing the adopter to check again later. We should be able to avoid that.

czechboy0:
Based on the discussion in the [other](https://github.com/apple/swift-openapi-generator/pull/558#issuecomment-2034319168) thread, I think we should not merge this PR, and instead we should simply filter out `null` from anyOf/allOf schemas, as OpenAPIKit will already represent the nullability on the parent schema.

simonjbeaumont:
We had a report that this is blocking generation for the Figma API: https://github.com/figma/rest-api-spec/blob/main/openapi/openapi.yaml, which I can confirm doesn't compile as-is.

Brett-Best:
It would be great to get this merged as it unblocks using the Figma API as Simon mentions.
Anything I can do to help?

czechboy0:
I think we can't merge it as-is, as it'd be an API break. Plus, see my earlier message, I think we should fix this slightly differently.

Abdullah4345 (approved):

Issue #286, opened 2023-09-20
Swift OpenAPI Generator Fails to Decode anyOf Structure for nullable Field


brandonmaul:
When using the Swift OpenAPI Generator and runtime to decode a JSON payload from an API endpoint, a decoding error occurs for a field defined with the anyOf keyword in the OpenAPI spec. The error message is "The anyOf structure did not decode into any child schema."

Here is a simplified snippet of the OpenAPI spec that defines the response schema:

```
"ResponseSchema": {
  "properties": {
    "status": { "type": "string", "title": "Status" },
    "timestampUTC": {
      "type": "string",
      "format": "date-time",
      "title": "Timestamputc"
    },
    "errorCode": {
      "anyOf": [{ "type": "integer" }, { "type": "null" }],
      "title": "Errorcode"
    },
    "errorMessage": {
      "anyOf": [{ "type": "string" }, { "type": "null" }],
      "title": "Errormessage"
    },
    "data": {
      "type": "object"
    }
  },
  "type": "object",
  "required": [
    "status",
    "timestampUTC",
    "errorCode",
    "errorMessage",
    "data"
  ]
}
```

Here is a snippet of the JSON payload received from the API endpoint:

```
{
  "status": "success",
  "timestampUTC": "2023-09-20T12:20:29.606380Z",
  "errorCode": null,
  "errorMessage": null,
  "data": {}
}

```

### Error Message:
`DecodingError: valueNotFound errorCodePayload - at CodingKeys(stringValue: "errorCode", intValue: nil): The anyOf structure did not decode into any child schema. (underlying error: <nil>)`

### Steps to Reproduce:
Generate Swift code using the provided OpenAPI spec.
Make an API request to receive the JSON payload.
Attempt to decode the JSON payload using the generated Swift code.
Expected Behavior:
The JSON payload should be successfully decoded into the corresponding Swift model.

### Actual Behavior:
Decoding fails with the error message mentioned above.

### Environment:
Swift OpenAPI Generator version: 0.1.3
Swift version: 5.9
Runtime: 0.1.4

czechboy0:
Just to be safe, can you try with the latest 0.2.x version and the feature flag `nullableSchemas`? It might not resolve it, but I'd like to make sure. Thanks 🙏

brandonmaul:
Updated to version 0.2.X, but still seeing the same behavior. updated original description

czechboy0:
[...]

PR #558, opened 2024-04-01
Improvements to nullable references.

brandonbloom:
### Motivation

A common pattern in OpenAPI schemas I have seen uses `oneOf` with references like the following:

```
oneOf:
  - $ref: '#/components/schemas/Thing'
  - type: 'null'
```

Prior to this patch, this returns a enum for the possible variants of this schema, but an `Optional<T>` would be more expected and ergonomic.

### Modifications

In addition to the basic support for explicit nulls (#557, included in this PR too), I've introduced a notion of an "optional root", which is a schema / type usage in which the optionality occurs at the top level and not through some indirection of references. This undoes the prior behavior of propagating optionality all the way to the top, instead adding it only where needed.

This new model of optionality relies on relies on enhancements to `TypeMatcher isOptional` to look for explicit nulls and to support following references and dealing with potentially recursive schemas.

Finally, TypesFileTranslator detects the `{oneOf: [{type: 'null'}, ...]}` pattern and returns a Swift `Optional<...remaining cases...>` instead.

### Result

The added `testOneOfRefOrNull` test shows the new behavior in action.

This would be a breaking change for anyone using this common pattern. A few places where enums were needed will need to be migrated to use optionals instead of the generated enums. It's a mechanical and straightforward migration, resulting in less and cleaner code.

### Test Plan

Several new unit tests were added. See diff.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateAllAnyOneOf.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateAllAnyOneOf.swift
@@ -38,7 +38,7 @@ extension TypesFileTranslator {
     /// - Throws: An error if there is an issue during translation.
     /// - Returns: A declaration representing the translated allOf/anyOf structure.
     func translateAllOrAnyOf(typeName: TypeName, openAPIDescription: String?, type: AllOrAnyOf, schemas: [JSONSchema])
-        throws -> Declaration
+        throws -> [Declaration]
     {
         let properties: [(property: PropertyBlueprint, isKeyValuePair: Bool)] = try schemas.enumerated()
             .map { index, schema in
@@ -107,7 +107,7 @@ extension TypesFileTranslator {
                 properties: propertyValues
             )
         )
-        return structDecl
+        return [structDecl]
     }
 
     /// Returns a declaration for a oneOf schema.
@@ -128,7 +128,7 @@ extension TypesFileTranslator {
         openAPIDescription: String?,
         discriminator: OpenAPI.Discriminator?,
         schemas: [JSONSchema]
-    ) throws -> Declaration {
+    ) throws -> [Declaration] {
         let cases: [(String, [String]?, Bool, Comment?, TypeUsage, [Declaration])]
         if let discriminator {
             // > When using the discriminator, inline schemas will not be considered.
@@ -148,7 +148,15 @@ extension TypesFileTranslator {
                 return (caseName, mappedType.rawNames, true, comment, mappedType.typeName.asUsage, [])
             }
         } else {
-            cases = try schemas.enumerated()
+            let (schemas, nullSchemas) = schemas.partitioned(by: { typeMatcher.isNull($0) })
+            if schemas.count == 1, nullSchemas.count > 0, let schema = schemas.first {
+                return try translateSchema(
+                    typeName: typeName,
+                    schema: schema,
+                    overrides: .init(isOptional: true))
+            }
+            cases = try schemas
+                .enumerated()
                 .map { index, schema in
                     let key = "case\(index+1)"
                     let childType = try typeAssigner.typeUsage(
@@ -242,6 +250,6 @@ extension TypesFileTranslator {
             conformances: Constants.ObjectStruct.conformances,
             members: caseDecls + codingKeysDecls + [decoder, encoder]
         )
-        return .commentable(comment, enumDecl)
+        return [.commentable(comment, enumDecl)]
     }
 }
--- a/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateSchema.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateSchema.swift
@@ -141,29 +141,26 @@ extension TypesFileTranslator {
                 arrayContext: arrayContext
             )
         case let .all(of: schemas, core: coreContext):
-            let allOfDecl = try translateAllOrAnyOf(
+            return try translateAllOrAnyOf(
                 typeName: typeName,
                 openAPIDescription: overrides.userDescription ?? coreContext.description,
                 type: .allOf,
                 schemas: schemas
             )
-            return [allOfDecl]
         case let .any(of: schemas, core: coreContext):
-            let anyOfDecl = try translateAllOrAnyOf(
+            return try translateAllOrAnyOf(
                 typeName: typeName,
                 openAPIDescription: overrides.userDescription ?? coreContext.description,
                 type: .anyOf,
                 schemas: schemas
             )
-            return [anyOfDecl]
         case let .one(of: schemas, core: coreContext):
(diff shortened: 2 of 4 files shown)


czechboy0:
Hi @brandonbloom,

thank for the PR! We might just need to split this up a bit before merging.

First thing - this is a breaking change, and if merged as-is, would require us to tag a 2.0. That's not really desirable at this stage, so I wonder what you suggest we do here.

Can you talk more about the reasoning of why you think the generated code needs to change? We've avoided special-casing specific patterns so far, as they don't evolve very well when a new case is added. Can you provide more details of why we should diverge from that approach here?

brandonbloom:
> We might just need to split this up a bit before merging.

Understood. I broke up the steps as separate commits for that reason.

> Can you talk more about the reasoning of why you think the generated code needs to change?

I found the behavior that exists prior to this change very confusing. The generated types don't match the schemas that they are generated from as it pertains to optionality. Instead, the optionality is bubbled up to the top level, and baked into the usage sites. This means that if you use a typealias for something that is expected to be `T?`, you actually just get a `T`.

However, the main justification is the `{oneOf: [{type: 'null'}, ...]}` pattern, which is used widely in a pair of OpenAPI specs I'm working with. In one case, it works way because the spec it itself generated from bespoke schema system which produces a referenceable schema for each named type and then uses the oneOf-null pattern everywhere. In the other case, the upstream schema used to use `nullable: true` keyword, but that's deprecated now in OpenAPI 3.1 and hence the move to the oneOf:null/... pattern.

> We've avoided special-casing specific patterns so far, as they don't evolve very well when a new case is added.  Can you provide more details of why we should diverge from that approach here?

That makes sense. Only null is handled specially in this change. I preserved the behavior that a single-case enum remains if null is not involved. I'm sensitive to the evolution argument, but I think that there are enough ergonomics for Optionals in Swift that it's worthwhile. With this change, my usage code has eliminated numerous case-lets and switches and other heavy syntax is favor of much lighter if-lets and `?.` operators.

> would require us to tag a 2.0. That's not really desirable at this stage, so I wonder what you suggest we do here.

There are two separate breaking changes. The "optional root" behavior and then the subsequent rule for simplifying the oneOf:null/... pattern. The former supports the latter, but the latter could be a setting or preference quite easily, if that's appropriate. Not sure if/how the former can be mitigated.

mattpolzin:
This reminds me a lot of [this comment](https://github.com/apple/swift-openapi-generator/issues/286#issuecomment-1728264815) (and the preceeding/following one). Are we talking about a different situation in this case?

mattpolzin:
If I recall correctly, one reason I want to avoid removing the `.null()` schema from the `oneOf` (and similar) is because it is lossy so OpenAPIKit no longer knows where to put the "nullability" when re-encoding (it could be a part of any of the subschemas).

czechboy0:
Good call, I forgot about that other issue. So a null case already gets propagated to the parent anyOf/allOf today? If so, does that mean we just need to filter out the null case in the generator and the rest should work?

mattpolzin:
> does that mean we just need to filter out the null case in the generator and the rest should work?

Correct; OpenAPIKit currently sets any/one/all-of schemas as nullable if any of the subschemas are nullable. 

Also, I keep waking up more as the morning goes on and now I'm thinking there is no current bug with the way OpenAPIKit handles `oneOf` (I speculated there was in my previous comment's EDIT).

czechboy0:
Ok great - @brandonbloom would you be interested in opening a PR for the filtering of nulls, which should hopefully improve how we handle these cases? It's my understanding we completely skip such anyOf/allOfs today, so they don't get generated. If we filter the null out, it should start getting generated as an optional anyOf/allOf, which I what I think we want. We can keep discussing the collapsing of that anyOf/allOf separately, but that'd be a feature flag anyway, as it'd be an API-break, so we shouldn't block this work for that.

jmg-duarte:
@czechboy0 I'm currently stuck on a compilation error stating:
```
Schema "null" is not supported, reason: "schema type", skipping [context: foundIn=Components.Schemas.Body_authSignin
```
This goes on for a bit with a bunch of other endpoints. 

I see that https://github.com/apple/swift-openapi-generator/pull/557 is open so I assume some work is being done to address this, but is there anything I can do to work around this issue?

czechboy0:
Hi @jmg-duarte - as a workaround, you could remove the anyOf that wraps your schema and a null. If the property isn't included in the object's `required` array, it'll be generated as optional _anyway_, so the extra wrapping in `anyOf` that also includes a null is not necessary, and in fact will make the generated code uglier in Swift.

Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateTypealias.swift
```diff
@@ -28,7 +28,7 @@ extension FileTranslator {
         let typealiasDescription = TypealiasDescription(
             accessModifier: config.access,
             name: typeName.shortSwiftName,
-            existingType: .init(existingTypeUsage.withOptional(false))
+            existingType: .init(existingTypeUsage)
```
x128, line 31: Why is this change necessary?
Let's say I have an optional string in my schema:
```
"line2": {
	"title": "Line2",
	"oneOf": [
		{
			"type": "string",
		},
		{
			"type": "null"
		}
	]
}
```
It's not marked as `required`. Full json is [attached](https://github.com/user-attachments/files/15789711/openapi.json).

So now, with this change, it generates a nested optional type:
```
/// - Remark: Generated from `#/components/schemas/Address/line2`.
public typealias line2Payload = Swift.String?
/// - Remark: Generated from `#/components/schemas/Address/line2`.
public var line2: Components.Schemas.Address.line2Payload?
```
And if we have an optional struct, not a String, it also generates a nested optional, so we have to access its values using `generatedObject??.property` which is a bit weird.

Without this change, everything seems to work. I didn't find an answer in test cases as well. I was thinking about forking and undoing this change, but maybe I'm missing something, so what's behind that?

P.S. Great job @brandonbloom. I have no idea how much effort is needed to support `anyOf`/`allOf` correctly. But in our project we probably need only `oneOf` for optionals, so this PR is a really nice work.

Sokartecnologi (approved):

jasondiab:
Can this be merged given it is now approved? It would help us avoid using scattered workarounds to make this work 🙏 thanks.

czechboy0:
Hi @jasondiab, this PR is not approved by any of the maintainers, and still needs work. If landed as-is, it'd be a major breaking change.

Sokartecnologi:
https://github.com/simibac/ConfettiSwiftUI.git

colincornaby:
Could this become an opt in feature? It is breaking, but it (along with #557) are causing a lot of churn in having to hand tweak OpenAPI being ingested by the parser. A required opt in would prevent it from breaking semantic versioning.

mattpolzin:
If I understand the above conversation fully then this PR might not represent the easiest solution which would hopefully simply involve filtering out .null() schemas before generating code. Food for thought if someone has time to try that.

PR #587, opened 2024-07-10
Only write `@frozen` on public interfaces

IanHoar:
### Motivation

Fixes a new warning in Xcode 16: `@frozen has no effect on non-public enums`

### Modifications

When rendering enums, first check if the frozen modifier is required, then check if the access modifier is explicitly public. If both conditions are satisfied, render the `@frozen` modifier

### Result

Frozen modifier will be in tact for public enums, but omitted for all other access levels 

### Test Plan

Integrated into my app, checked the warnings and they were fixed. 

| Before | After |
| --- | --- | 
| Screenshot 2024-07-10 at 9 37 18 AM | Screenshot 2024-07-10 at 9 39 50 AM |

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Renderer/TextBasedRenderer.swift
+++ b/Sources/_OpenAPIGeneratorCore/Renderer/TextBasedRenderer.swift
@@ -645,7 +645,9 @@ struct TextBasedRenderer: RendererProtocol {
 
     /// Renders the specified enum declaration.
     func renderEnum(_ enumDesc: EnumDescription) {
-        if enumDesc.isFrozen {
+        if enumDesc.isFrozen,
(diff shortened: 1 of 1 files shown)


czechboy0:
Thanks @IanHoar! Can you also try with the `package` modifier? Does it produce a warning and should also be included in this condition or not?

ojun9:
@czechboy0 

Since there were no changes in the PR https://github.com/apple/swift-openapi-generator/pull/587, I created a new PR https://github.com/apple/swift-openapi-generator/pull/595. Please let me know if there are any issues!

czechboy0:
Already addressed in https://github.com/apple/swift-openapi-generator/pull/595

PR #588, opened 2024-07-11
[GHA] Call the reusable package workflow

FranzBusch:
# Motivation

We are migrating our CI over to GH actions. We have created a reusable workflow inside the NIO repository that we want to move into a separate repo in the future. For testing purposes let's try out calling the reusable workflow from the this repo as well.

# Modification

Add a GH actions workflow that calls the reusable package workflow.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,0 +1,13 @@
+name: Pull Request
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review]
+ 
+jobs:
+  call-reusable-pull-request-workflow:
+    name: Checks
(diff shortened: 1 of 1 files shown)


czechboy0:
Thanks @FranzBusch - is there a way to call that workflow manually on this PR before merging it, to test it out?

FranzBusch:
> Thanks @FranzBusch - is there a way to call that workflow manually on this PR before merging it, to test it out?

Not that I know of. We have to land it on main once and can start iterating then. FYI I fully expect a few pipelines to fail since I need some fixes to handle the cross repo scripts. Would be great if we can land this though to start iterating. The checks aren't required so nothing should get blocked by them

czechboy0:

czechboy0 (approved):

PR #589, opened 2024-07-12
add support for visionOS

ruralharry:
### Motivation

When I use it in visionOS project, Xcode compiles and reports errors. When checking the platform, there is indeed little support for the latest visionOS.

### Modifications

add os(visionOS) when checking the platform 

### Result

Xcode has been compiled successfully and is used normally.

### Test Plan

Passed the test, does not affect other platforms

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/PlatformChecks.swift
+++ b/Sources/_OpenAPIGeneratorCore/PlatformChecks.swift
@@ -16,7 +16,7 @@
 // project.
 //
 // When compiling for MacCatalyst, the plugin is (erroneously?) compiled with os(iOS).
-#if !(os(macOS) || os(Linux) || (os(iOS) && targetEnvironment(macCatalyst)))
+#if !(os(macOS) || os(visionOS) || os(Linux) || (os(iOS) && targetEnvironment(macCatalyst)))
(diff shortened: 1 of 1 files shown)


ruralharry:
image
image
image
@czechboy0 As shown in the picture above, I thought it was a compilation error. I just tried again. The compilation was successful, but if you open this file, as shown in the third picture, there will be an error.
In this way, it seems that this error does not affect the normal use of the project.

czechboy0:
Can you post a pic of the "Link Binary with Libraries" section expanded, similar to the second one?

ruralharry:
image
image
@czechboy0

czechboy0:
Okay that looks configured correctly now. If you delete Derived Data and rebuild, does it work now?

ruralharry:
@czechboy0 Thank you for your patience. I cleaned the build folder, checked that the generated folder was also empty, and then recompiled it. It was successful. The files that should be generated also exist. I tried several times and succeeded every time.
It seems that the project can be used normally. Only when you open this PlatformChecks.swift file, you will find an error, and this error has no adverse effect.

czechboy0:
Ok glad it's working!

PR #591, opened 2024-07-12
[GHA] Configure license check

FranzBusch:


scripts/check-for-broken-symlinks.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
```
FranzBusch, line 1: @simonjbeaumont Was there a particular reason for using `bash` from the `env`. My understanding is that using `/bin/bash` is preferred since it specifies a particular executable and isn't open to path injections

simonjbeaumont: This is a script that devs or CI run. For devs, they want their environment to be taken into account. For CI we control the environment.

This can be useful for images that may have a Bash 3 and Bash 4, for example.

FranzBusch: If we want to support both paths then we need to adapt our license check script. NIO is only using `/bin/bash` whereas OpenAPI is using the env based variant.

FranzBusch: I assumed you want me to revert the changes that I did here. Currently the license script is hard coded to check for `#!/bin/bash` in `sh` files. We would need to extend this to also support `!/usr/bin/env bash`

simonjbeaumont: Only in the spirit of consistency, correct? As in, nothing is blocked on these being equal? Presumably both scripts will run equally fine regardless of what we do in this PR?

FranzBusch: Let me clarify. If I don't change `#!/usr/bin/env bash` to `#!/bin/bash` in this PR the license check will fail.

simonjbeaumont: Sorry, I misunderstood. I see now that the license check expects this shebang in all the shell scripts.

Let's just take this change and move forward. Thanks!

simonjbeaumont (approved):

.github/workflows/pull_request.yml
```diff
@@ -9,4 +9,6 @@ jobs:
     name: Checks
     uses: apple/swift-nio/.github/workflows/reusable_pull_request.yml@main
     with:
-      benchmarks_linux_enabled: false
\ No newline at end of file
+      benchmarks_linux_enabled: false
+      unacceptable_language_check_enabled: false
```
simonjbeaumont, line 13: Are we doing opt-out? Seems like it might be pretty bad to do that since we'll introduce CI breakages for adopters of the reusable workflow when we add new things that they don't want and/or can't use for some reason.

czechboy0: Adopters can lock to a specific commit hash, I believe, instead of depending on `main`.

I think it's fair to discuss which steps are opt-out (those that we expect/want all repos to have) and opt-in (optional steps that only some repos need).

PR #593, opened 2024-07-16
Initial message fix

akbashev:
Last week I've been checking bidirectional examples with latest Xcode 16 beta 3 and realised it uses HBv2, which is supported only by macOS 14+, and it's not perfect. After downgrading to HBv1—same problem with returning stream until some bytes are sent appeared. So, this PR fixes two things:

- Moved to Vapor for two reasons: 1) anyway can't use HBv2 now and 2) it could be more familiar to others;
- Moved back to URLSession and added `initial message` hack.

It's working for now. I'll be checking from time to time status of Swift Server tools, maybe we can improve it in the future.

### DIFF
--- a/Examples/bidirectional-event-streams-client-example/Package.swift
+++ b/Examples/bidirectional-event-streams-client-example/Package.swift
@@ -20,14 +20,14 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-openapi-generator", from: "1.0.0"),
         .package(url: "https://github.com/apple/swift-openapi-runtime", from: "1.2.0"),
-        .package(url: "https://github.com/swift-server/swift-openapi-async-http-client", from: "1.0.0"),
+        .package(url: "https://github.com/apple/swift-openapi-urlsession", from: "1.0.0"),
     ],
     targets: [
         .executableTarget(
             name: "BidirectionalEventStreamsClient",
             dependencies: [
                 .product(name: "OpenAPIRuntime", package: "swift-openapi-runtime"),
-                .product(name: "OpenAPIAsyncHTTPClient", package: "swift-openapi-async-http-client"),
+                .product(name: "OpenAPIURLSession", package: "swift-openapi-urlsession"),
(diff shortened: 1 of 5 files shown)


Examples/bidirectional-event-streams-client-example/Package.swift
```diff
@@ -20,14 +20,14 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-openapi-generator", from: "1.0.0"),
         .package(url: "https://github.com/apple/swift-openapi-runtime", from: "1.2.0"),
-        .package(url: "https://github.com/swift-server/swift-openapi-async-http-client", from: "1.0.0"),
+        .package(url: "https://github.com/apple/swift-openapi-urlsession", from: "1.0.0"),
```
czechboy0, line 23: I think it's fine to keep it using AHC? Since it doesn't require the initial message hack?

akbashev: @czechboy0 we could, it doesn't matter in the end as there something with transport happening (both for vapor and hbv1), so we anyway need initial message

czechboy0: > there something with transport happening (both for vapor and hbv1)

Can you elaborate what you're seeing? Are you saying even when using AHC, you still don't get the stream established without an initial message?

akbashev: > Can you elaborate what you're seeing? Are you saying even when using AHC, you still don't get the stream established without an initial message?

Yep, exactly. I will try to investigate, but for now think we can just update the example.

czechboy0 (changes requested):
@akbashev if you have some time, please investigate where this issue might be, I'd prefer not to switch the example until we have a good understanding of why it doesn't work

akbashev:
@czechboy0 will check, last time I've tried—it was a bit hard, cause it's hard to debug "not happening" events.

Issue #586, opened 2024-07-09
Warnings due to `@frozen` attribute on non-public enums with Xcode 16 Beta 3


ojun9:
### Description

Using Xcode 16 Beta 3, the generated `Type.swift` file now shows multiple warnings due to the `@frozen` attribute being applied to non-public enums. The `Type.swift` file is generated with non-public enums that include the `@frozen` attribute. For example:

```swift
@frozen internal enum Output: Sendable, Hashable { /* ... */ }
```

This results in the following warning:

```shell
@frozen has no effect on non-public enums
```

### Reproduction

To reproduce the issue, it is necessary to generate the `Type.swift` file.

**openapi-generator-config.yaml**
```yaml
generate:
  - types
  - client
```

### Package version(s)

`1.2.1`

### Expected behavior

The `@frozen` attribute should not be applied to non-public enums to avoid unnecessary warnings.

### Environment

- Swift version Info: 
  - Apple Swift version 6.0 (swiftlang-6.0.0.5.15 clang-1600.0.22.6)
- Xcode version Info: 
  - Xcode 16 Beta 3

### Additional information

This issue arises from a new warning introduced in Xcode 16 Beta 3.

czechboy0:
Thanks for reporting this, @ojun9! I think the fix here is what you suggested - only generate the frozen attribute when the access modifier is public or package. Would you consider opening a PR for that?

ojun9:
Thank you for the feedback! 
I’ll work on the fix over the weekend and open a PR to ensure the frozen attribute is only generated when the access modifier is public or package.

czechboy0:
Looks like #587 is already fixing this issue

ojun9:
@czechboy0

Since there were no changes in the PR https://github.com/apple/swift-openapi-generator/pull/587, I created a new PR https://github.com/apple/swift-openapi-generator/pull/595. Please let me know if there are any issues!

czechboy0:
Fixed and released in 1.3.0: https://github.com/apple/swift-openapi-generator/releases/tag/1.3.0

PR #595, opened 2024-07-21
Add the `frozen` keyword only to enums with the `public` or `package` modifiers

ojun9:
### Motivation
As reported in https://github.com/apple/swift-openapi-generator/issues/586, using Xcode 16 Beta 3, the generated Type.swift file now shows multiple warnings due to the `@frozen` attribute being applied to non-public enums. 

### Modifications
In order to resolve the above issue, I have modified the code to annotate the `@frozen` keyword only on `enum` with the `public` or `package` modifiers.

### Result
No warnings are issued when the `@frozen` keyword is annotated on enums with the `public` or `package` modifiers.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Renderer/TextBasedRenderer.swift
+++ b/Sources/_OpenAPIGeneratorCore/Renderer/TextBasedRenderer.swift
@@ -898,7 +898,7 @@ extension TextBasedRenderer {
         renderer.renderExpression(expression)
         return renderer.renderedContents()
     }
-    
+
     /// Checks if the given enum description requires a @frozen annotation.
     /// - Parameter enumDesc: The enum description to check.
     /// - Returns: A boolean value indicating whether the enum description requires a @frozen annotation.
--- a/Sources/_OpenAPIGeneratorCore/Renderer/TextBasedRenderer.swift
+++ b/Sources/_OpenAPIGeneratorCore/Renderer/TextBasedRenderer.swift
@@ -645,12 +645,7 @@ struct TextBasedRenderer: RendererProtocol {
 
     /// Renders the specified enum declaration.
     func renderEnum(_ enumDesc: EnumDescription) {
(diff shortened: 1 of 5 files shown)


Sources/_OpenAPIGeneratorCore/Renderer/TextBasedRenderer.swift
```diff
@@ -645,7 +645,12 @@ struct TextBasedRenderer: RendererProtocol {
 
     /// Renders the specified enum declaration.
     func renderEnum(_ enumDesc: EnumDescription) {
-        if enumDesc.isFrozen {
+        func shouldAnnotateAsFrozen(_ enumDesc: EnumDescription) -> Bool {
```
czechboy0, line 648: Could you break this method out and add unit tests for it, and update the reference tests as well please?

czechboy0 (changes requested):

Sources/_OpenAPIGeneratorCore/Renderer/TextBasedRenderer.swift
ojun9: I've broken out the method into an extension and added unit tests for it. I've also updated the reference tests accordingly. Please review and let me know if any further changes are needed!

czechboy0 (approved):
Great, thank you!

czechboy0:

czechboy0:

czechboy0:
For the 5.9 and 5.9.0 pipelines, we're running into the deprecation warning being escalated to an error. We already disabled warnings-as-errors on most other pipelines, please also comment out this line in the two files for 5.9 and 5.9.0, then the CI should pass.

https://github.com/apple/swift-openapi-generator/blob/285ebba6300019d5eabcde15173809d098ac14f4/docker/docker-compose.2204.59.yaml#L14

czechboy0:

czechboy0 (approved):

PR #597, opened 2024-07-22
Fix multipart + additionalProperties + string support

### LINKED ISSUE #596: multipart/form-data request body schemas with additionalProperties are a bit confusing to use, also doesn't compile
### Description

OpenAPI allows using the following construct to represent a `[String: String]` dictionary:

```
type: object
additionalProperties:
  type: string
```

However, this doesn't work correctly with multipart/form-data

### Reproduction

Considering the following code...

Schema request body:

```yaml
      requestBody:
        required: true
        content:
          multipart/form-data:
            schema:
              $ref: '#/components/schemas/JobParameters'
```

Schema components:

```yaml
components:
  schemas:
    JobParameters:
      type: object
      additionalProperties:
        type: string
    JobParameters2:
      type: object
      additionalProperties:
        type: string
```

This generated the following Swift code:

```swift
            /// - Remark: Generated from `#/paths/job/{jobName}/buildWithParameters/POST/requestBody`.
            @frozen public enum Body: Sendable, Hashable {
                /// - Remark: Generated from `#/paths/job/{jobName}/buildWithParameters/POST/requestBody/content/multipart\/form-data`.
                case multipartForm(OpenAPIRuntime.MultipartBody<Components.Schemas.JobParameters>)
            }
```

```swift
/// Types generated from the components section of the OpenAPI document.
public enum Components {
    /// Types generated from the `#/components/schemas` section of the OpenAPI document.
    public enum Schemas {
        /// - Remark: Generated from `#/components/schemas/JobParameters`.
        @frozen public enum JobParameters: Sendable, Hashable {
            case additionalProperties(OpenAPIRuntime.MultipartDynamicallyNamedPart<Swift.String>)
        }
        /// - Remark: Generated from `#/components/schemas/JobParameters2`.
        public struct JobParameters2: Codable, Hashable, Sendable {
            /// A container of undocumented properties.
            public var additionalProperties: [String: Swift.String]
            /// Creates a new `JobParameters2`.
            ///
            /// - Parameters:
            ///   - additionalProperties: A container of undocumented properties.
            public init(additionalProperties: [String: Swift.String] = .init()) {
                self.additionalProperties = additionalProperties
            }
            public init(from decoder: any Decoder) throws {
                additionalProperties = try decoder.decodeAdditionalProperties(knownKeys: [])
            }
            public func encode(to encoder: any Encoder) throws {
                try encoder.encodeAdditionalProperties(additionalProperties)
            }
        }
        ...
```

### Package version(s)

swift-openapi-generator 1.2.1
swift-openapi-runtime 1.4.0

### Expected behavior

The inclusion of JobParameters2 is merely for illustration. That's what I would have expected JobParameters to look like.

It was initially quite unclear to me how to pass a `[String: String]` into this construct, as I would have expected something like this at the call site:

```swift
body: .multipartForm(.init(additionalProperties: parameters))
```

but you actually need to use this:

```swift
body: .multipartForm(.init(parameters.map {
            .additionalProperties(.init(payload: $0.value, name: $0.key))
        }))
```

However, the final problem is this:

```swift
                        encoding: { part in
                            switch part {
                            case let .additionalProperties(wrapped):
                                var headerFields: HTTPTypes.HTTPFields = .init()
                                let value = wrapped.payload
                                let body = try converter.setRequiredRequestBodyAsBinary(
                                    value,
                                    headerFields: &headerFields,
                                    contentType: "text/plain"
                                )
                                return .init(
                                    name: wrapped.name,
                                    filename: wrapped.filename,
                                    headerFields: headerFields,
                                    body: body
                                )
                            }
                        }
```

The generated code doesn't compile, as `value` is a `String`, but an `HTTPBody` is expected there.

I was able to work around the compile error by adding to my target:

```swift
@_spi(Generated) import OpenAPIRuntime
extension Converter {
    func setRequiredRequestBodyAsBinary(_ value: String, headerFields: inout HTTPFields, contentType: String) throws -> HTTPBody {
        let body = HTTPBody(value)
        headerFields[.contentType] = contentType
        if case let .known(length) = body.length { headerFields[.contentLength] = String(length) }
        return body
    }
}
```

That seems to work at runtime as well.

### Environment

N/A

### Additional information

_No response_

czechboy0:
### Motivation

Fixes #596.

The bug was computing the `TypeUsage` from the raw schema, before resolving it with multipart semantics.

Covers using the multipart content type with additional properties of a concrete type (string).

This lead to the generator producing half of the code that expected a HTTPBody (a streaming byte type) and another half expected a String, leading to a build failure.

### Modifications

First resolve the schema, then calculate the `TypeUsage` from the resolved schema.

### Result

The generator correctly generates this as `HTTPBody` everywhere now.

### Test Plan

Added a unit test to cover this case.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/Multipart/MultipartContentInspector.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Multipart/MultipartContentInspector.swift
@@ -148,14 +148,6 @@ extension FileTranslator {
         case .disallowed: break
         case .allowed: parts.append(.undocumented)
         case .typed(let schema):
-            let typeUsage = try typeAssigner.typeUsage(
-                usingNamingHint: Constants.AdditionalProperties.variableName,
-                withSchema: .b(schema),
-                components: components,
-                inParent: typeName
-            )!
-            // The unwrap is safe, the method only returns nil when the input schema is nil.
-            let typeName = typeUsage.typeName
             guard
                 let (info, resolvedSchema) = try parseMultipartPartInfo(
                     schema: schema,
@@ -167,7 +159,15 @@ extension FileTranslator {
                     message: "Failed to parse multipart info for additionalProperties in \(typeName.description)."
                 )
             }
-            parts.append(.otherDynamicallyNamed(.init(typeName: typeName, partInfo: info, schema: resolvedSchema)))
+            let partTypeUsage = try typeAssigner.typeUsage(
+                usingNamingHint: Constants.AdditionalProperties.variableName,
+                withSchema: .b(resolvedSchema),
+                components: components,
+                inParent: typeName
+            )!
+            // The unwrap is safe, the method only returns nil when the input schema is nil.
+            let partTypeName = partTypeUsage.typeName
+            parts.append(.otherDynamicallyNamed(.init(typeName: partTypeName, partInfo: info, schema: resolvedSchema)))
         case .any: parts.append(.otherRaw)
         }
         let requirements = try parseMultipartRequirements(
--- a/Tests/OpenAPIGeneratorCoreTests/StructureHelpers.swift
+++ b/Tests/OpenAPIGeneratorCoreTests/StructureHelpers.swift
@@ -154,7 +154,7 @@ extension BindingKind {
     }
 }
 
-extension Expression {
+extension _OpenAPIGeneratorCore.Expression {
     var info: ExprInfo {
         switch self {
         case .literal(let value): return .init(name: value.name, kind: .literal)
--- a/Tests/OpenAPIGeneratorReferenceTests/FileBasedReferenceTests.swift
+++ b/Tests/OpenAPIGeneratorReferenceTests/FileBasedReferenceTests.swift
@@ -151,14 +151,17 @@ extension FileBasedReferenceTests {
         )
     }
 
-    private func temporaryDirectory(fileManager: FileManager = .default) throws -> URL {
-        let directoryURL = fileManager.temporaryDirectory.appendingPathComponent(UUID().uuidString, isDirectory: true)
-        try fileManager.createDirectory(at: directoryURL, withIntermediateDirectories: true)
+    private func temporaryDirectory() throws -> URL {
+        let directoryURL = FileManager.default.temporaryDirectory.appendingPathComponent(
+            UUID().uuidString,
+            isDirectory: true
+        )
+        try FileManager.default.createDirectory(at: directoryURL, withIntermediateDirectories: true)
         addTeardownBlock {
             do {
-                if fileManager.fileExists(atPath: directoryURL.path) {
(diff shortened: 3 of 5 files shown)


Sources/_OpenAPIGeneratorCore/Translator/Multipart/MultipartContentInspector.swift
```diff
@@ -167,7 +159,15 @@ extension FileTranslator {
...
-            parts.append(.otherDynamicallyNamed(.init(typeName: typeName, partInfo: info, schema: resolvedSchema)))
+            let partTypeUsage = try typeAssigner.typeUsage(
+                usingNamingHint: Constants.AdditionalProperties.variableName,
+                withSchema: .b(resolvedSchema),
+                components: components,
+                inParent: typeName
+            )!
+            // The unwrap is safe, the method only returns nil when the input schema is nil.
+            let partTypeName = partTypeUsage.typeName
+            parts.append(.otherDynamicallyNamed(.init(typeName: partTypeName, partInfo: info, schema: resolvedSchema)))
```
czechboy0, line 170: Just moved the code down and renamed the local variable for clarity (it was shadowing one higher up).

Tests/OpenAPIGeneratorCoreTests/StructureHelpers.swift
```diff
@@ -154,7 +154,7 @@ extension BindingKind {
     }
 }
 
-extension Expression {
+extension _OpenAPIGeneratorCore.Expression {
```
czechboy0, line 157: Foundation added an [Expression type](https://github.com/apple/swift-foundation/blob/c0a485e336e5e840574ae8a8429904ba587812fc/Sources/FoundationEssentials/Predicate/Expression.swift#L14), leading to a build error due to an ambiguous type.

Tests/OpenAPIGeneratorReferenceTests/FileBasedReferenceTests.swift
```diff
@@ -151,14 +151,17 @@ extension FileBasedReferenceTests {
...
+        )
+        try FileManager.default.createDirectory(at: directoryURL, withIntermediateDirectories: true)
         addTeardownBlock {
             do {
-                if fileManager.fileExists(atPath: directoryURL.path) {
-                    try fileManager.removeItem(at: directoryURL)
-                    XCTAssertFalse(fileManager.fileExists(atPath: directoryURL.path))
+                if FileManager.default.fileExists(atPath: directoryURL.path) {
+                    try FileManager.default.removeItem(at: directoryURL)
+                    XCTAssertFalse(FileManager.default.fileExists(atPath: directoryURL.path))
```
czechboy0, line 164: Sendability fix, FileManager isn't Sendable so just use the singleton in all places for now. Nobody was customizing the parameter anyway, so a no-op.

simonjbeaumont (approved):
Nice patch, thanks @czechboy0!

PR #604, opened 2024-07-24
ci: Finish migration to Github Actions

FranzBusch:
# Motivation

I made more progress on the reusable workflows in NIO so it's even easier to adopt them.

# Modification

This PR fixes up the renamed workflow references and adds unit tests and cxx-interop checks.

# Result

Green GH actions CI


FranzBusch:
@simonjbeaumont @czechboy0 Would appreciate some eyes on the failing format check

czechboy0:
> @simonjbeaumont @czechboy0 Would appreciate some eyes on the failing format check

I suspect you want to add `Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources` to the ignore list for swift-format.

czechboy0:
See here: https://github.com/apple/swift-openapi-generator/blob/99859083e53912612b56b4b8713a971c20dab3ef/scripts/run-swift-format.sh#L34

czechboy0:
@FranzBusch It's still failing, seems the `Generated` entry isn't working, it's meant to match the word `Generated` anywhere in the path. The affected paths here are e.g. `Examples/manual-generation-generator-cli-example/Sources/ManualGeneratorInvocationClient/Generated/Client.swift`

czechboy0:
Ok, soundness is failing now:
```
** ERROR: Unsupported file extension for file (exclude or update this script): .swiftformatignore
```

I think it just requires you to add `.swiftformatignore` to the ignore list in `scripts/check-license-headers.sh`

FranzBusch:
@czechboy0 @simonjbeaumont This passes the format check now. Please take a look at the changes. I also added the integration tests and example tests here. Once we have this merged the only thing that is missing is
- Turning on the license check and turning off the old CI
- Adding the compatibility test

simonjbeaumont:
@FranzBusch noticed this got reopened. Should I assume it's still a WIP for now and wait for a ping from you to review?

FranzBusch:
@simonjbeaumont @czechboy0 This PR is ready to review. The CI is all passing. I only had to skip the unacceptable language mode check and the shell check steps. I leave this up to you to re-enable them. The language mode depends on the old soundness CI to remove the text file with the words and the shell check depends on fixing up a few scripts. I would also encourage you to look into the examples CI and if you can reduce build times by using the same build dirs. You are currently rebuilding the same dependencies multiple times.

FranzBusch:
The compatibility test is also missing but similarly to the above could you please set that up in a follow up. It would be a good exercise to see if the reusable workflows give you want you need.

.github/workflows/pull_request.yml
```diff
@@ -1,12 +1,50 @@
...
+            linux_5_9_arguments_override: "--explicit-target-dependency-import-check error"
+            linux_5_10_arguments_override: "--explicit-target-dependency-import-check error"
+            linux_nightly_6_0_arguments_override: "--explicit-target-dependency-import-check error"
+            linux_nightly_main_arguments_override: "--explicit-target-dependency-import-check error"
+
+    cxx-interop:
+        name: Cxx interop
+        uses: apple/swift-nio/.github/workflows/cxx_interop.yml@main
+        with:
+            linux_5_8_enabled: false
```
simonjbeaumont, line 30: This wasn't something we had before. Do we want it in all projects?

.github/workflows/pull_request.yml
```diff
@@ -1,12 +1,50 @@
...
+            matrix_linux_5_8_enabled: false
+            matrix_linux_command: "apt-get update -y -q && apt-get install -y -q jq && ./scripts/run-integration-test.sh"
+
+    example-tests:
+        name: Example Tests
+        uses: apple/swift-nio/.github/workflows/swift_matrix.yml@main
+        with:
+            name: "Example tests"
+            matrix_linux_5_8_enabled: false
+            matrix_linux_command: "./scripts/test-examples.sh"
```
simonjbeaumont, line 46: Starting a thread here to discuss your comment you made at the PR level about the example tests.

The script already uses a shared directory for cache and scratch, which, to be fair, opens it up to not catching a missing dependency in an example that is run later than an earlier one that had it.

You'd be surprised just how long it took if we _didn't_ do the current level of sharing :)

.github/workflows/pull_request.yml
```diff
@@ -1,12 +1,50 @@
...
+
+    swift-6-language-mode:
+        name: Swift 6 Language Mode
+        uses: apple/swift-nio/.github/workflows/swift_6_language_mode.yml@main
```
simonjbeaumont, line 50: We've made no effort to switch to Swift 6 language mode in this project _yet(!)_. Can we have this one be marked as skipped for now?

.github/workflows/pull_request.yml
```diff
@@ -1,12 +1,50 @@
...
-    with:
-      benchmarks_linux_enabled: false
\ No newline at end of file
+    soundness:
+        name: Soundness
+        uses: apple/swift-nio/.github/workflows/soundness.yml@main
+        with:
+            license_header_check_project_name: "SwiftOpenAPIGenerator"
+            unacceptable_language_check_enabled: false
+            shell_check_enabled: false
```
simonjbeaumont, line 14: I think it's fair to leave the `unacceptable_language_check` pipeline disabled because we need to do a hot-swap with the other pipeline.

But for the shellcheck pipeline, what's the issue. I ask because this project already _has_ a shellcheck pipeline, which is passing. So why can't this one be passing?

simonjbeaumont (commented):
Thanks for all the work here @FranzBusch! Left some comments.

.github/workflows/pull_request.yml
FranzBusch: This is only needed for projects that do not `yet` support the last three versions. In fact with the Swift 6 release this will go away and OpenAPI will also support the last 3 versions.

.github/workflows/pull_request.yml
FranzBusch: Interesting. Just from looking at the logs it does seem to be building the same things multiple times but I trust you on this.

.github/workflows/pull_request.yml
FranzBusch: This is up to you but I would encourage you to add it as a non-required workflow. We have the same in NIO for two reasons:
1. Remind us that we have work to do here
2. Make sure that we at least exercise building with Swift 6 language mode

.github/workflows/pull_request.yml
FranzBusch: When it was active there were a few shell check issues that it caught. Do you know why there would be different warnings caught?

```
In scripts/check-license-headers.sh line 90:
    swift) EXPECTED_FILE_HEADER=$(sed -e 's|@@|//|g' <<<"${EXPECTED_FILE_HEADER_TEMPLATE}") ;;
                                  ^-- SC2001 (style): See if you can use ${variable//search/replace} instead.


In scripts/check-license-headers.sh line 91:
    yml) EXPECTED_FILE_HEADER=$(sed -e 's|@@|##|g' <<<"${EXPECTED_FILE_HEADER_TEMPLATE}") ;;
                                ^-- SC2001 (style): See if you can use ${variable//search/replace} instead.


In scripts/check-license-headers.sh line 92:
    sh) EXPECTED_FILE_HEADER=$(cat <(echo '#!/bin/bash') <(sed -e 's|@@|##|g' <<<"${EXPECTED_FILE_HEADER_TEMPLATE}")) ;;
                                                           ^-- SC2001 (style): See if you can use ${variable//search/replace} instead.
```

.github/workflows/pull_request.yml
simonjbeaumont: Sorry, the question wasn't about disabling 5.8, it was about whether we needed a C++ interop pipeline in all projects.

FranzBusch: I feel like it is worth having right now since it adds very little overhead. We can always remove it in the future

simonjbeaumont: I don't think it really makes sense for this project, which is mainly a dev-tool.

FranzBusch:
@simonjbeaumont Gentle ping. I think we can merge this. It should replicate everything that Jenkins is currently providing.

simonjbeaumont:
I'll take this, this week. Sorry for letting it linger.

.github/workflows/pull_request.yml
```diff
@@ -1,12 +1,68 @@
...
+              SWIFT_OPENAPI_COMPATIBILITY_TEST_ENABLE: "true"
+              SWIFT_OPENAPI_COMPATIBILITY_TEST_SKIP_BUILD: "true"
+              SWIFT_OPENAPI_COMPATIBILITY_TEST_FILTER: OpenAPIGeneratorReferenceTests.CompatibilityTest
+              SWIFT_OPENAPI_COMPATIBILITY_TEST_PARALLEL_CODEGEN: "true"
+              SWIFT_OPENAPI_COMPATIBILITY_TEST_NUM_BUILD_JOBS: 1
+            run: swift test --filter ${SWIFT_OPENAPI_COMPATIBILITY_TEST_FILTER}
+
+    example-packages:
+        name: Example packages
+        uses: apple/swift-nio/.github/workflows/swift_matrix.yml@main
```
simonjbeaumont, line 60: @czechboy0 @FranzBusch do you think we need to do this for all the Swift versions?

simonjbeaumont (commented):
@FranzBusch this is now at a state where I'm happy with it.

I realise this was initially your PR, but could you take a look.

@czechboy0 once @FranzBusch and I are aligned. I think you should give it a look too :)

.github/workflows/pull_request.yml
simonjbeaumont: It's also still my opinion that we should axe this until we're ready to tackle it. We have a bunch of contributors filing PRs and I don't really want them put off. WDYT, @czechboy0 ?

czechboy0: Since supporting both language modes in the same codebase is unlikely to be worth our time, I think we should only care about supporting Swift 6 compiler + Swift 5 language mode + all the "strict" flags.

I do not think we should actually care about Swift 6 language mode until we can switch our `tools-version` to >=6.0.

.github/workflows/pull_request.yml
simonjbeaumont: Ok. Maybe not on the nightly though? That one takes the most time and significant slow down on getting a green PR.

czechboy0: You mean that we'd only run this on the nightly? Sure, we don't have to run it on PRs.

simonjbeaumont: No, I meant we could run it for all Swift versions except the nightly.

.github/workflows/pull_request.yml
simonjbeaumont: I've left this in the YAML but disabled it for now.

.github/workflows/pull_request.yml
simonjbeaumont: I've changed the matrix jobs to run without the nightly Swift version for pull requests and added a scheduled job that includes the nightly Swift version.

czechboy0 (approved):

simonjbeaumont (approved):

simonjbeaumont:
@FranzBusch I'm happy with this now and so is Honza. Can you do a final pass with your Github Actions hat on and then hit merge? :)

.github/workflows/pull_request.yml
```diff
@@ -1,12 +1,71 @@
...
+            shell_check_enabled: true
+            unacceptable_language_check_enabled: true
+
+    unit-tests:
+        name: Unit tests
+        uses: apple/swift-nio/.github/workflows/unit_tests.yml@main
+        with:
+            linux_5_8_enabled: false
+            linux_5_9_arguments_override: "--explicit-target-dependency-import-check error"
+            linux_5_10_arguments_override: "--explicit-target-dependency-import-check error"
```
FranzBusch, line 27: Should we track enabling warnings as errors for those pipelines at some point?

.github/workflows/pull_request.yml
```diff
@@ -1,12 +1,71 @@
...
+        uses: apple/swift-nio/.github/workflows/soundness.yml@main
+        with:
+            api_breakage_check_enabled: true
+            broken_symlink_check_enabled: true
+            docs_check_enabled: true
+            format_check_enabled: true
+            license_header_check_enabled: true
+            license_header_check_project_name: "SwiftOpenAPIGenerator"
+            shell_check_enabled: true
+            unacceptable_language_check_enabled: true
```
FranzBusch, line 19: Why are we enabling them explicitly? The default is `true`

.github/workflows/pull_request.yml
```diff
@@ -1,12 +1,71 @@
...
+            linux_nightly_main_enabled: false
+
+    integration-test:
+        name: Integration test
+        uses: apple/swift-nio/.github/workflows/swift_matrix.yml@main
+        with:
+            name: "Integration test"
+            matrix_linux_command: "apt-get update -yq && apt-get install -yq jq && ./scripts/run-integration-test.sh"
+            matrix_linux_5_8_enabled: false
+            matrix_linux_nightly_main_enabled: false
```
FranzBusch, line 38: Why not run them on the nightly main?

.github/workflows/pull_request.yml
```diff
@@ -1,12 +1,71 @@
...
+            run: swift test --filter ${SWIFT_OPENAPI_COMPATIBILITY_TEST_FILTER}
+
+    example-packages:
+        name: Example packages
+        uses: apple/swift-nio/.github/workflows/swift_matrix.yml@main
+        with:
+            name: "Example packages"
+            matrix_linux_command: "./scripts/test-examples.sh"
+            matrix_linux_5_8_enabled: false
+            matrix_linux_nightly_main_enabled: false
```
FranzBusch, line 66: Same here. Why not nightly?

docker/docker-compose.yaml
```diff
@@ -26,25 +26,25 @@ services:
```
FranzBusch, line 26: Do you intend to keep the docker files around?

.github/workflows/pull_request.yml
simonjbeaumont: Because it's more visible. The fact that some are default to true and some to false is confusing IMO. I like to be able to see what tests we're running in the YAML.

.github/workflows/pull_request.yml
simonjbeaumont: More than happy to. But that's over-and-above what we currently had in this project. We can track it separately.

.github/workflows/pull_request.yml
simonjbeaumont: Because it's materially slower and not blocking a PR. Note that this PR does use the nightly Swift versions in the scheduled workflow so we'll still be alerted. I'd really like to only have PR pipelines that are gating PR merges, especially when we have external contributors.

.github/workflows/pull_request.yml
simonjbeaumont: Answered above.

docker/docker-compose.yaml
simonjbeaumont: I hadn't actually considered removing them altogether but I guess the only downside to doing so is that the existing CI will immediately fail because it can't find them. Maybe that's fine since we want it turned off anyway.

@czechboy0 any objections to me just deleting all the docker-based CI files?

.github/workflows/pull_request.yml
FranzBusch: I am fine with explicitly stating but just to be clear no check is defaulting to `false`.

.github/workflows/pull_request.yml
FranzBusch: I am surprised. Nightly main is slower than nightly 6.0? If so we should report that since this hints at a compiler speed regression.

FranzBusch:
I can't approve since I am the author but LGTM!

docker/docker-compose.yaml
czechboy0: No, let's get us moved over fully.

.github/workflows/pull_request.yml
simonjbeaumont: No no. It's slower because it's far less likely to have the image available.

What's really annoying (but I realise a Github Action limitation) is that this will still do a bunch of the warmup anyway because the `if: false` is only at the step-level.

I have some ideas on how we can make that better though, if you're open to them.

docker/docker-compose.yaml
FranzBusch: You can turn off the web hooks yourself in the repo settings.

.github/workflows/pull_request.yml
simonjbeaumont: OK, my mistake. Thanks for confirming :)

.github/workflows/pull_request.yml
FranzBusch: I have some ideas here as well. We could dynamically generate the matrix instead of having it hard coded like now.

Issue #180, opened 2023-08-09
Enforce error diagnostics by aborting execution


czechboy0:
I vote we keep it; maybe even rename it fatal. WDYT? Could you open an issue and we can converge on a stance? Even if we end up removing it, we can close the loop on this.

_Originally posted by @simonjbeaumont in https://github.com/apple/swift-openapi-generator/pull/178#discussion_r1288038440_
            
See the conversation above.

Today, we have the issue that `.error` diagnostics don't terminate execution. We should decide whether to remove that error severity, and simply use error throwing to propagate fatal issues, or to somehow make emitting an error diagnostic become fatal.

czechboy0:
@simonjbeaumont If we keep it, we have to find a way to propagate the signal that an error diagnostic was emitted to the generator. I'd like to avoid using `fatalError` (as that produces a crash log), and ideally have something testable (so `exit(1)` also isn't great).

But I wonder if we're over-engineering it. Essentially, the place that emits the error diagnostic wants to terminate execution anyway, which is what error throwing is for in Swift. Otherwise we need to find a way that some code in the generator emits an error diagnostic, which gets added to the diagnostic collector, and the signal that it was fatal is propagated back to the caller. Maybe the call to `emit` can just throw if the severity is `.error`? That'd work, I think.

simonjbeaumont:
> Maybe the call to `emit` can just throw if the severity is `.error`? That'd work, I think.

That works for me.

czechboy0:
Ugh, I forgot that `emit` isn't throwing today. We can change that though.

czechboy0:
@arthurcro, sure, go ahead! 🙏

arthurcro:
Hi! :wave: I had a chance to look at implementing this change. 
If my understanding is correct we want to change `DiagnosticCollector.emit(_:)` to be throwing. I am wondering if this is the best approach to propagate fatal issues.

First, changing the `DiagnosticCollector` to:

```swift
public protocol DiagnosticCollector {
  /// Submits a diagnostic to the collector.
  /// - Parameter diagnostic: The diagnostic to submit.
  /// - Throws: Throws when the diagnostic severity is `error`.
  func emit(_ diagnostic: Diagnostic) throws
}
```

would be an API breaknig chnge. 

Further, although it would be mentioned in the documentation of `DiagnosticCollector` there is no way to enforce that a error diagnostic would make  all the implementations of `DiagnosticCollector`  actually throw an error which mght be misleading.

[...]

PR #607, opened 2024-08-02
Enforce error diagnostics by aborting execution

PARAIPAN9:
### Motivation

- Fixes https://github.com/apple/swift-openapi-generator/issues/180

### Modifications

- Create `ErrorThrowingDiagnosticCollector` wrapper collector.

### Result

- The `ErrorThrowingDiagnosticCollector` throws an error when a diagnostic with severity `.error` is emited.

### Test Plan

- Make sure all tests pass and add additional test.


PARAIPAN9:
I have a question because I'm a bit confused about what to do after the creation of the wrapper. Should we use the wrapper in place of the extensions from DiagnosticCollector?

czechboy0:
I think `DiagnosticCollector`'s `emit` method should become throwing, and then create a new concrete collector implementation that looks something like this:

```swift
struct ErrorThrowingDiagnosticCollector: DiagnosticCollector {
    var upstream: any DiagnosticCollector

    func emit(_ diagnostic: Diagnostic) throws {
        upstream.emit(diagnostic)
        if diagnostic.severity == .error { throw diagnostic }
    }
}
```

And then use this collector by default, wrapping today's collector.

PARAIPAN9:
> I think `DiagnosticCollector`'s `emit` method should become throwing, and then create a new concrete collector implementation that looks something like this:
> 
> ```swift
> struct ErrorThrowingDiagnosticCollector: DiagnosticCollector {
>     var upstream: any DiagnosticCollector
> 
>     func emit(_ diagnostic: Diagnostic) throws {
>         upstream.emit(diagnostic)
>         if diagnostic.severity == .error { throw diagnostic }
>     }
> }
> ```
> 
> And then use this collector by default, wrapping today's collector.

It makes sense, perfect!!

Sources/swift-openapi-generator/GenerateOptions+runGenerator.swift
```diff
@@ -83,7 +83,7 @@ extension _GenerateOptions {
             try finalizeDiagnostics()
         } catch let error as Diagnostic {
             // Emit our nice Diagnostics message instead of relying on ArgumentParser output.
-            diagnostics.emit(error)
+            try ErrorThrowingDiagnosticCollector(upstream: diagnostics).emit(error)
```
czechboy0, line 86: We should only create `ErrorThrowingDiagnosticCollector` in one place, not in the translate methods (those will already get a `DiagnosticCollector` value, which will already have this collector).

Just update the block around line 45 to something like:

```swift
let innerDiagnostics: any DiagnosticCollector & Sendable
if let diagnosticsOutputPath {
    let _diagnostics = _YamlFileDiagnosticsCollector(url: diagnosticsOutputPath)
    finalizeDiagnostics = _diagnostics.finalize
    innerDiagnostics = _diagnostics
} else {
    innerDiagnostics = StdErrPrintingDiagnosticCollector()
    finalizeDiagnostics = {}
}
let diagnostics = ErrorThrowingDiagnosticCollector(upstream: innerDiagnostics)
```

And please remove all the other places where you're creating `ErrorThrowingDiagnosticCollector`. The only changes in the individual translate* files should be the adaptation to the `emit` method now throwing.

Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
```diff
@@ -110,7 +110,9 @@ func makeGeneratorPipeline(
     }
     let validateDoc = { (doc: OpenAPI.Document) -> OpenAPI.Document in
         let validationDiagnostics = try validator(doc, config)
-        for diagnostic in validationDiagnostics { diagnostics.emit(diagnostic) }
+        for diagnostic in validationDiagnostics {
+            try ErrorThrowingDiagnosticCollector(upstream: diagnostics).emit(diagnostic)
```
czechboy0, line 114: The `diagnostics: any DiagnosticCollector` (line 104) should already be a `ErrorThrowingDiagnosticCollector`, so this shouldn't be necessary (see my other comment at the bottom).

Sources/_OpenAPIGeneratorCore/Diagnostics.swift
```diff
@@ -165,7 +165,33 @@ public protocol DiagnosticCollector {
 
     /// Submits a diagnostic to the collector.
     /// - Parameter diagnostic: The diagnostic to submit.
```
czechboy0, line 167: Yup these changes look great 🙏

czechboy0:

czechboy0:
Ok the code looks good, would you be able to add one test that constructs the pipeline with a malformed input and ends up throwing the error? Just to verify that it's really working?

PARAIPAN9:
Sure.

czechboy0:

czechboy0:
Got some build failures on Linux:

```
/code/Tests/OpenAPIGeneratorTests/Test_GenerateOptions.swift:38:34: error: cannot call value of non-function type 'String'
        let arguments = [docPath.path(), "--config", configPath.path]
                                 ^   ~~
                                     
/code/Tests/OpenAPIGeneratorTests/Test_GenerateOptions.swift:42:94: error: 'nil' requires a contextual type
            try await generator.runGenerator(outputDirectory: outputDirectory, pluginSource: nil, isDryRun: false)
```

You can fix the first one by using `docPath.path` (the property) instead. For the second, just pass the "build plugin" plugin source case.

czechboy0:

czechboy0:
```
** ERROR: Unsupported file extension for file (exclude or update this script): Tests/OpenAPIGeneratorTests/Resources/Docs/malformed-openapi.yaml
```

You'll need to add the file to the license check ignore list.

czechboy0:
Hmm this actually breaks 5.9.0, because it creates a dependency of a test target on an executable target. https://ci.swiftserver.group/job/swift-openapi-generator-swift590-prb/246/console

To fix this, you'll need to start depending on `_OpenAPIGeneratorCore` instead of `swift-openapi-generator`, and move `_YamlFileDiagnosticsCollector` to `_OpenAPIGeneratorCore`, and create a method called e.g. `preparedDiagnosticsCollector(...) -> DiagnosticCollector & Sendable` that lives in `_OpenAPIGeneratorCore`. Then just test that method, and call the method from the `swift-openapi-generator` code. 

I don't think we'll be able to make the new dependency work otherwise.

Sources/swift-openapi-generator/GenerateOptions+runGenerator.swift
```diff
@@ -42,17 +42,21 @@ extension _GenerateOptions {
...
+                finalizeDiagnostics = {}
+            }
+            innerDiagnostics = _diagnostics
         } else {
-            diagnostics = StdErrPrintingDiagnosticCollector()
+            innerDiagnostics = StdErrPrintingDiagnosticCollector()
             finalizeDiagnostics = {}
         }
-
+        let diagnostics = ErrorThrowingDiagnosticCollector(upstream: innerDiagnostics)
```
czechboy0, line 59: I was thinking you could move all this logic into the prepare... function, so that it can be called from tests and also from here.

The method would take an optional URL, depending on whether the user provided a diagnostic URL or not. But the whole diagnostics collector construction would happen in that method.

PARAIPAN9: Aha, ok.

czechboy0:

Package.swift
```diff
@@ -114,6 +114,14 @@ let package = Package(
             swiftSettings: swiftSettings
         ),
 
+        // Test Target for swift-openapi-generator
+        .testTarget(
+            name: "OpenAPIGeneratorTests",
+            dependencies: ["swift-openapi-generator"],
```
czechboy0, line 120: You also need to add an explicit dependency on ArgumentParser, the 5.9.0 CI pipeline is failing with:

```
/code/Tests/OpenAPIGeneratorTests/Test_GenerateOptions.swift:39: error: undefined reference to '$s14ArgumentParser17ParsableArgumentsPAAE5parseyxSaySSGSgKFZ'
```

czechboy0:

czechboy0:

czechboy0:
@PARAIPAN9 The CI is happy, but I believe we still need to move the remaining diagnostic collection code into the new `prepare...` function, so that it's called from the CLI and also from tests. Right now we're still re-creating the collectors using different code in the CLI vs the tests.

czechboy0 (changes requested):
See comment above.

czechboy0 (approved):
Thank you @PARAIPAN9, looks great now!

PR #618, opened 2024-09-07
Generate enums for server variables

theoriginalbit:
### Motivation

Refer to proposal https://github.com/apple/swift-openapi-generator/pull/629

<details>
<summary>PR description prior to raising proposal</summary>
### Motivation

Recently in a project I was using a spec which defined variables similar to below
```yaml
servers:
  - url: https://{environment}.example.com/api/{version}
    variables:
      environment:
        default: prod
        enum:
          - prod
          - staging
          - dev
      version:
        default: v1
```

The generated code to create the default server URL was easy enough being able to utilise the default parameters
```swift
let serverURL = try Servers.server1()
```

But when I wanted to use a different variable I noticed that the parameter was generated as a string and it didn't expose the other allowed values that were defined in the OpenAPI document. It generated the following code:
```swift
/// Server URLs defined in the OpenAPI document.
internal enum Servers {
    ///
    /// - Parameters:
    ///   - environment:
    ///   - version:
    internal static func server1(
        environment: Swift.String = "prod",
        version: Swift.String = "v1"
    ) throws -> Foundation.URL {
        try Foundation.URL(
            validatingOpenAPIServerURL: "https://{environment}.example.com/api/{version}",
            variables: [
                .init(
                    name: "environment",
                    value: environment,
                    allowedValues: [
                        "prod",
                        "staging",
                        "dev"
                    ]
                ),
                .init(
                    name: "version",
                    value: version
                )
            ]
        )
    }
}
```

This meant usage needed to involve runtime checks whether the supplied variable was valid and if the OpenAPI document were to ever remove an option it could only be discovered at runtime.

```swift
let serverURL = try Servers.server1(environment: "stg") // might be a valid environment, might not
```

Looking into the OpenAPI spec for server templating and the implementation of the extension `URL.init(validatingOpenAPIServerURL:variables:)` I realised that the variables could very easily be represented by an enum in the generated code. By doing so it would also provide a compiler checked way to use a non-default variable.

### Modifications

I have introduced a new set of types translator functions in the file `translateServersVariables.swift` which can create the enum declarations for the variables. If there are no variables defined then no declaration is generated.

Each variable defined in the OpenAPI document is generated as an enum with a case that represents each enum in the document. Each enum is also generated with a static computed property with the name `default` which returns the default value as required by the OpenAPI spec. These individual variable enums are then namespaced according to the server they are applicable for, for example `Server1`, allowing servers to have identically named variables with different enum values. Finally each of the server namespace enums are members of a final namespace, `Variables`, which exists as a member of the pre-existing `Servers` namespace. A truncated example:

```swift
enum Servers { // enum generated prior to this PR
  enum Variables {
    enum Server1 {
      enum VariableName1 {
        // ...
      }
      enum VariableName2 {
        // ...
      }
    }
  }
  static func server1(/* ... */) throws -> Foundation.URL { /* declaration prior to this PR */ }
}
```

To use the new translator functions the `translateServers` function has been modified to call the `translateServersVariables` function and insert the declarations as a member alongside the existing static functions for each of the servers. The `translateServer(index:server:)` function was also edited to make use of the generated variable enums, and the code which generated the string array for `allowedValues` has been removed; runtime validation should no longer be required, as the `rawValue` of a variable enum is the value defined in the OpenAPI document.

### Result

The following spec
```yaml
servers:
  - url: https://{environment}.example.com/api/
    variables:
      environment:
        default: prod
        enum:
          - prod
          - staging
          - dev
```
Would currently generate to the output
```swift
/// Server URLs defined in the OpenAPI document.
internal enum Servers {
    ///
    /// - Parameters:
    ///   - environment:
    internal static func server1(environment: Swift.String = "prod") throws -> Foundation.URL {
        try Foundation.URL(
            validatingOpenAPIServerURL: "https://{environment}.example.com/api/",
            variables: [
                .init(
                    name: "environment",
                    value: environment,
                    allowedValues: [
                        "prod",
                        "staging",
                        "dev"
                    ]
                )
            ]
        )
    }
}
```
But with this PR it would generate to be
```swift
/// Server URLs defined in the OpenAPI document.
internal enum Servers {
    /// Server URL variables defined in the OpenAPI document.
    internal enum Variables {
        /// The variables for Server1 defined in the OpenAPI document.
        internal enum Server1 {
            /// The "environment" variable defined in the OpenAPI document.
            ///
            /// The default value is "prod".
            internal enum Environment: Swift.String {
                case prod
                case staging
                case dev
                /// The default variable.
                internal static var `default`: Environment {
                    return Environment.prod
                }
            }
        }
    }
    ///
    /// - Parameters:
    ///   - environment:
    internal static func server1(environment: Variables.Server1.Environment = Variables.Server1.Environment.default) throws -> Foundation.URL {
        try Foundation.URL(
            validatingOpenAPIServerURL: "https://{environment}.example.com/api/",
            variables: [
                .init(
                    name: "environment",
                    value: environment.rawValue
                )
            ]
        )
    }
}
```

Now when it comes to usage
```swift
let url = try Servers.server1() // ✅ works

let url = try Servers.server1(environment: .default)  // ✅ works

let url = try Servers.server1(environment: .staging)  // ✅ works

let url = try Servers.server1(environment: .stg)  // ❌ compiler error, stg not defined on the enum

// some time later staging gets removed from OpenAPI document
let url = try Servers.server1(environment: . staging)  // ❌ compiler error, staging not defined on the enum
```

If the document does not define enum values for the variable, an enum is still generated with a single member (the default required by the spec).

```yaml
servers:
  - url: https://example.com/api/{version}
    variables:
      version:
        default: v1
```

Before this PR:
```swift
/// Server URLs defined in the OpenAPI document.
internal enum Servers {
    ///
    /// - Parameters:
    ///   - version:
    internal static func server1(version: Swift.String = "v1") throws -> Foundation.URL {
        try Foundation.URL(
            validatingOpenAPIServerURL: "https://example.com/api/{version}",
            variables: [
                .init(
                    name: "version",
                    value: version
                )
            ]
        )
    }
}
```

With this PR:
```swift
/// Server URLs defined in the OpenAPI document.
internal enum Servers {
    /// Server URL variables defined in the OpenAPI document.
    internal enum Variables {
        /// The variables for Server1 defined in the OpenAPI document.
        internal enum Server1 {
            /// The "version" variable defined in the OpenAPI document.
            ///
            /// The default value is "v1".
            internal enum Version: Swift.String {
                case v1
                /// The default variable.
                internal static var `default`: Version {
                    return Version.v1
                }
            }
        }
    }
    ///
    /// - Parameters:
    ///   - version:
    internal static func server1(version: Variables.Server1.Version = Variables.Server1.Version.default) throws -> Foundation.URL {
        try Foundation.URL(
            validatingOpenAPIServerURL: "https://example.com/api/{version}",
            variables: [
                .init(
                    name: "version",
                    value: version.rawValue
                )
            ]
        )
    }
}
```

</details>

### Result

Refer to https://github.com/apple/swift-openapi-generator/pull/618#issuecomment-2346037222

### Test Plan

I have updated the petstore unit tests to reflect the changes made in this PR, see diff.


simonjbeaumont:
Thanks @theoriginalbit!. This is just a courtesy note to acknowledge I’ve seen it but I’ll probably get to it early next week.

czechboy0:
This is great, @theoriginalbit!

I only skimmed it, but two things I suspect we need to change before landing it:
1. This is an API breaking change, so it needs to be hidden behind a feature flag (see FeatureFlags.swift), and will need to be opt-in. As existing generated code that adopters are already calling today cannot change its API in a breaking way.
2. The enum should only be generated for variables that actually have "allowedValues" defined, but other raw strings should continue to be just strings.

Thanks again! I'm sure we'll be able to land this with a few tweaks.

theoriginalbit:
Thanks so much for taking a look and providing feedback @czechboy0.

When I was writing the PR description I did wonder about how this breaking change would be handled, I totally forgot to add the question to the description.

I will definitely take a look at the `FeatureFlags` and add one in for this, is there a particular convention for naming the feature flags that would be appropriate to enable this generation? And is there any unit testing infrastructure already set up for testing feature flags?

> 2. The enum should only be generated for variables that actually have "allowedValues" defined, but other raw strings should continue to be just strings.

So just to make sure I'm understanding correctly, in the scenario of the OpenAPI document

```yaml
servers:
  - url: https://{environment}.example.com/api/{version}
    variables:
      environment:
        default: prod
        enum:
          - prod
          - staging
          - dev
      version:
        default: v1
```

You'd like to see it output as the following generated code?

```swift
/// Server URLs defined in the OpenAPI document.
internal enum Servers {
    /// Server URL variables defined in the OpenAPI document.
    internal enum Variables {
        /// The variables for Server1 defined in the OpenAPI document.
        internal enum Server1 {
            /// The "environment" variable defined in the OpenAPI document.
            ///
            /// The default value is "prod".
            internal enum Environment: Swift.String {
                case prod
                case staging
                case dev
                /// The default variable.
                internal static var `default`: Environment {
                    return Environment.prod
                }
            }
        }
    }
    ///
    /// - Parameters:
    ///   - version:
    internal static func server1(
        environment: Variables.Server1.Environment = Variables.Server1.Environment.default,
        version: Swift.String = "v1"
    ) throws -> Foundation.URL {
        try Foundation.URL(
            validatingOpenAPIServerURL: "https://example.com/api/{version}",
            variables: [
                .init(
                    name: "environment",
                    value: environment.rawValue
                ),
                .init(
                    name: "version",
                    value: version
                )
            ]
        )
    }
}
```

I'm very happy to make this change, though one thing I want to confirm with you all first since perhaps my understanding of the OpenAPI spec was wrong.

My understanding of the spec is that a default-only variable, `version` in this case, is defined because the provider knows at some point there will be additional values but there are none to yet define in an enum property. So the following two configurations would be functionally equivalent

```yaml
# Syntax sugar
servers:
  - url: https://example.com/api/{version}
    variables:
      version:
        default: v1

# Verbose
servers:
  - url: https://example.com/api/{version}
    variables:
      version:
        default: v1
        enum:
          - v1
```

If that is the case, then I believe generating a default-only variable as an enum with a single case would be better for the consumer as if the provider later adds extra enum values then updating to that spec doesn't introduce a breaking change (it still uses the default value). Also, perhaps it better communicates that there is only one available option. I could see with the above document a consumer invoking as such simply because it's a string
```swift
let url = try Servers.server1(version: "v2")
```
Which wouldn't even be a runtime issue since default-only variables in the current implementation don't get `allowedValues` passed as a parameter. My thinking with this is if it were to be an enum with a single case the compiler would enforce that the consumer is only able to provide `v1` and not any other undocumented value.

What are your thoughts on this, have I misunderstood the purpose of default-only variables? Should they be open and unrestricted strings?

simonjbeaumont:
Quick (half-baked) thought on how we could do this in a non-API-breaking way so as to be able to land this without a feature flag or major version: could we make these enums `ExpressibleByStringLiteral`? IIUC this would allow the newly generated code to continue to work, where folks are passing in string literals for the parameters, but bring in the compile-time guarantees for those that want them.

Whether this is something we can do will depend on how we interpret the OpenAPI spec w.r.t. whether values outside of those defined in the document are permitted.

simonjbeaumont:
With regard to the OAS:

> Field Name | Type | Description
> -- | -- | --
> enum | [string] | An enumeration of string values to be used if the substitution options are from a limited set. The array MUST NOT be empty.
> default | string | REQUIRED. The default value to use for substitution, which SHALL be sent if an alternate value is not supplied. Note this behavior is different than the Schema Object’s treatment of default values, because in those cases parameter values are optional. If the enum is defined, the value MUST exist in the enum’s values.
> description | string | An optional description for the server variable. [CommonMark] syntax MAY be used for rich text representation.
>
> — source: https://spec.openapis.org/oas/latest.html#server-variable-object

> An enumeration of string values to be used if the substitution options are from a limited set. The array MUST NOT be empty.

I take this to mean that, if an enum is provided, then it _is_ a limited set.

theoriginalbit:
Ah yeah, I totally misread and misunderstood the OAS, that wording definitely suggests that if there is no enum values defined then it's an open field. I'll update the PR later today to revert those specific definitions to be open Strings, and only apply the enum generation to variables defined with an enum field.

Using `ExpressibleByStringLiteral` is an interesting thought 🤔 Though the only way I think I see that working would be that each enum needs to have an `undocumented(String)` case generated; at runtime that case would throw the same existing `RuntimeError` that was thrown when validating the allowed values. Since the OAS does specify "limited set" I imagine that's why the current latest release performs the runtime checks and throws an error if the value is outside of what was documented. It feels to me that this better suits a breaking (but feature flagged) change so that going forward there is no option to provide a string which then requires runtime checks. Thoughts?

theoriginalbit:
Apologies, I ran out of time to address the feedback the other day.

The changes I've just pushed fixes both of the issues raised. I've introduced a feature flag, `ServerVariablesEnums`, and it will only generate enums for variables that have `enum` defined in the OpenAPI Document.

So given the following definition

```yaml
servers:
  - url: https://example.com/api
    description: Example service deployment.
    variables:
      environment:
        description: Server environment.
        default: prod
        enum:
          - prod
          - staging
          - dev
      version:
        default: v1
```

With the `ServerVariablesEnums` feature flag not defined the output would be
```swift
/// Server URLs defined in the OpenAPI document.
internal enum Servers {
    /// Example service deployment.
    ///
    /// - Parameters:
    ///   - environment: Server environment.
    ///   - version: The overall API version
    internal static func server1(
        environment: Swift.String = "prod",
        version: Swift.String = "v1"
    ) throws -> Foundation.URL {
        try Foundation.URL(
            validatingOpenAPIServerURL: "https://example.com/api",
            variables: [
                .init(
                    name: "environment",
                    value: environment,
                    allowedValues: [
                        "prod",
                        "staging",
                        "dev"
                    ]
                ),
                .init(
                    name: "version",
                    value: version
                )
            ]
        )
    }
}
```

With the `ServerVariablesEnums` feature flag provided the output would be
```swift
/// Server URLs defined in the OpenAPI document.
internal enum Servers {
    /// Server URL variables defined in the OpenAPI document.
    internal enum Variables {
        /// The variables for Server1 defined in the OpenAPI document.
        internal enum Server1 {
            /// Server environment.
            ///
            /// The "environment" variable defined in the OpenAPI document. The default value is "prod".
            internal enum Environment: Swift.String {
                case prod
                case staging
                case dev
                /// The default variable.
                internal static var `default`: Environment {
                    return Environment.prod
                }
            }
        }
    }
    /// Example service deployment.
    ///
    /// - Parameters:
    ///   - environment: Server environment.
    ///   - version:
    internal static func server1(
        environment: Variables.Server1.Environment = Variables.Server1.Environment.default,
        version: Swift.String = "v1"
    ) throws -> Foundation.URL {
        try Foundation.URL(
            validatingOpenAPIServerURL: "https://example.com/api",
            variables: [
                .init(
                    name: "environment",
                    value: environment.rawValue
                ),
                .init(
                    name: "version",
                    value: version
                )
            ]
        )
    }
}
```

I've also added a new test case to validate the output when the feature flag is enabled.

Sources/_OpenAPIGeneratorCore/FeatureFlags.swift
```diff
@@ -28,6 +28,10 @@
 public enum FeatureFlag: String, Hashable, Codable, CaseIterable, Sendable {
     // needs to be here for the enum to compile
     case empty
+
+    /// When enabled any server URL templating variables that have a defined
+    /// enum field will be generated as an enum instead of raw Strings.
+    case serverVariablesAsEnums = "ServerVariablesEnums"
```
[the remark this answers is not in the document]
theoriginalbit: Awesome thanks!

Tests/OpenAPIGeneratorReferenceTests/FileBasedReferenceTests.swift
```diff
@@ -43,6 +43,11 @@ class FileBasedReferenceTests: XCTestCase {
     }
 
     func testPetstore() throws { try _test(referenceProject: .init(name: .petstore)) }
+    
+    func testPetstoreWithServerVariablesEnumsFeatureFlag() throws {
```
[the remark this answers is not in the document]
theoriginalbit: Ah good to know, can do!

It didn't feel great duplicating the entire file, I figured maintenance was going to be a nightmare, but the file tests were all that had failed with my changes. Will move over to the snippet-based reference tests.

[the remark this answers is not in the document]
theoriginalbit: I've run out of time tonight to finalise this change and push it, I'll finish it off in the morning before work.

I have created a proposal though as requested in this review. Let me know if there is anything else I need to do for it.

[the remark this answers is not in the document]
czechboy0: Amazing, thank you so much! Added a few comments there, but no rush - we appreciate your help here 🙂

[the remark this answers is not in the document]
theoriginalbit: I've finished adding some snippet-based tests.

I decided to limit the amount of permutations to test to cover the main scenarios, if there are any additional scenarios you'd like to see tested let me know and I can add them.

The only thing I couldn't figure out was how to test with multiple servers in the YAML, I'm not familiar enough with YAML to know the syntax required and the YAMLDecoder kept failing to decode the input strings.

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
```diff
@@ -5473,6 +5673,21 @@ extension SnippetBasedReferenceTests {
...
+        featureFlags: FeatureFlags = [],
+        file: StaticString = #filePath,
+        line: UInt = #line
+    ) throws {
+        continueAfterFailure = false
+        let server = try YAMLDecoder().decode(OpenAPI.Server.self, from: serverYAML)
+        let translator = try makeTypesTranslator(accessModifier: accessModifier, featureFlags: featureFlags)
+        let translation = translator.translateServers([server])
+        try XCTAssertSwiftEquivalent(translation, expectedSwift, file: file, line: line)
+    }
```
czechboy0, line 5690:
```suggestion
    func assertServersTranslation(
        _ serversYAML: String,
        _ expectedSwift: String,
        accessModifier: AccessModifier = .public,
        featureFlags: FeatureFlags = [],
        file: StaticString = #filePath,
        line: UInt = #line
    ) throws {
        continueAfterFailure = false
        let servers = try YAMLDecoder().decode([OpenAPI.Server].self, from: serversYAML)
        let translator = try makeTypesTranslator(accessModifier: accessModifier, featureFlags: featureFlags)
        let translation = translator.translateServers(servers)
        try XCTAssertSwiftEquivalent(translation, expectedSwift, file: file, line: line)
    }
```

If you make this change, you can then pass an array of servers for tests, e.g.

```
- url: ...
  description: ...
- url: ...
```

Tests/OpenAPIGeneratorReferenceTests/FileBasedReferenceTests.swift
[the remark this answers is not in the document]
czechboy0: Ah I see, yeah if you change the util function to take a YAML array of servers, you should be able to test multiple: https://github.com/apple/swift-openapi-generator/pull/618/files#r1766866948

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
theoriginalbit: Hmm. I tried that and the decoder threw a decoding error about a dictionary or something…maybe I had the indentation wrong on a line and didn't notice. will try it again, thanks :)

theoriginalbit: Ok yep I must've just had some weird indentation. That works now, I've introduced two tests for multiple servers, one with and one without the feature flag enabled.

czechboy0:
Okay the implementation looks good, now let's wait for the review to run its course and we'll finalize that once accepted - thanks again! 🙏

theoriginalbit:
> Okay the implementation looks good, now let's wait for the review to run its course and we'll finalize that once accepted - thanks again! 🙏 

Awesome, no worries at all. Hopefully some people can provide feedback on the namespace names, otherwise looking forward to this going through the finalising steps.

Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateServers.swift
```diff
@@ -64,15 +36,70 @@ extension TypesFileTranslator {
...
+                                        label: "variables",
+                                        expression: .literal(.array(variables.map(\.initializer)))
+                                    )
                                 ])
                         )
                     )
                 ]
+            ).deprecate(
+                if: isDeprecated,
+                description: .init(message: "Migrate to the new type-safe API for server URLs.")
```
czechboy0, line 51: Could we use the `renamed:` parameter of the deprecation, to allow people to migrate by accepting a Fix-It?

Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Types.swift
```diff
@@ -154,25 +154,87 @@ extension APIProtocol {
 /// Server URLs defined in the OpenAPI document.
 public enum Servers {
     /// Example Petstore implementation service
+    public enum Server1 {
+        public static func url() throws -> Foundation.URL {
```
czechboy0, line 158: We should also generate a comment on the `url()` static function. Might make sense to simply duplicate the user-provided comment that we already put on the namespace (e.g. above `Server1` here)

Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Types.swift
```diff
@@ -154,25 +154,87 @@ extension APIProtocol {
...
+        ///
+        /// - Parameters:
+        ///   - _protocol:
+        ///   - subdomain: A subdomain name.
+        ///   - port:
+        ///   - basePath: The base API path.
+        public static func url(
+            _protocol: Swift.String = "https",
+            subdomain: Swift.String = "test",
+            port: Port = Port._443,
```
czechboy0, line 204:
```suggestion
            port: Port = ._443,
```

Can we omit repeating the type, to match what users would type manually?

Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateServers.swift
theoriginalbit: I hadn't included it because of the point you raised in the forums discussion ([link](https://forums.swift.org/t/proposal-soar-0012-generate-enums-for-server-variables/74737/4))

> I don't think we need the `renamed:` parameter - as the Fix-It wouldn't be able to mechanically fix the issue anyway for the adopter. If the adopter sees the deprecation warning, it's because they're _explicitly_ passing in a `String` parameter in a place that now has an enum variant. They'll need to e.g. change `"staging"` to `.staging` in their code to switch to the new variant. I'm happy for the warning to guide them to it by being descriptive.

Though I was very close to adding it when I was making the implementation, because it would get the adopter most (or all if they're using defaults) of the way to the new interface.

Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Types.swift
theoriginalbit: Yeah another case of "it's how I originally did it", but I wasn't entirely sure, since it would be a duplication of the comment on the namespace. I'd also toyed with having it on the function and not the namespace.

Happy to have it on both, then adopters will be able to see it through auto-complete, quick help, and the quick help inspector no matter the symbol they have selected. Will make this change with the other changes once they've had some extra discussion.

Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Types.swift
theoriginalbit: I had 2 reasons for having it there
1. **Consistency.** Other types have been explicit (e.g. `Foundation.URL` and `Swift.String`) which I figured was more to disambiguate if any types conflicted, but doing it here kept it consistent with being explicit.
1. **Speed/Efficiency.** The compiler wouldn't need to use type inference and consume as many resources on the type checking; which given the other issues raised around compile speed I figured that being explicit would help on that front, especially on larger documents with a larger number of generated symbols.


Would you agree with my justification here?

Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateServers.swift
czechboy0: I believe I made my comment when the proposed new API was identical to the existing API, just with different parameter types. In that case, the warning would really only be seen by people who were explicitly passing the "wrong" type (String instead of an enum value), and a Fix-It wouldn't be able to help them with that.

But with the latest shape of the proposed API, where we've completely deprecated the existing static method on the `Servers` namespace, and instead moved it to `Servers.Server1.url()`, the warning will be seen by everyone who updates to the latest generator, and will need to switch to using the new API.

For that, we can provide a `renamed:` parameter to the deprecation, as it'll be fixable using a Fix-It. Most people will use that natural upgrade path.

And only those who are passing a String to what's newly an enum will, after accepting the Fix-It, see a build error (instead of a warning) as the new API doesn't have a String-based variant anymore, and it should be clear how they need to fix it (it'll be a normal Swift parameter type mismatch error).

Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Types.swift
czechboy0: Yeah I believe we already duplicate some user-provided comments on multiple generated APIs, so this wouldn't be the first time.

Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Types.swift
czechboy0: It's fine either way.

Re 1, we really do generate fully qualified type names in most places to avoid possibly clashing with generated types. But default values don't need to be fully qualified, see below.

Re 2, it's fine performance wise, as the type of the parameter is explicitly specified in the declaration, so the compiler doesn't have to infer anything, it just needs to verify the default value can satisfy the type.

If you see that we're generating a fully qualified name in a default parameter value anywhere, let me know, we might not need to do that.

Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateServers.swift
theoriginalbit: Yeah fair points, I'll add a renamed parameter so adopters get a Fix-It for the new structures 👍

Is the current message provided acceptable?

Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Types.swift
theoriginalbit: Interesting, I had seen compile performance issues previously with type inference on parameters; though it was quite a while ago now, so maybe it's been long fixed and I'm just stuck in the past.

I'll remove the explicit type from the parameter with my next update.

Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateServers.swift
czechboy0: Maybe we don't even need an explicit message if we have the `renamed` param? What do you think?

theoriginalbit: Yeah it's probably better without the message. Thinking about it now it would mean the default message it shows would be the fact it's renamed. Which provides the adopter the immediate feedback of what the Fix-It is doing. I'll remove the message and push an update for all the comments

Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Types.swift
```diff
@@ -154,25 +154,89 @@ extension APIProtocol {
...
+        /// Example Petstore implementation service
+        public static func url() throws -> Foundation.URL {
+            try Foundation.URL(
+                validatingOpenAPIServerURL: "https://example.com/api",
+                variables: []
+            )
+        }
+    }
+    /// Example Petstore implementation service
+    @available(*, deprecated, renamed: "Servers.Server1.url")
```
czechboy0, line 167:
```suggestion
    @available(*, deprecated, renamed: "Servers.Server1.url()")
```

Should it also contain the `()` or not? Can you check that accepting the Fix-it does the right thing?

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
```diff
@@ -5181,6 +5181,297 @@ final class SnippetBasedReferenceTests: XCTestCase {
...
+                variables:
+                    version:
+                        default: v1
+              - url: 'https://api.example.com/'
+                description: Vanity URL for production.api.example.com
+            """,
+            """
+            public enum Servers {
+                public enum Server1 {
+                    @frozen public enum Environment: Swift.String {
```
czechboy0, line 5352: Can you also generate the enums as `Sendable`?

czechboy0 (changes requested):
A few minor change requests, but looks good overall! Thanks

czechboy0:
Oh, one more thing:

- [ ] please find and replace all references to `Servers.server1()` (and similar) in this repo with the new syntax, especially in documentation and examples, to make sure that out of the box, folks use the new syntax and don't see the warning

Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateServersVariables.swift
```diff
@@ -0,0 +1,214 @@
...
+                context: context
+            )
+        }
+    }
+
+    // MARK: Generators
+    
+    /// Represents a server variable and the function of generation that should be applied.
+    protocol ServerVariableGenerator {
+        /// Returns the declaration (enum) that should be added to the `Variables.Server#`
```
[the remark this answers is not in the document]
theoriginalbit: Nope, not sure how that happened must have accidentally typed it and not noticed

Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Types.swift
theoriginalbit: I tested this as I was writing it, and it seems to not matter. Whether the renamed function takes no parameters, many parameters, parameters with defaults the Fix-It just seems to work; weirdly I was even able to get Fix-Its like `"Servers.Server3.url(a:)"` to work as well.

theoriginalbit:
> * please find and replace all references to `Servers.server1()` (and similar) in this repo with the new syntax, especially in documentation and examples, to make sure that out of the box, folks use the new syntax and don't see the warning

None of the examples appeared to use the generated servers URL functions.

I also opted to keep one test assertion, for the sake of regression testing, which tested the deprecated API threw if an invalid variable was supplied. Once the deprecated functions are no longer being generated I figured the assertion could be removed, but at least for now it serves a purpose.

theoriginalbit:
I've addressed the latest round of feedback @czechboy0 :)

> * [ ] please find and replace all references to `Servers.server1()` (and similar) in this repo with the new syntax, especially in documentation and examples, to make sure that out of the box, folks use the new syntax and don't see the warning

None of the examples appeared to use the generated servers URL functions.

I also opted to keep one test assertion, for the sake of regression testing, which tested the deprecated API threw if an invalid variable was supplied. Once the deprecated functions are no longer being generated I figured the assertion could be removed, but at least for now it serves a purpose.

Tests/OpenAPIGeneratorReferenceTests/Resources/ReferenceSources/Petstore/Types.swift
czechboy0: Fair enough, thanks for checking. Let's leave it as-is.

czechboy0:
> > * [ ]  please find and replace all references to `Servers.server1()` (and similar) in this repo with the new syntax, especially in documentation and examples, to make sure that out of the box, folks use the new syntax and don't see the warning
> 
> None of the examples appeared to use the generated servers URL functions.

There are places in `Examples` and `IntegrationTest` that'll need updating, but we'll need to do that as a follow-up, once this change lands and gets released. All the other places you've already updated, thank you 🙂

Tests/PetstoreConsumerTests/Test_Types.swift
```diff
@@ -222,14 +222,17 @@ final class Test_Types: XCTestCase {
...
+        XCTAssertEqual(try Servers.Server3.url(), URL(string: "https://test.example.com:443/v1"))
         XCTAssertEqual(
-            try Servers.server3(subdomain: "bar", port: "8443", basePath: "v2/staging"),
+            try Servers.Server3.url(subdomain: "bar", port: ._8443, basePath: "v2/staging"),
             URL(string: "https://bar.example.com:8443/v2/staging")
         )
+        // Intentionally using the deprecated static function to check for regressions.
+        // Once the deprecated functions are no longer being generated this assertion is
+        // unnecessary and can be removed.
         XCTAssertThrowsError(try Servers.server3(port: "foo")) { error in
```
czechboy0, line 236: Please just remove the `XCTAssertThrowsError`, we don't want to have that warning in our tests as the goal is to be warning-free and run with warnings-as-errors in CI. The underlying validation is run as part of Swift OpenAPI Runtime tests, so I'm happy for us to drop it here.

czechboy0 (changes requested):
One small request, otherwise looks great! Kicking off CI.

czechboy0 (approved):

czechboy0:
lgtm, we just need to get the CI green. The format check just requires you to run e.g. `act --bind workflow_call -j soundness --input format_check_enabled=true` and commit the chages.

czechboy0:
Oh and you'll need to move the nested protocol out, as we still support 5.9:

```
/__w/swift-openapi-generator/swift-openapi-generator/Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateServersVariables.swift:51:14: error: protocol 'ServerVariableGenerator' cannot be nested inside another declaration
```

theoriginalbit:
Sorry took a bit, had to download `act` and Docker; then the inevitable debugging to get them to actually work.

Hopefully all fixed now, @czechboy0.

czechboy0:
Thank you for the great contribution, @theoriginalbit! 👏

Issue #573, opened 2024-05-17
Add ability to attach protocols to generated models


armintelker:
### Motivation

We need all of our models generated by `swift-openapi-generator` to comply with protocols like `Equatable`or `Identifiable` to work with our overlaying architecture. Manually extending each generated model to conform to these protocols is insufficient and cumbersome, especially as the number of models increases.

### Proposed solution

Add a feature to the `swift-openapi-generator` that allows specifying which protocols the generated models should conform to.  This could be implemented as a configuration option in the generator, enabling automated protocol conformance during the model generation process.

### Alternatives considered

The current workaround is manually extending each generated model, like so:
```swift
extension ModelA: Equatable {}
extension ModelB: Equatable {}
extension ModelC: Equatable {}
extension ModelD: Equatable, Identifyable {
    var id: Int 
    //...
}
```
However, this approach is not scalable and becomes unwieldy as the number of models grows.

### Additional information

If there is already an existing feature to achieve this, i apologize for the oversight and kindly request guidance to the relevant documentation. I could not find any reference to such a feature in the current documentation.

References:

[Identifiable Documentation](https://developer.apple.com/documentation/swift/identifiable)
[Equatable Documentation](https://developer.apple.com/documentation/swift/equatable)
Thank you for considering this request.

armintelker:
Hey, thanks for the quick response. 
I have overseen that the DTO is already conforming to Hashable. This makes, of course, everything a lot easier. So my request gets kind of needless.
Indeed the `Identifiable` protocol suggestion would be more like a "nice to have". 
I could imagine that the generator could add the `Identifiable` to the model as soon it sees a property `id` exists that also conforms to the [constraint](https://developer.apple.com/documentation/swift/identifiable/id-swift.associatedtype) of the protocol.

simonjbeaumont:
This is similar to #383, where could conceivably attach arbitrary conformances or macros that are stated in the config file and it be up to the adopter to ensure these are available in their project. 

It wouldn't be too different from the ability we have know to express `additionalImports`.

czechboy0:
That's true, @simonjbeaumont, and the macro itself would contain the logic for how to e.g. implement `Identifiable`, instead of the generator having to know how to do it. I like that.

shadone:
+1 to this. In my use case it would also be useful to have ability to add Identifiable conformance to some types. There is literally no other way to do this - I have a library that uses swift-openapi-generator to generate the type and expose them publicly.

[...]

Issue #383, opened 2023-11-17
Can I add some parameters to specify that the generated Schema and API support swift macros?


HonQii:
It would be nice to be able to add macros to the generated Schema and API requests in order to extend the functionality, e.g. all of my request return structures have an optional error field, so if I could add a macro that would allow me to extend the support for a protocol for them, I could easily handle the error uniformly

simonjbeaumont:
@HonQii could you elaborate a little on what you'd like to see? You can currently write extensions on the types generated and conform them to protocols that you define. Are you asking for the protocol conformance to be automatically added for you by the generator, potentially by specifying a macro you'd like to run in the generator config?

Maybe you could provide a quick snippet of how you'd like the generated code to be different.

czechboy0:
Maybe @HonQii means that they could configure some macro name to be automatically attached to all generated types, similar to how we allow adding additional imports? But the user would still need to provide the macro implementation.

(Which is an interesting idea that would indeed provide quite a lot of flexibility for sufficiently motivated adopters.)

HonQii:
> Maybe @HonQii means that they could configure some macro name to be automatically attached to all generated types, similar to how we allow adding additional imports? But the user would still need to provide the macro implementation.🔄  ❓
> 
> (Which is an interesting idea that would indeed provide quite a lot of flexibility for sufficiently motivated adopters.)🔄  ❓

Right, what I mean is that I can configure macros to the generated Schema as well as the Client so that I can do some wrapping. An example:

```swift
// Each of my request response structures will have an error field with a message and code if there is an error
struct BaseResponse {
      var error: Error? // empty if there is no error.
}
```

Normally I would need to define a protocol for all structures to follow, and then I could handle these errors consistently. But in the case of autogeneration, I think it's more appropriate to use a macro for this purpose

Similarly, I need a layer of encapsulation for each request in the client to handle exceptions, which can also be accomplished using a macro

simonjbeaumont:
> ```swift
> // Each of my request response structures will have an error field with a message and code if there is an error
> struct BaseResponse {
>       var error: Error? // empty if there is no error.
> }
> ```

OOI, are you the author of this document? Because, if so, for this example, you could have the generated code make use of shared types by writing this in the OpenAPI document itself. For example, you can make use of of a common type in `#/components/schemas`, or you could define a shared response in `#/components/responses`, or you could try and combine these with an `allOf`[^1].

[...]

PR #635, opened 2024-09-30
Additional api protocol conformances

peterringset:
### Motivation

The generated code from this tool offers a very good starting point for integrating an API into a larger project. One challenge that I've run into when using this tool is testability and mocking. While it is possible to write mocks by hand I have found it very useful to use code generation tools (e.g. Sourcery) to do most of the heavy lifting. To enable behavior like this it is often necessary to annotate the client/server source in some way. This change offers a way to let the `APIProtocol` declaration have a conformance to one or more user-defined protocols to achieve this.

### Modifications

- Added an option to the generate command to list protocols that the user wishes the API to conform to
- Added entry to the config-file to list protocols that the user wishes the API to conform to
- Added protocol conformance of user-defined protocols to the `APIProtocol`
- Added test to verify correct behavior

### Result

The generate command and config file will get an extra option (`additionalAPIProtocols`) for the user to list additional protocols they wish that the `APIProtocol` type conforms to.

### Test Plan

- Added a reference test that generates an empty API with conformance to an additional example protocol
- Modified existing config file test to include this change

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Config.swift
+++ b/Sources/_OpenAPIGeneratorCore/Config.swift
@@ -32,6 +32,9 @@ public struct Config: Sendable {
     /// Additional imports to add to each generated file.
     public var additionalImports: [String]
 
+    /// Additional protocol conformances to add to the APIProtocol type
+    public var additionalAPIProtocols: [String]
+
     /// Filter to apply to the OpenAPI document before generation.
     public var filter: DocumentFilter?
 
@@ -49,12 +52,14 @@ public struct Config: Sendable {
         mode: GeneratorMode,
         access: AccessModifier,
         additionalImports: [String] = [],
+        additionalAPIProtocols: [String] = [],
         filter: DocumentFilter? = nil,
         featureFlags: FeatureFlags = []
     ) {
         self.mode = mode
         self.access = access
         self.additionalImports = additionalImports
+        self.additionalAPIProtocols = additionalAPIProtocols
         self.filter = filter
         self.featureFlags = featureFlags
     }
--- a/Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateAPIProtocol.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateAPIProtocol.swift
@@ -28,7 +28,7 @@ extension TypesFileTranslator {
         let protocolDescription = ProtocolDescription(
             accessModifier: config.access,
             name: Constants.APIProtocol.typeName,
-            conformances: Constants.APIProtocol.conformances,
+            conformances: Constants.APIProtocol.conformances + config.additionalAPIProtocols,
             members: functionDecls
         )
         let protocolComment: Comment = .doc("A type that performs HTTP operations defined by the OpenAPI document.")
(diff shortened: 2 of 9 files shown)


simonjbeaumont:
@peterringset thanks for putting up this patch and your interest in the project!

We've heard of folks wanting this before (e.g. https://github.com/apple/swift-openapi-generator/issues/573, https://github.com/apple/swift-openapi-generator/issues/383).

While I'm not against it, I'd like to understand the motivation more. Unlike the issue that asked to attach user-defined macros, IIUC this PR would allow users only to _declare_ a protocol conformance on the generated `protocol APIProtocol`. Some questions:

What's the developer experience when the protocol doesn't exist? Presumably the generated code just fails to compile but we need to make sure this surfaces well, because that conformance is now in a derived source file.

What's next for the adopter? Are we relying on them having to go and provide the protocol witnesses to actually _conform_ to the protocol? If this is the case then does this bring much value? It seems like it will just separate the conformance from the conformance declaration in user code, which might actually be more confusing?

Potentially this is mainly focused on protocols that rely only on default implementations so _declaring_ conformance is enough?

peterringset:
Hi! Thanks for the quick feedback. I'll write my replies one by one below:

> What's the developer experience when the protocol doesn't exist? Presumably the generated code just fails to compile but we need to make sure this surfaces well, because that conformance is now in a derived source file.

Yes, that's true. My initial thought would be that this is something that the developer opts in to, and it's hopefully not something that they would unintentionally add to their code generation. With that in mind I hope that all developers adopting this will understand how to use this option and declare the necessary protocol(s) in a different file that's available in the generated code's scope. I don't know if there's anything more that could/should be done? Maybe adding some comments to document what's happening?

> What's next for the adopter? Are we relying on them having to go and provide the protocol witnesses to actually _conform_ to the protocol? If this is the case then does this bring much value? It seems like it will just separate the conformance from the conformance declaration in user code, which might actually be more confusing?

This usage would in most cases mean that the protocol(s) to be added already exists somehow in the package that we're generating code in. In my particular case it's necessary for `APIProtocol` to inherit from the custom protocols (`AutoMockable` in my case with Sourcery) for the mocks generation to work properly. I.e. `protocol APIProtocol: Sendable, AutoMockable { ...` works, but `extension APIProtocol: AutoMockable { ...` does not work, since protocols cannot declare inheritance to other protocols like that. This allows me to generate a mock for `APIProtocol`, and not for `Client`.

Another use case could be that the custom protocol requirements exactly match the `APIProtocol`s signatures, such that `APIProtocol` is already implementing the protocol by just declaring conformance. Although, this could also be solved other ways.

> Potentially this is mainly focused on protocols that rely only on default implementations so _declaring_ conformance is enough?

Yep, I think that is the most common use-case for a feature like this. The Sourcery use-case for instance has an empty protocol `AutoMockable` that works kind of like an annotation, and allows Sourcery to perform the mock generation for protocols that inherit from `AutoMockable`.

czechboy0:
Hi @peterringset,

yeah this is an interesting situation. I'm also not necessarily against landing this PR, but I'd like to explore how much this really helps compared to:

```
public typealias AutoMockableAPIProtocol = APIProtocol & AutoMockable
```

and then using `AutoMockableAPIProtocol` in your codebase, which has the advantage of not mixing the generated code from the two separate tools? Or would Sourcery not generate the code if the conformance is declared in the typealias?

peterringset:
@czechboy0 No, unfortunately Sourcery doesn't pick up the `AutoMockableAPIProtocol` typealias as an eligible type for mocks. AFAIK it only cares about protocols directly inheriting `AutoMockable`, so composition will also not work in this particular case.

czechboy0:
I see. From briefly skimming its sources, I wonder if support for inspecting typealiases could be added, something around https://github.com/krzysztofzablocki/Sourcery/blob/8acc3bf2ba1828e5f1126c1acc23a9e79561652d/Templates/Templates/AutoMockable.stencil#L335C10-L335C55 ?

czechboy0 (changes requested):
Marking as blocked for now, we suggested a possible workaround above. If that doesn't work, we can discuss continuing to move this PR forward, however.

PR #638, opened 2024-10-01
[Refactor] Introduce TranslatorContext (1/2)

czechboy0:
### Motivation

To make upcoming PRs less noisy, we need a way to propagate configuration all the way deep to types like TypeAssigner/TypeMatcher, which historically didn't have that need.

### Modifications

Introduced a new wrapper type called TranslatorContext, which wraps the existing closure that converts arbitrary strings into safe Swift names (which might in the future also need to be further customized).

The idea is that this new struct is what we'd attach more config to shortly.

There'll be a few more smaller PRs like this, I'm breaking them up for easier review.

### Result

Generalized the way to propagate config into low level translator utilities.

### Test Plan

All tests still pass.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/ClientTranslator/ClientTranslator.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/ClientTranslator/ClientTranslator.swift
@@ -37,8 +37,7 @@ struct ClientFileTranslator: FileTranslator {
         let imports =
             Constants.File.clientServerImports + config.additionalImports.map { ImportDescription(moduleName: $0) }
 
-        let clientMethodDecls =
-            try OperationDescription.all(from: doc.paths, in: components, asSwiftSafeName: swiftSafeName)
+        let clientMethodDecls = try OperationDescription.all(from: doc.paths, in: components, context: context)
             .map(translateClientMethod(_:))
 
         let clientStructPropertyDecl: Declaration = .commentable(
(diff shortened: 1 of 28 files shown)


simonjbeaumont (approved):
Nice patch. I don't love threading context around but can see the value. Don't love the name of the closure as I thought we prefer to avoid _as_ methods in API design, but since it's not public API I'm not gonna bikeshed it 👍

czechboy0:
> Don't love the name of the closure

Yeah same, but I just moved it, it's not a new name. Happy for us to rename it globally separately.

theoriginalbit:
Would it make sense for a 'part 3' PR which adds the `Config` instance to this context? I think it would further help reduce the noise in PRs like #627 which needs to access a value on config at that same deep `TypeMatcher` type.

czechboy0:
No, that was my original plan, but Context contains a lot more information than we need to propagate, which would make testing unnecessary noisy.

The idea is that we copy/extract whatever we need from Config onto TranslationContext as specific properties.

For example, for the UUID PR, there should be a new Bool property added on TranslationContext called `generateTypesafeUUID` or similar.

theoriginalbit:
Ok that makes a lot of sense, thanks.

Though in the context of the UUID PR wouldn't the `featureFlags` property be more desirable to copy over to avoid the noise of construction when N other similar PRs pop up and want to have their own properties too?

czechboy0:
That's one way, but I think it's better to be very explicit about the exact inputs that are needed, rather than passing all of feature flags or the entire Config in. As that makes testing later, as it's not clear which parts need to be mocked in tests and which are ignored.

theoriginalbit:
True, if the config or feature flags end up quite large then it would be difficult to know what needs to be mocked without a full understanding of everywhere it's used.

So I guess it'll be up to the UUID PR to now set the precedent of adding a property. 👍

PR #640, opened 2024-10-01
[Refactor] Introduce TranslatorContext (2/2)

czechboy0:
### Motivation

Depends on https://github.com/apple/swift-openapi-generator/pull/638, so don't review until that lands.

Continuation of #638, broke up into multiple PRs for easier review.

Also a non-functional change, purely a refactoring.

### Modifications

Changed a bunch of TypeMatcher methods from static to being instance methods, which gives them access to the translator context.

Moved a few other methods that depended on the static methods to be extensions on the closest type that already has access to the context.

### Result

Now TypeAssigner/TypeMatcher can take runtime configuration, e.g. through feature flags.

### Test Plan

All tests still pass.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeMatcher.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeMatcher.swift
@@ -42,7 +42,7 @@ struct TypeMatcher {
     func tryMatchBuiltinType(for schema: JSONSchema.Schema) -> TypeUsage? {
         Self._tryMatchRecursive(
             for: schema,
-            test: { schema in Self._tryMatchBuiltinNonRecursive(for: schema) },
+            test: { schema in _tryMatchBuiltinNonRecursive(for: schema) },
             matchedArrayHandler: { elementType, nullableItems in
                 nullableItems ? elementType.asOptional.asArray : elementType.asArray
             },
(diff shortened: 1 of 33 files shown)


Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeMatcher.swift
```diff
@@ -69,9 +68,9 @@ struct TypeMatcher {
         try Self._tryMatchRecursive(
             for: schema.value,
             test: { (schema) -> TypeUsage? in
-                if let builtinType = Self._tryMatchBuiltinNonRecursive(for: schema) { return builtinType }
+                if let builtinType = _tryMatchBuiltinNonRecursive(for: schema) { return builtinType }
                 guard case let .reference(ref, _) = schema else { return nil }
-                return try TypeAssigner(asSwiftSafeName: asSwiftSafeName).typeName(for: ref).asUsage
+                return try TypeAssigner(context: context).typeName(for: ref).asUsage
```
simonjbeaumont, line 73: Do we need to mint a new `TypeAssigner` on each of these calls? Is this the only place `context` is used? Maybe we're better having the assigner be a stored property instead? WDYT?

czechboy0: Possibly, but I think both types need a fresh look anyway, I'm hoping we can do that when we revisit the handling of optionals (as this code will be affected). This PR really just replaced the single closure with a wrapper around that closure, and tried to change nothing else.

simonjbeaumont:
I'll wait until the other PR lands, but will approve it when it's all rebased.

czechboy0:
> I'll wait until the other PR lands, but will approve it when it's all rebased.

Okay, ready to go.

simonjbeaumont (approved):

PR #649, opened 2024-10-10
Support default values for schema properties

### LINKED ISSUE #66: There is no default values for components.
The generator skips the default values for Components.Schemas. It would be nice to have a default value for the property itself and in the initializer. I made an example:

Spec example:
Screenshot 2023-06-13 at 20 22 25

Current:
Screenshot 2023-06-13 at 20 20 58

Desired:
Screenshot 2023-06-13 at 20 21 52

PARAIPAN9:
### Motivation

- Fixes #66.

### Modifications

- Update schema translations.

### Result

- Schema translations will support default values.

### Test Plan

- Run test suite. TODO - Write more tests.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Layers/StructuredSwiftRepresentation.swift
+++ b/Sources/_OpenAPIGeneratorCore/Layers/StructuredSwiftRepresentation.swift
@@ -105,6 +105,11 @@ enum LiteralDescription: Equatable, Codable {
     /// For example `42`.
     case int(Int)
 
+    /// A double literal.
+    ///
+    /// For example `42.24`.
+    case double(Double)
+
     /// A Boolean literal.
     ///
     /// For example `true`.
--- a/Sources/_OpenAPIGeneratorCore/Renderer/TextBasedRenderer.swift
+++ b/Sources/_OpenAPIGeneratorCore/Renderer/TextBasedRenderer.swift
@@ -453,6 +453,7 @@ struct TextBasedRenderer: RendererProtocol {
                 write("\"\(string)\"")
             }
         case let .int(int): write("\(int)")
+        case let .double(double): write("\(double)")
         case let .bool(bool): write(bool ? "true" : "false")
         case .nil: write("nil")
         case .array(let items):
--- a/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateAllAnyOneOf.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateAllAnyOneOf.swift
@@ -73,10 +73,19 @@ extension TypesFileTranslator {
                 } else {
                     associatedDeclarations = []
                 }
+
+                var defaultValue: PropertyBlueprint.DefaultValue? = nil
+                if let schemaDefaultValue = schema.defaultValue {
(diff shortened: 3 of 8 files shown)


PARAIPAN9:
Opened it as a draft. Not sure if I addressed correctly all the schema translations. I'd appreciate your input so far.

czechboy0:
Hi @PARAIPAN9,

thanks for starting on this work! I think roughly the direction is good, but it's difficult for me to fully understand the impact of the changes on the generated code.

I wonder if we could flip it, and first just write a bunch of snippet tests (that'll fail, initially), that show what code we want generated? That's where we can discuss which defaults we want to support, and which ones we might want to hold off on for now. I'm seeing defaults for things like typealiases and arrays and I realize I'm not sure I know how _those_ should work.

Note that you might also want to add a sublist under `default` here: https://github.com/apple/swift-openapi-generator/blob/7893a6c542c9be486378011ca90e8487c8159cbf/Sources/swift-openapi-generator/Documentation.docc/Articles/Supported-OpenAPI-features.md?plain=1#L171

Write out all the schema types we support in general, and have an empty box for each of them. Then we can better track which schemas we have a plan/implementation for, and which ones don't support defaults yet.

How does that sound?

PARAIPAN9:
Sounds perfect @czechboy0!! Thank you very much for your guidance!!

PARAIPAN9:
I've created this table to show you how defaults would be generated according to the changes from this PR to give you a better undestanding. Feel free to comment.
|                | OpenAPI Document | Generated Code| Comments |
|----------------|----------------|----------------|----------------|
| **Object Schema** |objectSchemaOpenAPIDoc| objectSchemaGenerated|  ...  |
| **Array Schema** |arrayOpenAPIDoc|arrayGenerated|  ...  |
| **Primitive Schemas** | typealiasOpenAPIDoc|typealiasGenerated|  ...  |

simonjbeaumont:
The object schema defaults look sensible (default parameter for the init). 

For the other two, the static `default` property is on `Components.Schemas` which is probably wrong. How would multiple of these compose? Also, because we're generating type aliases for these, it's not even clear to me where the default _should_ go.

It wouldn't make sense, for example, to add an extension because we'd be adding an extension to e.g. `[Swift.Bool]`. Had we been generating nominal types for these, we might have been able to do it. 

What I was thinking is that we don't generate this static property but we propagate the default value from the document to the places where the type is used as a parameter. E.g. anything that was generating a parameter of this type would continue to do so, but also include a default value for that parameter. 

WDYT?

PARAIPAN9:
Yes, it makes sense. Thanks!!

czechboy0:
We already have something similar for optionality of values - they affect the call site of the type, rather than the definition of the type. So there's some precedent at least.

PARAIPAN9:
FYI: I've been busy these past couple of months, but I haven't forgotten about this PR.

jpsim:
Looking forward to this one!

simonjbeaumont:
@PARAIPAN9 just checking in on this one. Is it something you still anticipate to find time for? 🙏

PARAIPAN9:
@simonjbeaumont, yes I will find time for it and update it as soon as I can.

PARAIPAN9:
An update on this PR. I’ve tried to find time to work on it, but I haven’t been able to. I don’t want to keep it blocked, so if someone else wants to take over, please feel free to do so.

PR #652, opened 2024-10-15
[CI] Update to Swift 6 CI

FranzBusch:
# Motivation

We just updated our CI matrix in NIO to only support 5.9, 5.10 and 6.

# Modification

This PR updates the trigger files in this repo. Since this repo was always 5.9+ this is easy.

# Result

Up to date CI

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -10,22 +10,16 @@ jobs:
         uses: swiftlang/github-workflows/.github/workflows/soundness.yml@main
         with:
             api_breakage_check_enabled: false
-            broken_symlink_check_enabled: true
-            docs_check_enabled: true
-            format_check_enabled: true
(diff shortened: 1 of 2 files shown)


.github/workflows/pull_request.yml
```diff
@@ -9,23 +9,16 @@ jobs:
         name: Soundness
         uses: swiftlang/github-workflows/.github/workflows/soundness.yml@main
         with:
-            api_breakage_check_enabled: false
```
czechboy0, line 12: I think we want to keep `api_breakage_check_enabled: false`? We recently disabled it, as we don't have any public API from this package, only plugins and executables, and the one library here is explicitly _not_ stable.

FranzBusch: Good catch. That was my bad

czechboy0 (approved):
Thanks!

Issue #651, opened 2024-10-14
Generate a convenience method to create a response case without any parameters


czechboy0:
### Motivation

Today, some responses have no headers nor body (for example, 204 No Content).

In generated code it looks like this:

```swift
internal enum Output: Sendable, Hashable {
    internal struct NoContent: Sendable, Hashable {
        /// Creates a new `NoContent`.
        internal init() {}
    }
    case noContent(Operations.addPollAnswer.Output.NoContent)
}
```

And when writing a server handler, adopters have to spell it as:

```
return .noContent(.init())
```

The `(.init())` bit is unnecessary, and we should make this common case prettier.

### Proposed solution

One possible way to fix this (I welcome other ideas, though) is by generating a `static var` for any response cases that have _no required_ values in the initializer. In practice, since response bodies are always required, this would only apply to responses with _no body_ and with _only optional response headers_.

The generated code would look closer to the following:

```swift
internal enum Output: Sendable, Hashable {
    internal struct NoContent: Sendable, Hashable {
        /// Creates a new `NoContent`.
        internal init() {}
    }
    case noContent(Operations.addPollAnswer.Output.NoContent)
    internal static var noContent: Self { .noContent(.init()) } // <<< this is new
}
```

which would allow the server handler to just have:

```swift
return .noContent
```

simonjbeaumont:
Hunch: Agree that this is a nice QOL improvement for this one case.

Thought: we initially tried to have the generator work without too much "look-ahead", i.e. that it wouldn't need to reach into other parts to work out whether it could or couldn't provide a nicety. We've now got a few places where we do that, but each of these will increase the runtime of the generation. I wonder how common this is, because if we have to pay the price of working out that we _can't_ do it for almost every operation, it may not be worth it?

Question: should we consider (instead, or as well as) having the headers parameter `Operations.addPollAnswer.Output.NoContent.init` be defaulted to `.init()` if there are no required headers.

czechboy0:
Discussed offline: let's do this only when there are no headers or body, but do not do it if there are only non-required headers.

czechboy0:
Okay @arthurcro, feel free to pick this up

arthurcro:
Hey @czechboy0! I made some progress on the implementation and I have a draft version working. However, before I open a pull request, I'd like to discuss some specific details with you to ensure I've correctly interpreted the requirements.

Adding the static property to this function won't work because it needs to be generated outside the struct declaration.

[...]

PR #656, opened 2024-10-20
Generate a static property to create a responses without any parameters

arthurcro:
### Motivation

Closes https://github.com/apple/swift-openapi-generator/issues/651.
Today, some responses have no headers nor body (for example, 204 No Content).

In generated code it looks like this:
```swift
internal enum Output: Sendable, Hashable {
    internal struct NoContent: Sendable, Hashable {
        /// Creates a new `NoContent`.
        internal init() {}
    }
    case noContent(Operations.addPollAnswer.Output.NoContent)
}
```

And when writing a server handler, adopters have to spell it as:
```swift
return .noContent(.init())
```

The `(.init())` bit is unnecessary, and we should make this common case prettier.

### Modifications

Adds a static property to the generated `Output` enum for any response which does not have a header or a body and does not require a status code.

### Result

The following is code is generated: 
```swift
internal enum Output: Sendable, Hashable {
    internal struct NoContent: Sendable, Hashable {
        /// Creates a new `NoContent`.
        internal init() {}
    }
    case noContent(Operations.addPollAnswer.Output.NoContent)
    internal static var noContent: Self {
        .noContent(.init())
    }
}
```

### Test Plan

Update `FileBasedReferenceTests.swift` to reflect the changes in the generated `Types.swift`.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponseOutcome.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponseOutcome.swift
@@ -63,7 +63,7 @@ extension TypesFileTranslator {
         let enumCaseDesc = EnumCaseDescription(name: enumCaseName, kind: .nameWithAssociatedValues(associatedValues))
         let enumCaseDecl: Declaration = .commentable(enumCaseDocComment, .enumCase(enumCaseDesc))
 
-        var staticMemberDecl: Declaration?
+        let staticMemberDecl: Declaration?
         let responseHasNoHeaders = typedResponse.response.headers?.isEmpty ?? true
         let responseHasNoContent = typedResponse.response.content.isEmpty
         if responseHasNoContent && responseHasNoHeaders && !responseKind.wantsStatusCode {
@@ -83,6 +83,8 @@ extension TypesFileTranslator {
                 ]
             )
             staticMemberDecl = .commentable(enumCaseDocComment, .variable(staticMemberDesc))
+        } else {
+            staticMemberDecl = nil
         }
 
         let throwingGetterDesc = VariableDescription(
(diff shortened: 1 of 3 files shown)


Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponseOutcome.swift
```diff
@@ -55,14 +56,34 @@ extension TypesFileTranslator {
...
+                    .expression(
+                        .functionCall(
+                            calledExpression: .dot(enumCaseName),
+                            arguments: [.functionCall(calledExpression: .dot("init"))]
+                        )
+                    )
+                ]
+            )
+            staticMemberDecl = .commentable(enumCaseDocComment, .variable(staticMemberDesc))
+        }
```
czechboy0, line 86:
```suggestion
        let staticMemberDecl: Declaration?
        let responseHasNoHeaders = typedResponse.response.headers?.isEmpty ?? true
        let responseHasNoContent = typedResponse.response.content.isEmpty
        if responseHasNoContent && responseHasNoHeaders && !responseKind.wantsStatusCode {
            let staticMemberDesc = VariableDescription(
                accessModifier: config.access,
                isStatic: true,
                kind: .var,
                left: .identifier(.pattern(enumCaseName)),
                type: .member(["Self"]),
                getter: [
                    .expression(
                        .functionCall(
                            calledExpression: .dot(enumCaseName),
                            arguments: [.functionCall(calledExpression: .dot("init"))]
                        )
                    )
                ]
            )
            staticMemberDecl = .commentable(enumCaseDocComment, .variable(staticMemberDesc))
        } else {
            staticMemberDecl = nil
        }
```

czechboy0: Nit: use a `let staticMemberDecl` instead, and assign `nil` to it explicitly. That's the pattern we've used in the generator, to avoid local vars unless we really want to write to them multiple times.

czechboy0 (changes requested):
One small suggestion, otherwise looks great!

Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponseOutcome.swift
arthurcro: That makes sense, I will change that thank you!

czechboy0 (approved):
Looks great, thank you!

PR #658, opened 2024-10-21
Include path item parameters in filter

czechboy0:
### Motivation

When filtering by a specific operationId, any path item-level parameters were not included (only operation-level parameters).

### Modifications

Include path item level parameters.

### Result

No errors for missing references when filtering for operations that have a path item-level parameter with a reference.

### Test Plan

Adapted unit tests to cover this.


simonjbeaumont (approved):
Nice catch, thanks!

PR #659, opened 2024-10-24
[Generator] Add support of deepObject style in query params

arthurcro:
### Motivation

The generator changes for: https://github.com/apple/swift-openapi-generator/issues/259

Depends on https://github.com/apple/swift-openapi-runtime/pull/100 landing first and getting released, and the version dependency being bumped here.

### Modifications

Added deepObject style to serializer & parser in order to support nested keys on query parameters.

### Result

Support nested keys on query parameters.

### Test Plan

Adapted snippet tests (SnippetBasedReferenceTests)

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
@@ -130,7 +130,10 @@ extension FileTranslator {
             let location = parameter.location
             switch location {
             case .query:
-                guard case .form = style else {
+                switch style {
+                case .form, .deepObject:
+                    break
+                default:
                     try diagnostics.emitUnsupported(
                         "Query params of style \(style.rawValue), explode: \(explode)",
                         foundIn: foundIn
@@ -243,6 +246,7 @@ extension OpenAPI.Parameter.SchemaContext.Style {
     var runtimeName: String {
         switch self {
         case .form: return Constants.Components.Parameters.Style.form
+        case .deepObject: return Constants.Components.Parameters.Style.deepObject
         default: preconditionFailure("Unsupported style")
         }
     }
--- a/Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
+++ b/Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
@@ -2514,6 +2514,18 @@ final class SnippetBasedReferenceTests: XCTestCase {
                       type: array
                       items:
                         type: string
+                  - name: sort
+                    in: query
+                    required: true
+                    style: deepObject
+                    explode: true
+                    schema:
+                      type: object
+                      properties:
+                        option1:
+                          type: string
+                        option2:
+                          type: string
                 responses:
                   default:
                     description: Response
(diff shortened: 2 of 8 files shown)


czechboy0:
We discussed this offline, this is the PR https://github.com/apple/swift-openapi-generator/pull/538 reopened again, so that we can finish it up even while the original author is busy 🙂 Thanks @arthurcro for adding you both to the commit, it's important that we give credit to @kstefanou52 for the work already done!

czechboy0:
Kicked off CI to see what state we're in.

Sources/_OpenAPIGeneratorCore/Translator/Parameters/TypedParameter.swift
```diff
@@ -130,7 +130,10 @@ extension FileTranslator {
             let location = parameter.location
             switch location {
             case .query:
-                guard case .form = style else {
+                switch style {
+                case .form, .deepObject:
+                    break
+                default:
```
czechboy0, line 136:
```suggestion
                switch style {
                case .form:
                    break
                case .deepObject where explode:
                    break
                default:
```

czechboy0: style: deepObject + explode: true is supported, but not with explode: false, so let's catch it here and provide a descriptive error.

See in the runtime changes that we only support explode: true: https://github.com/apple/swift-openapi-runtime/pull/100/files#diff-a382c8525d6209dfbc3e90daaf0262692d9df73f1b7a914b9eebe143f04c287dR76

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
```diff
@@ -2524,18 +2536,36 @@ final class SnippetBasedReferenceTests: XCTestCase {
...
+                                case option2
+                            }
+                        }
+                        public var sort: Operations.get_sol_foo.Input.Query.sortPayload
                         public init(
                             single: Swift.String? = nil,
                             manyExploded: [Swift.String]? = nil,
-                            manyUnexploded: [Swift.String]? = nil
+                            manyUnexploded: [Swift.String]? = nil,
+                            sort: Operations.get_sol_foo.Input.Query.sortPayload
```
czechboy0, line 2559:
```suggestion
                            sort: Operations.get_sol_foo.Input.Query.sortPayload = nil
```

This should end up being defaulted to nil as well. It's possible it'll require improvements in inspecting the nested object and whether it can be initialized with `.init()`.

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
```diff
@@ -2524,18 +2536,36 @@ final class SnippetBasedReferenceTests: XCTestCase {
...
                         ) {
                             self.single = single
                             self.manyExploded = manyExploded
                             self.manyUnexploded = manyUnexploded
+                            self.sort = sort
                         }
                     }
                     public var query: Operations.get_sol_foo.Input.Query
-                    public init(query: Operations.get_sol_foo.Input.Query = .init()) {
+                    public init(query: Operations.get_sol_foo.Input.Query) {
```
czechboy0, line 2568:
```suggestion
                    public init(query: Operations.get_sol_foo.Input.Query = .init()) {
```

Yeah this should not go away, since all the query items are still optional, so it should be possible to init the Input with just `.init()`.

czechboy0 (changes requested):
Added a few comments of the next things to fix up, but we're almost there.

kstefanou52:
Sorry, I couldn’t find the time to complete this PR yet. I made some progress, but there seems to be an issue with the unit tests—possibly something with the runtime library. I’ll have some time this weekend, so if @arthurcro would like any help, please let me know.

arthurcro:
> Sorry, I couldn’t find the time to complete this PR yet. I made some progress, but there seems to be an issue with the unit tests—possibly something with the runtime library. I’ll have some time this weekend, so if @arthurcro would like any help, please let me know.

Thanks for the update @kstefanou52! I've got some availability this weekend too and would be happy to help . If you'd prefer to complete it yourself given the progress you've made, that's completely fine - just let me know what works best for you!

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
arthurcro: That parameter has `required: true` though, I don't think it should be defaulted to`nil`. It should be defaulted to `.init()` instead I think?

czechboy0: Oh that's interesting! What does it mean for a query parameter of type object to be required, when all properties are optional?

On the wire, it won't look any different to the whole parameter being optional.

Maybe we should just change the sample to have an optional parameter? I think that's more common. Alternatively, make one of the properties required. Either is fine.

kstefanou52:
> > Sorry, I couldn’t find the time to complete this PR yet. I made some progress, but there seems to be an issue with the unit tests—possibly something with the runtime library. I’ll have some time this weekend, so if @arthurcro would like any help, please let me know.
> 
> Thanks for the update @kstefanou52! I've got some availability this weekend too and would be happy to help . If you'd prefer to complete it yourself given the progress you've made, that's completely fine - just let me know what works best for you!

Thank you so much for the understanding @arthurcro. I'm just giving it a try and I will let you know about the outcome until Sunday end of the day.

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
arthurcro: Right, that's interesting. Updating the sample to have an optional parameter would resolve the issues you highlighted in the review with this test. 

What if all the properties of the object have default values though? I tried getting an idea of what it would look like by updating the tests but I can get any default value to show in the generated swift types.

czechboy0:
Now that #538 landed, I'll close this.

PR #674, opened 2024-11-16
Skip duplicate enum values

czechboy0:
### Motivation

Some OpenAPI docs, usually through imperfect conversion from another representation, end up with duplicate raw values in `enum` schemas.

The OpenAPI specification (by referencing the JSON Schema specification) [says](https://datatracker.ietf.org/doc/html/draft-wright-json-schema-validation-00#section-5.20):

```
The value of this keyword MUST be an array.  This array SHOULD have at least one element.  Elements in the array SHOULD be unique.
```

So elements **should** be unique, but don't have to be. Today the generator fails to generate documents with such duplicate values.

### Modifications

Gracefully handle duplicate values by emitting a warning diagnostic and skipping it.

### Result

This unblocks generating OpenAPI documents with duplicate enum values, such as the [Bluesky OpenAPI doc](https://github.com/bluesky-social/bsky-docs/blob/main/atproto-openapi-types/spec/api.json).

### Test Plan

Added a unit test for duplicates.


theoriginalbit:
I don't believe this applies to enums outside of the schema definitions, should it? 

E.g. omit and raise a warning for duplicates for server enum definitions. I don't remember explicitly testing duplicates when I made the recent changes

czechboy0:
@theoriginalbit correct, this is only for schema enums, as that's what the JSON Schema spec is concerned with.

The server enum values also (inexplicably) aren't required to be empty, so would you mind opening a PR fixing that? https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.4.md#fixed-fields-5

Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateRawEnum.swift
```diff
@@ -42,32 +59,48 @@ extension FileTranslator {
...
-            .map { anyValue in
-                switch backingType {
-                case .string:
-                    // In nullable enum schemas, empty strings are parsed as Void.
-                    // This is unlikely to be fixed, so handling that case here.
-                    // https://github.com/apple/swift-openapi-generator/issues/118
-                    if isNullable && anyValue is Void { return (context.asSwiftSafeName(""), .string("")) }
+        var seen: Set<EnumValue> = []
+        var cases: [(String, LiteralDescription)] = []
+        func shouldAdd(_ value: EnumValue) throws -> Bool {
```
simonjbeaumont, line 64: This maybe-add-to-dictionary flow, that maintains a set seems unnecessarily complex.

We already had an array of key-value pairs, which we can then just pass to `Dictionary.init(_:uniquingKeysWith:)`:

```swift
        let dict = Dictionary(cases, uniquingKeysWith: { first, _ in
            // emit diagnostic
            return first
        })
```

Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateRawEnum.swift
```diff
@@ -23,6 +23,23 @@ enum RawEnumBackingType {
...
+
+    /// An integer value.
+    case integer(Int)
+
+    var description: String {
+        switch self {
+        case .string(let value): return "\"\(value)\""
+        case .integer(let value): return String(value)
+        }
+    }
```
simonjbeaumont, line 40: Is `description` the right thing to use to generate the representation we need to be safe in generated code? Should we promote this to a more formally named computed property?

simonjbeaumont (changes requested):
Good catch. Wondering if we can simplify the logic a bit though.

Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateRawEnum.swift
czechboy0: This is only used in the diagnostic description, we have custom code for creating the Swift-safe names.

Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateRawEnum.swift
czechboy0: We'd lose the user-specified ordering of cases from the OpenAPI document and the generated enum could have a different order every time it's generated.

In theory we could use OrderedDictionary from OpenAPIKit, but we'd only construct it to get the duplicate cases filtered out, and immediately pull the key/value pairs as tuples back out into an array and pass them to `translateRawRepresentableEnum`. Should I do that?

simonjbeaumont: > We'd lose the user-specified ordering of cases from the OpenAPI document...

I don't think that's a problem... but...

> ...and the generated enum could have a different order every time it's generated.

...I concede that _this_ is a problem.

Could we use `OrderedSet` for this for the `cases`, rather than rolling this?

czechboy0: Ok updated

simonjbeaumont (approved):
Thanks for taking the time to polish this!

PR #678, opened 2024-11-21
Make error message for missing "openapi" key clearer.

deirdresmtoo:
Accidentally hit this error message due to a copy/paste oops and found the resulting error message confusing.

### Motivation

Initially when I saw the message, I thought of key as in public/private keys, not key as in key/value pair.

### Modifications

Slight change in wording to make it clearer that a key with the name openapi is what's missing.

### Result

Error message will be updated.

### Test Plan

Updated test results also and verified tests pass.


czechboy0 (approved):
Great point, thank you!

PR #686, opened 2024-12-02
Enable strict concurrency

czechboy0:
### Motivation:

Catch potential data races at build time.

### Modifications:

- Enabled unconditional strict concurrency complete checking.
- Enabled warnings-as-errors on Swift 6 Linux pipelines.
- Made a few fixes to be strict concurrency-clean.

### Result:

Fewer potential data races can sneak in.

### Test Plan

Ran tests locally, did not see any more warnings or errors.


glbrntt (approved):
LGTM

PR #688, opened 2024-12-05
Updated examples and integration test with the new servers syntax

czechboy0:
### Motivation

A drive-by fixup, I noticed the warnings when running the integration test locally.

### Modifications

Fixed up the examples and integrate test to use the new syntax. This also meant regenerating the two examples that check in generated code, which took in a few more changes since 1.0.

### Result

Syntax matches our current best practices.

### Test Plan

Looks good locally.


simonjbeaumont (approved):
LGTM, thanks!

PR #694, opened 2024-12-09
Enable MemberImportVisibility check on all targets

rnro:
Enable MemberImportVisibility check on all targets. Use a standard string header and footer to bracket the new block for ease of updating in the future with scripts.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -147,3 +147,14 @@ let package = Package(
         ),
     ]
 )
+
+// ---    STANDARD CROSS-REPO SETTINGS DO NOT EDIT   --- //
+for target in package.targets {
+    if target.type != .plugin {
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -147,3 +147,11 @@ let package = Package(
         ),
     ]
 )
+
+for target in package.targets {
+    if target.type != .plugin {
+        var settings = target.swiftSettings ?? []
+        settings.append(.enableUpcomingFeature("MemberImportVisibility"))
```
czechboy0, line 154: Could we add this to the `swiftSettings` variable at the top?

czechboy0 (approved):

czechboy0 (approved):

simonjbeaumont (approved):

PR #696, opened 2024-12-10
Assign to properties with explicit self in `init(from decoder:)`

### LINKED ISSUE #695: Can't use `container` as a property name of a model
### Description

In our API, one of the model has a `container` property, this conflict with the `container` from the `decoder`.

Capture d’écran 2024-12-10 à 13 41 43

wouldn't using `self.` help in mitigating this problem ? 

### Reproduction

openapi-generator-config.yaml

```
generate:
  - types
  - client
accessModifier: public
```

### Package version(s)

- swift-openapi-generator : 1.5.0
- swift-openapi-runtime: 1.7.0
- swift-openapi-urlsession: 1.0.2

### Expected behavior

I can use `container` as a property of a model in my schema

### Environment

```
swift-driver version: 1.115 Apple Swift version 6.0 (swiftlang-6.0.0.9.10 clang-1600.0.26.2)
Target: arm64-apple-macosx15.0
```

### Additional information

_No response_

simonjbeaumont:
### Motivation

Our generated `init(from decoder:)` assigns to properties without using explicit self. This is a problem because it can produce conflicts with the local variables created in the initializer, e.g. the decoding container, which will then fail to compile for schemas with properties named `container`, as found by an adopter in #695.

We should probably take a broader pass with regard to the use of explicit self in our generated code, but for now we can resolve this by focussing on this problematic generation, where a schema with named and additional properties.

### Modifications

- Assign to properties with explicit self in `init(from decoder:)`

### Result

- Fixes #695.

### Test Plan

- Add a snippet test.
- Update the reference tests.


czechboy0 (approved):
Thanks for the quick fix! 🙏

PR #698, opened 2024-12-11
Assign to properties with explicit self in encode(to encoder: any Encoder)

Dragna:
### Motivation

The generated `encode(to encoder: any Encoder)` assigns to properties without using explicit self. This is a problem because it can produce conflicts with the local variables created in the function, e.g. the encoding container, which will then fail to compile for schemas with properties named `container`, as found in https://github.com/apple/swift-openapi-generator/issues/695.

### Modifications

- Assign to properties with explicit self in `encode(to encoder: any Encoder)`

### Test Plan

- Update the reference tests.

@simonjbeaumont it seems to be working on my project with this change, sadly I didn't know what updates I should make for the snippet tests 😕. I you have any insight on this that would be helpful 🙏 .


djnivek (approved):
Well done! 👏

simonjbeaumont:
@Dragna thanks for taking the time to put the patch up. I think we raced on this one. There's another PR up which adds the relevant snippet and reference tests: #699.

PR #699, opened 2024-12-11
Access properties with explicit self in `encode(to encoder:)`

simonjbeaumont:
### Motivation

In #696 we avoided clashing with a property called `container` in certain schemas by making use of explicit self in `init(from decoder:)`. We didn't have access to the OpenAPI document and hadn't caught the case where this was used in a schema that supported additional properties, where we also need to avoid the clash in `encode(to encoder:)`.

### Modifications

- Access properties with explicit self in `encode(to encoder:)`.

### Result

- Fixes #695 

### Test Plan

More snippet tests and reference tests.

PR #700, opened 2024-12-11
Use self for member access/assignment in all initializers, decoders, and encoders

simonjbeaumont:
### Motivation

In order to address #695 we made some targeted fixes (#696, #699) to use explicit-self to avoid clashing with schema property names. This addressed a clash with a specific variable with properties of only certain schemas, but there are more places where we might encounter future issues. Specifically, we have other variable names which might cause a clash (e.g. `storage` and `discriminator`), and there are other flavours of initializers, encoders, and decoders that were not updated in the targeted fix.

 In this PR, we update all access and assignment in the generated code dealing with schemas to use explicit-self.

### Modifications

- Use self for member access and assignment in all initializers, decoders, and encoders.
- Update the reference tests and snippet tests.

### Result

Removed some cases where valid OpenAPI docs would produce non-compiling code.

### Test Plan

Snippet and reference tests. The latter is actually compiled during the CI.


czechboy0 (approved):
Looks great, thanks! 🙏

PR #706, opened 2024-12-19
Update Examples and IntegrationTest to use idiomatic naming

czechboy0:
### Motivation

Since idiomatic naming is our new recommended setting, use it in Examples and IntegrationTest.

### Modifications

Added idiomatic naming to all examples and made any necessary changes. Plus a few modernizations, like moving to Hummingbird 2.0 and to Prometheus 2.0.

### Result

More up-to-date examples, easier to copy out by adopters, they now get our recommended starting settings.

### Test Plan

Ran and tested locally.


Examples/tracing-middleware-example/Sources/HelloWorldVaporServer/HelloWorldVaporServer.swift
```diff
@@ -16,36 +16,45 @@ import OpenAPIVapor
...
         let handler = Handler()
         try handler.registerHandlers(on: transport, serverURL: URL(string: "/api")!, middlewares: [TracingMiddleware()])
-        try await app.execute()
+        try await withThrowingTaskGroup(of: Void.self) { group in
+            group.addTask { try await app.execute() }
+            group.addTask { try await tracer.run() }
+            group.addTask { try await processor.run() }
+            _ = try await group.next()
+            group.cancelAll()
+        }
```
simonjbeaumont, line 58: Can we drop a comment pointing people to Swift Service Lifecycle here, because this probably isn't what we want folks to do in practice, right?

```suggestion
        // Consider using Swift Service Lifecycle — https://github.com/swift-server/swift-service-lifecycle
        try await withThrowingTaskGroup(of: Void.self) { group in
            group.addTask { try await app.execute() }
            group.addTask { try await tracer.run() }
            group.addTask { try await processor.run() }
            _ = try await group.next()
            group.cancelAll()
        }
```

simonjbeaumont (approved):

PR #708, opened 2024-12-21
Add a CLI option for selecting the naming strategy

czechboy0:
### Motivation

The recently introduced naming strategy is only configurable in the config file, but pure CLI users might also want to use it without creating a config file.

### Modifications

Add a `--naming-strategy` option to the `swift-openapi-generator` CLI.

### Result

Users can use the new naming strategy without creating a config file, if they didn't have one before.

### Test Plan

Tested manually on a sample project.

### DIFF
--- a/Sources/swift-openapi-generator/GenerateOptions.swift
+++ b/Sources/swift-openapi-generator/GenerateOptions.swift
@@ -28,10 +28,13 @@ struct _GenerateOptions: ParsableArguments {
             "The Swift files to generate. Options: \(GeneratorMode.prettyListing). Note that '\(GeneratorMode.client.rawValue)' and '\(GeneratorMode.server.rawValue)' depend on declarations in '\(GeneratorMode.types.rawValue)'."
     ) var mode: [GeneratorMode] = []
 
+    @Option(help: "The access modifier to use for the API of generated code. Default: \(Config.defaultAccessModifier.rawValue)")
+    var accessModifier: AccessModifier?
+
     @Option(
         help:
-            "The access modifier to use for the API of generated code. Default: \(Config.defaultAccessModifier.rawValue)"
-    ) var accessModifier: AccessModifier?
+            "The strategy for converting OpenAPI names into Swift names. Default: \(Config.defaultNamingStrategy.rawValue)"
+    ) var namingStrategy: NamingStrategy?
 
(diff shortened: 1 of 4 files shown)


Sources/swift-openapi-generator/GenerateOptions.swift
```diff
@@ -78,7 +84,10 @@ extension _GenerateOptions {
     /// Returns the naming strategy requested by the user.
     /// - Parameter config: The configuration specified by the user.
     /// - Returns: The naming strategy requestd by the user.
-    func resolvedNamingStrategy(_ config: _UserConfig?) -> NamingStrategy { config?.namingStrategy ?? .defensive }
+    func resolvedNamingStrategy(_ config: _UserConfig?) -> NamingStrategy {
+        if let namingStrategy { return namingStrategy }
+        return config?.namingStrategy ?? Config.defaultNamingStrategy
+    }
```
simonjbeaumont, line 90: Just calling out that we've picked up some asymmetry here with regard to how we handle `accessModifier` and `namingStrategy`.

Both of these have defaults, defined as statics in `Config` and both of these must have a value (unlike other options, e.g. filter, where `nil` is reasonable).

The `resolveAccessModifier` and `resolveNamingStrategy` differ with regard to whether they return an optional.

The `accessModifier` and `namingStrategy` parameters to `Config.init` also differ: the former expects an argument, moving the use of default to the call site; but the latter has the default value in the parameter list of the initializer.

IMO we should clean this up rather than accumulate some unnecessary inconsistency here.

But this doesn't need to be done as part of this PR. Feel free to convert this to an issue.

simonjbeaumont (approved):
Great, this is a useful addition!

Added some feedback for some inconsistency that's crept in (but not just in this PR). We can address it out of band if you prefer.

Sources/swift-openapi-generator/GenerateOptions.swift
simonjbeaumont: Thanks, what about the use of the default in `Config.init`? No need to keep that API stable IIUC.

czechboy0: Whoops noticed this only after the auto-merge, fixed here: https://github.com/apple/swift-openapi-generator/pull/713

PR #709, opened 2024-12-21
Add missing imports to fix Swift 6.1+

czechboy0:
### Motivation

Currently unit tests on Swift 6.1+ toolchains are failing due to a few missing imports.

That these are coming up is good, that's why we've enabled `MemberImportVisibility`.

### Modifications

Add the missing imports to stop relying on implicitly imported extension methods.

### Result

Unit tests compile again on Swift 6.1+.

### Test Plan

Tested locally, will further verify CI is happy.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -29,6 +29,7 @@ jobs:
         with:
             name: "Integration test"
             matrix_linux_command: "apt-get update -yq && apt-get install -yq jq && SWIFT_OPENAPI_GENERATOR_REPO_URL=file://${GITHUB_WORKSPACE} ./scripts/run-integration-test.sh"
+            matrix_linux_nightly_main_enabled: false
 
     compatibility-test:
       name: Compatibility test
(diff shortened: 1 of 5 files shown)


.github/workflows/pull_request.yml
```diff
@@ -56,4 +55,3 @@ jobs:
         with:
             name: "Example packages"
             matrix_linux_command: "./scripts/test-examples.sh"
-            matrix_linux_nightly_main_enabled: false
```
simonjbeaumont, line 59: I'm +1 on us having the nightly main enabled for unit tests, but do we gain much by having the integration test and examples here, other than noise. Especially since the examples pipeline is currently the slowest, so the limiting factor on getting PRs green, and IIUC the nightly main setup time is longer than all others because it's likely it will not have a warm image.

simonjbeaumont (approved):
I'm not convinced we want the pipeline on the examples or integration test matrix, but I'll leave that up to you.

.github/workflows/pull_request.yml
czechboy0: Fair enough, I'll disable the examples/int tests ones. I was mainly interested in seeing unit test results.

czechboy0: 34311fd

PR #713, opened 2025-01-06
[Cleanup] Make Config's namingStrategy and accessModifier params consistent

czechboy0:
### Motivation

A missed fixup from #708: https://github.com/apple/swift-openapi-generator/pull/708#discussion_r1903932794

### Modifications

Removed the default from Config.init and forced all callers to make a choice on the naming strategy, just like for accessModifier.

### Result

Consistent handling of these two parameters.

### Test Plan

Tests passed locally.


simonjbeaumont (approved):
Great—thanks!

PR #731, opened 2025-02-07
Update README.md

bdiesel:
This commit fixes an issue when starting Vapor in swift 6.  I believe earlier versions of Swift throw a warning about this.

### Motivation

I was following along with documentations but Vapor was throwing an error: Initializer 'init' is unavailable from asynchronous contexts; This initialiser cannot be used in async contexts, use Application.make(_:_:) instead

### Modifications

Just a very simple update to the documentation so others getting started don't hit the same issue

### Result

Vapor server will start when following docs.

### Test Plan

manual tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ struct Handler: APIProtocol {
 
 @main struct HelloWorldVaporServer {
     static func main() async throws {
-        let app = Vapor.Application()
+        let app = try await Application.make(.detect())
         let transport = VaporTransport(routesBuilder: app)
         let handler = Handler()
         try handler.registerHandlers(on: transport, serverURL: URL(string: "/api")!)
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -66,7 +66,7 @@ struct Handler: APIProtocol {
 
 @main struct HelloWorldVaporServer {
     static func main() async throws {
-        let app = Vapor.Application()
+        let app = try await Application.make(.detect())
```
czechboy0, line 69: Is the `.detect()` part necessary?

bdiesel: It's not necessary, it's from another example I was referencing from the Vapor docs.  Let me remove that part.  Thank you for reviewing this!

czechboy0:
Thanks @bdiesel, I believe there are a few more places in the repo, could you grep for any other occurrences? The docc landing page might also have this example, possibly also tutorials and examples. Thanks!

czechboy0 (approved):
Thank you!

PR #733, opened 2025-02-09
[Docs] Error handling

czechboy0:
### Motivation

Depends on: https://github.com/apple/swift-openapi-runtime/pull/141

_[Explain here the context, and why you're making that change. What is the problem you're trying to solve.]_

### Modifications

_[Describe the modifications you've made.]_

### Result

_[After your change, what will change.]_

### Test Plan

_[Describe the steps you took, or will take, to qualify the change - such as adjusting tests and manual testing.]_

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -54,7 +54,8 @@ let package = Package(
         // Tests-only: Runtime library linked by generated code, and also
         // helps keep the runtime library new enough to work with the generated
         // code.
-        .package(url: "https://github.com/apple/swift-openapi-runtime", from: "1.3.2"),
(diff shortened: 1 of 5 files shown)


Package.swift
```diff
@@ -54,7 +54,8 @@ let package = Package(
         // Tests-only: Runtime library linked by generated code, and also
         // helps keep the runtime library new enough to work with the generated
         // code.
-        .package(url: "https://github.com/apple/swift-openapi-runtime", from: "1.3.2"),
+//        .package(url: "https://github.com/apple/swift-openapi-runtime", from: "1.3.2"),
+        .package(path: "../swift-openapi-runtime"),
```
czechboy0, line 58: TODO: undo once we have a runtime release with the error mappers

PR #734, opened 2025-02-09
[Bug] Exclude webhooks when filter is provided

### LINKED ISSUE #712: Problems with GitHub's OpenAPI 3.1 spec
### Description

```bash
Considering target 'GitHubAPI':
- Trying OpenAPI code generation.
Error: Failed to satisfy: JSONSchema reference can be found in components/schemas at path: .webhooks.branch-protection-configuration-disabled.post.requestBody.content['application/json'].schema
Failed to satisfy: JSONSchema reference can be found in components/schemas at path: .webhooks.branch-protection-configuration-enabled.post.requestBody.content['application/json'].schema
```
There are a lot more similar `Failed to satisfy` errors.

### Reproduction

Try to use the OpenAPI Generator on https://github.com/github/rest-api-description/blob/main/descriptions-next/api.github.com/api.github.com.yaml

### Package version(s)

generator 1.6.0
runtime 1.7.0

### Expected behavior

work.

### Environment

Swift 6.0.3

### Additional information

_No response_

czechboy0:
### Motivation

Fixes #712.

When an OpenAPI document includes webhooks that include schema references, and a filter is provided, the generator would fail unless all the schemas needed by all webhooks were explicitly included in the filter.

### Modifications

Since we don't currently support filtering webhooks, just exclude them always when a filter is provided. This fixes the bug.

In the future, if we add support for webhooks, we can add support for filtering them as well.

### Result

Unblocks generating GitHub's OpenAPI 3.1.0 doc + filters.

### Test Plan

Adapted the filtering unit test, verified it fails without the fix, and passes with it.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Hooks/FilteredDocument.swift
+++ b/Sources/_OpenAPIGeneratorCore/Hooks/FilteredDocument.swift
@@ -128,6 +128,9 @@ struct FilteredDocumentBuilder {
             }
         }
         filteredDocument.components = components
+        // Always remove webhooks, as they may contain references to schemas that were not included.
+        // Webhooks are not supported for filtering, so whenever a filter is used, all webhooks are excluded.
+        filteredDocument.webhooks = [:]
         return filteredDocument
     }
 
--- a/Tests/OpenAPIGeneratorCoreTests/Hooks/Test_FilteredDocument.swift
+++ b/Tests/OpenAPIGeneratorCoreTests/Hooks/Test_FilteredDocument.swift
@@ -75,6 +75,13 @@ final class Test_FilteredDocument: XCTestCase {
                         $ref: '#/components/schemas/B'
                 Empty:
                   description: success
+            webhooks:
+              my-webhook:
+                post:
+                  description: Webhook
+                  responses:
+                    '200':
+                      description: OK
             """
         let document = try YAMLDecoder().decode(OpenAPI.Document.self, from: documentYAML)
         assert(filtering: document, filter: DocumentFilter(), hasPaths: [], hasOperations: [], hasSchemas: [])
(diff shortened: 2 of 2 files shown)


simonjbeaumont:
> Since we don't currently support filtering webhooks, just exclude them always when a filter is provided. This fixes the bug.
> 
> In the future, if we add support for webhooks, we can add support for filtering them as well.

It's true that we don't explicitly support filtering _only_ webhooks, but looking at the OAS, `webhooks:` is a `Map[string, Path Item Object]`[^0], and path items contain operations, which, in turn _can_ have operation IDs and/or tags.

[^0]: https://spec.openapis.org/oas/latest.html#fixed-fields

The filtering feature IMO is document-centric and so should respect the document, especially since we provide a standalone command for `filter`.

IMO the bug here is that we do not filter the webhooks according to the filter parameters, which we should.

czechboy0:
If we want to properly support filtering of webhooks in this PR, then it'll be a bit more work. For example, the [`allOperationIds`](https://github.com/mattpolzin/OpenAPIKit/blob/749ceba9a6e91d40081f27996fd91fd3de9a8dfe/Sources/OpenAPIKit/Document/Document.swift#L234) property on Document in OpenAPIKit doesn't include operations from webhooks, which will break for example [this invariant](https://github.com/apple/swift-openapi-generator/blob/1db77dd7e9ec2bccf0245c46c0666067be5bf676/Sources/_OpenAPIGeneratorCore/Hooks/FilteredDocument.swift#L176), if we allow operationIds from webhook operations. Not sure if that means we should wait for a fix in OpenAPIKit (cc @mattpolzin) or if we should try to reimplement it on our end.

I guess that means that the standalone filter command on the CLI doesn't work today on documents that contain webhooks.

mattpolzin:
Happy to fix `allOperationIds` to include webhooks. I’ve got the work tracked [here](https://github.com/mattpolzin/OpenAPIKit/issues/392).

mattpolzin:
Fixed in [OpenAPIKit 3.4.1](https://github.com/mattpolzin/OpenAPIKit/releases/tag/3.4.1)

simonjbeaumont:
Once again @mattpolzin, thank you! 🙏

czechboy0:
Thanks @mattpolzin!

I started implementing the webhook filtering properly, but it's quickly growing into more work, so I'll have to put this aside for now and come back to when I have more time. Or someone else can pick it up if we need it sooner. For now we'll need to recommend adopters use the 3.0 GitHub OpenAPI doc, rather than 3.1, so they shouldn't be blocked.

PR #737, opened 2025-02-12
Adjust the platform-based linkage check

jakepetroules:
Limit this to Apple platforms except macOS in order to unblock use of swift-openapi-generator on more platforms like Windows, Android, FreeBSD, and so on.


czechboy0 (approved):
Thanks!

simonjbeaumont (approved):

PR #742, opened 2025-03-07
Only apply standard swift settings on valid targets

rnro:
Only apply standard swift settings on valid targets. The current check ignores plugins but that is not comprehensive enough.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -150,11 +150,13 @@ let package = Package(
 
 // ---    STANDARD CROSS-REPO SETTINGS DO NOT EDIT   --- //
 for target in package.targets {
-    if target.type != .plugin {
+    switch target.type {
+    case .regular, .test, .executable:
         var settings = target.swiftSettings ?? []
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -150,11 +150,13 @@ let package = Package(
...
         var settings = target.swiftSettings ?? []
         // https://github.com/swiftlang/swift-evolution/blob/main/proposals/0444-member-import-visibility.md
         settings.append(.enableUpcomingFeature("MemberImportVisibility"))
         target.swiftSettings = settings
+    case .macro, .plugin, .system, .binary: ()  // not applicable
+    @unknown default: ()  // we don't know what to do here, do nothing
     }
-}
-// --- END: STANDARD CROSS-REPO SETTINGS DO NOT EDIT --- //
+}// --- END: STANDARD CROSS-REPO SETTINGS DO NOT EDIT --- //
```
rnro, line 162:
```suggestion
}
// --- END: STANDARD CROSS-REPO SETTINGS DO NOT EDIT --- //
```

czechboy0 (approved):

PR #752, opened 2025-03-28
Only run Darwin CI on macOS targets

rnro:
Only run Darwin CI on macOS targets as that's the only supported one.


czechboy0 (approved):

simonjbeaumont (approved):
Thanks @rnro!

PR #764, opened 2025-04-24
Allow substituting types

simonbility:
### Motivation

Picking up after some time from this issue https://github.com/apple/swift-openapi-generator/issues/375

There are some usecases described here that i think could be addressed by this.

I suspect there are some "bigger picture" decisions (maybe proposals) needing to happen so i wanted to get the ball rolling :) 

### Modifications

I made a small change allowing to "swap in" any type instead of generating one, by using a vendor-extension (`x-swift-open-api-substitute-type`)

### Result

The following spec 

```yaml
openapi: 3.1.0
info:
  title: api
  version: 1.0.0
components:
  schemas:
    MyCustomString:
      type: string
      x-swift-open-api-substitute-type: MyLibrary.MyCustomString
```
would generate code like this (abbreviated)
```swift
public enum Components {
    public enum Schemas {
        /// - Remark: Generated from `#/components/schemas/MyCustomString`.
        public typealias MyCustomString = MyLibrary.MyCustomString
    }
}
```

### Test Plan

I did write a test but suspect theres, other parts affected that i missed

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateSchema.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateSchema.swift
@@ -87,6 +87,21 @@ extension TypesFileTranslator {
                 )
             )
         }
+        if let substituteTypeName = schema.vendorExtensions["x-swift-open-api-substitute-type"]?.value
+            as? String
+        {
+            try diagnostics.emit(.note(message: "Substituting type \(typeName) with \(substituteTypeName)"))
+            let substitutedType = TypeName(swiftKeyPath: substituteTypeName.components(separatedBy: ".")).asUsage
+            
+            let typealiasDecl = try translateTypealias(
+                named: typeName,
+                userDescription: overrides.userDescription ?? schema.description,
+                to: substitutedType.withOptional(
+                    overrides.isOptional ?? typeMatcher.isOptional(schema, components: components)
+                )
+            )
+            return [typealiasDecl]
+        }
 
         // If this type maps to a referenceable schema, define a typealias
         if let builtinType = try typeMatcher.tryMatchReferenceableType(for: schema, components: components) {
--- a/Tests/OpenAPIGeneratorCoreTests/Translator/TypesTranslator/Test_translateSchemas.swift
+++ b/Tests/OpenAPIGeneratorCoreTests/Translator/TypesTranslator/Test_translateSchemas.swift
@@ -47,4 +47,30 @@ class Test_translateSchemas: Test_Core {
             XCTAssertEqual(collector.diagnostics.map(\.description), diagnosticDescriptions)
         }
     }
+    
+    func testSchemaTypeSubstitution() throws {
+        let typeName = TypeName(swiftKeyPath: ["Foo"])
+
+        let schema = try loadSchemaFromYAML(
+            #"""
+            type: string
+            x-swift-open-api-substitute-type: MyLibrary.MyCustomType
+            """#
+        )
+        let collector = AccumulatingDiagnosticCollector()
+        let translator = makeTranslator(diagnostics: collector)
+        let translated = try translator.translateSchema(typeName: typeName, schema: schema, overrides: .none)
+
+        XCTAssertEqual(
+            collector.diagnostics.map(\.description),
+            ["note: Substituting type Foo with MyLibrary.MyCustomType"]
+        )
+        XCTAssertTrue(translated.count == 1, "Should have one translated schema")
+        guard case let .typealias(typeAliasDescription) = translated.first?.strippingTopComment else {
+            XCTFail("Expected typealias description got")
+            return
+        }
+        XCTAssertEqual(typeAliasDescription.name, "Foo")
+        XCTAssertEqual(typeAliasDescription.existingType, .member(["MyLibrary", "MyCustomType"]))
+    }
 }


simonjbeaumont:
Sorry this went unacknowledged until now, and thank you for taking the time to think about this.

> I suspect there are some "bigger picture" decisions (maybe proposals) needing to happen so i wanted to get the ball rolling :)

I'm lightly in favour of this approach since, until we have a holistic solution, it provides an escape-hatch.

We've also previously mentioned that extensions might be a suitable solution to some of these problems: https://github.com/apple/swift-openapi-generator/issues/375#issuecomment-2367484485.

It definitely opens up a bunch of questions and caveats. I see this was motivated by the discussion on providing more strongly typed scalar types and agree this is tempting for things like using a UUID type, instead of a string. My biggest concern would be how this would be used with complex schemas and the burden this shifts to adopters for providing a type that satisfies the schema, especially when that schema is itself referenced in other schemas.

Relatedly there was some work in #627 to move targeted types over which is something I'm supportive of but was lacking an API evolution story. Potentially the config to opt-in might be good enough.

Do you see this feature only being valuable if it can provide substitution for richer types or is there value exploring something more targeted?

@czechboy0 WDYT to this too?

simonbility:
> Sorry this went unacknowledged until now, and thank you for taking the time to think about this.

No worries at all! i appreciate the response.

I also would classify it as a "escape-hatch" solution and it is inherently "unsafe" as you are relying on manually ensuring that it adheres to the schema...

and i would happily take that tradeoff (obviously 😅)

In our project, where we would like to migrate to generating our models, instead of hand-maintaining them.

The main reasons this escape-hatch would be welcomed are:
* strong-scalars (this is a pretty common and established pattern in our codebase)
* types from external libraries (we would deem it "good enough" to defer the en/de-coding to the library)

I can see this approach breaking down if the schema uses custom "format" without defining a custom named schema.

This would IMO something worth-wile to investigate as well.
In our case this isn't a issue since the spec is in our control, but could be problematic when using a spec vendored by a third-party.

```yaml
 User:
    type: object
    properties:
      name:
        type: string
      avatar_url: # there is (currently) no way to swap out this type
        type: string
        format: uri
```

czechboy0:
I think it's a reasonable approach, yeah. An important decision will be whether this would only support primitive type replacement, ie to solve the typesafe scalars problem, or if it'd be supported to replace even object and other composite types - would be nice, just a heads up that it might be more involved implementation-wise. It'll mean not generating inline types, not validating them, etc.

But I'm in support of this in principle - to use vendor extensions to override which type reference is generated.

czechboy0:
Might be easiest for you to try to create a draft PR, while ensuring all tests still pass? Will be easier to iterate there. Once you have confidence that your approach works, writing a lightweight proposal would be great (we have a template for those). Thanks for looking into this!

simonbility:
This PR already contains a POC version of the proposed changes, and the tests pass.

Should i create a dedicated github-issue for this as well? (and a additional PR introducing the evolution proposal)

czechboy0:
I see, would you be able to add more tests, for other schema types? Especially: object, allOf/oneOf/anyOf (including nesting), nested object types, things like that. I'd also propose a slightly different name: `x-swift-openapi-name-override` (and we can further refine that during the proposal review).

simonbility:
@simonjbeaumont @czechboy0 I added some more tests and also opened a PR on `OpenAPIKit` to address a issue preventing this from working on `any/all/one-Of` types https://github.com/mattpolzin/OpenAPIKit/pull/409

I also have written a draft of an evolution proposal (currently this is in this PR and i can move that to a separate PR)
https://github.com/apple/swift-openapi-generator/blob/108386aa79dcc00dfaec78505da335d00613bbee/Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0014.md

In writing this i am reconsidering a bit if vendor-extensions are the best approach to this as the config approach in "Alternatives Considered" also seems quite straightforward, while also being easily usable with third-party open-api documents

Any Feedback on the implementation or the proposal would be welcom

czechboy0:
This is great progress, @simonbility! 

Since this might end up being a pretty large change, once all said and done, I wonder if we could consider scoping the first iteration down a bit, and extending it later in a separate PR/extension.

Specifically, I wonder if the most constrained version that still fulfills a big chunk of the requirement could look like:
1. Only support replacing top level types in `#/components/schemas`, so postpone supporting schemas anywhere else till later.
2. Use a config mapping from the original schema name to a replacement Swift type, rather than a vendor extension.

While that might require some OpenAPI docs to get changed, it'd just be to move some inline types to be named top level types. Otherwise no change needed, and it'd be config driven.

I suspect that'll end up a much smaller implementation, because only the generation of Types.swift needs to be changed, and only in the one place where we iterate through top level types. It should also play nicely with both filtering and cycle detection.

If we can deliver this successfully, that'll prove out the concept without closing any doors for future enhancements. WDYT, @simonbility @simonjbeaumont?

simonbility:
Im perfectly happy to keep this scoped to top-level types with the config based mapping.

Thats fairly close to my initial version where only later on i added inline properties,… for "completeness" sake.


Ill make the adaptations for the scoped down version at some point in the following days.

czechboy0:
> for example when converting to a schema to query-parameters the generator needs to be "fully aware" of the schema (and cannot delegate it to Codable and assume it will be compliant)

Can you elaborate on this point? My assumption was that the adopter is responsible for making sure that their replacement type is indeed compatible with the replaced type. For example, if you replace an object with a string, the generator will let you do it (because it doesn't know what the type will be at runtime that you're replacing it with), but it likely won't work.

What's the awareness that you think the generator should have here?

simonbility:
> > for example when converting to a schema to query-parameters the generator needs to be "fully aware" of the schema (and cannot delegate it to Codable and assume it will be compliant)
> 
> Can you elaborate on this point? My assumption was that the adopter is responsible for making sure that their replacement type is indeed compatible with the replaced type. For example, if you replace an object with a string, the generator will let you do it (because it doesn't know what the type will be at runtime that you're replacing it with), but it likely won't work.
> 
> What's the awareness that you think the generator should have here?

I assumed wronglythe code converting a type to query parameters was generated by the generator, but i now see that its actually based on top of Codable types.
Sorry for the confusion

simonbility:
I would still have a question about how to approach the configuration.

Short JSON Name
```yaml
typeOverrides:
  UUID: Foundation.UUID
```

Fully qualified JSON Path
```
typeOverrides:
  '#/components/schemas/UUID': UUID
```

The fully qualified JSON Path theoretically allows overriding not only `#/comonents/schemas`.
But i am doubtful that actually makes sense.
Most of the other components have some extra fields in them (for instance. `responses` have the contentTypes dict, ...) and usually wrap a schema in some form.

Based on that my approach would be to keep this scoped to '#/components/schemas' and go with the Short JSON Name approach.

This is also whats implemented at the moment.

Do you agree with that?

czechboy0:
Mostly yes, I do wonder if we can make more room for others with:

```
typeOverrides:
  schemas:
    Foo: Bar.Baz
```

And in the future, if we support inline schemas, we'll recognize it from the key being a full JSON pointer, rather than just a single word.

simonbility:
I like that approach 👍 

Allows this to stay narrowly focused on schemas while not closing any doors for future extensions

simonbility:
Made the adaptations here https://github.com/apple/swift-openapi-generator/pull/764/commits/f14f7101c9beca72e148f84c798c373d05b4fd5b

Examples/replace-types-example/Package.swift
```diff
@@ -0,0 +1,37 @@
+// swift-tools-version:5.9
```
czechboy0, line 1:
```suggestion
// swift-tools-version:5.10
```

We've bumped the minimum version since this PR was opened 🙂

Examples/replace-types-example/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+This allows you to write extensions or other helper functions that use these
+types and use them in both the client and server code.
+
+## Usage
+
+Build and run the server using:
+
+```console
+% swift run hello-world-server
```
czechboy0, line 21: This might need updating, since this example doesn't currently have any executable targets.

Sources/_OpenAPIGeneratorCore/Translator/CommonTypes/Constants.swift
```diff
@@ -372,7 +372,7 @@ enum Constants {
         /// The substring used in method names for the multipart coding strategy.
         static let multipart: String = "Multipart"
     }
-
+
```
czechboy0, line 375:
```suggestion

```

Sources/_OpenAPIGeneratorCore/TypeOverrides.swift
```diff
@@ -0,0 +1,25 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftOpenAPIGenerator open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftOpenAPIGenerator project authors
```
czechboy0, line 5:
```suggestion
// Copyright (c) 2025 Apple Inc. and the SwiftOpenAPIGenerator project authors
```

Sources/_OpenAPIGeneratorCore/TypeOverrides.swift
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+public struct TypeOverrides: Sendable {
```
czechboy0, line 15: Please add a doc comment for the type

Tests/OpenAPIGeneratorCoreTests/Translator/TypesTranslator/Test_typeOverrides.swift
```diff
@@ -0,0 +1,88 @@
...
+                  id:
+                    $ref: '#/components/schemas/UUID'
+              UUID:
+                type: string
+                format: uuid
+            """#
+        )
+        let translator = makeTranslator(
+            components: components,
+            schemaOverrides: ["#/components/schemas/UUID": "Foundation.UUID"]
```
czechboy0, line 36: Outdated syntax?
```suggestion
            schemaOverrides: ["UUID": "Foundation.UUID"]
```

czechboy0 (commented):
Thanks @simonbility - I think the proposal is ready to be reviewed by the community. Could you open a separate PR with just the proposal, and we'll kick off the review process? Thanks for working on this! 🙏

Sources/swift-openapi-generator/GenerateOptions+runGenerator.swift
```diff
@@ -43,11 +44,19 @@ extension _GenerateOptions {
...
                 namingStrategy: resolvedNamingStragy,
                 nameOverrides: resolvedNameOverrides,
+                typeOverrides: resolvedTypeOverrides,
                 featureFlags: resolvedFeatureFlags
             )
         }
         let (diagnostics, finalizeDiagnostics) = preparedDiagnosticsCollector(outputPath: diagnosticsOutputPath)
         let doc = self.docPath
+        let typeOverridesDescription = """
+
```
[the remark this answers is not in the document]
simonbility: This one actually was on purpose.
The intent was to add a linebreak after `TypeOverrides` and then have all the overrides (only schemas at the moment) indented under it.

```
- TypeOverrides:<-this is the linebreak
  - Schemas: "Old"->"New" 
```

Without it would look like this
```
- TypeOverrides:  - Schemas: "Old"->"New" 
```

simonbility:
Addressed all the Feedback, and added a separate PR with only the Proposal here https://github.com/apple/swift-openapi-generator/pull/774

czechboy0:
Thanks @simonbility - the proposal was received well, so let's get this PR ready to land. Let me take another, more detailed review pass now.

Examples/replace-types-example/Sources/Types/Generated/Types.swift
czechboy0, line 1: Let's not check in the generated file, the example still drives home the point of how to type overrides.

Examples/replace-types-example/Sources/Types/openapi-generator-config.yaml
```diff
@@ -0,0 +1,10 @@
+generate:
+  - types
+accessModifier: package
+namingStrategy: idiomatic
+additionalImports:
+  - Foundation
+  - ExternalLibrary
+typeOverrides:
+  schemas:
+    UUID: Foundation.UUID
```
czechboy0, line 10: Is this example up-to-date? Seems this is showing replacing UUID, but above you're using `PrimeNumber`.

czechboy0 (changes requested):
Started a review but noticed that this PR might not be fully up-to-date with the proposal yet. When you have time, could you prepare this PR to be fully in line with the proposal, and we can do another review pass then? Thanks

Examples/replace-types-example/Sources/Types/Generated/Types.swift
simonbility: I switched to applying the plugin and removed the checked in sources.
Note the project currently fails to build because the plugin version of swift-open-api-generator does not yet support typeoverrides.

This will need to be updated after the feature becomes available

simonbility:
I think the PR should now reflect whats in the proposal.

FYI i did also added this https://github.com/apple/swift-openapi-generator/pull/764/commits/7956564205e6bdcbfcc0915f6eb1e4b8d29cf301
It emits a warning when a override is specified for a schema that doesnt exist.

Examples/type-overrides-example/Package.swift
```diff
@@ -0,0 +1,39 @@
+// swift-tools-version:5.10
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftOpenAPIGenerator open source project
+//
+// Copyright (c) 2024 Apple Inc. and the SwiftOpenAPIGenerator project authors
```
czechboy0, line 6:
```suggestion
// Copyright (c) 2025 Apple Inc. and the SwiftOpenAPIGenerator project authors
```

Examples/type-overrides-example/Package.swift
```diff
@@ -0,0 +1,39 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import PackageDescription
+
+let package = Package(
+    name: "type-overrides-example",
+    platforms: [.macOS(.v14)],
```
czechboy0, line 19:
```suggestion
    platforms: [.macOS(.v10_15)],
```

I don't think we're using any new features in this example beyond the minimum platform version?

Examples/type-overrides-example/README.md
```diff
@@ -0,0 +1,18 @@
+# Overriding types
+
+An example project using [Swift OpenAPI Generator](https://github.com/apple/swift-openapi-generator).
+
+> **Disclaimer:** This example is deliberately simplified and is intended for illustrative purposes only.
+
+## Overview
+
+This example shows how to use the TypeOverrides feature of the Swift OpenAPI Generator.
```
czechboy0, line 9:
```suggestion
This example shows how to use [type overrides](https://swiftpackageindex.com/apple/swift-openapi-generator/documentation/swift-openapi-generator/configuring-the-generator) with Swift OpenAPI Generator.
```

czechboy0: Please document this feature in https://swiftpackageindex.com/apple/swift-openapi-generator/1.8.0/documentation/swift-openapi-generator/configuring-the-generator, from there please also link the proposal which will land at https://swiftpackageindex.com/apple/swift-openapi-generator/documentation/swift-openapi-generator/soar-0014

Examples/type-overrides-example/Sources/ExternalLibrary/PrimeNumber.swift
```diff
@@ -0,0 +1,38 @@
...
+            throw DecodingError.dataCorruptedError(in: container, debugDescription: "The number is not prime.")
+        }
+        self = value
+    }
+    public func encode(to encoder: any Encoder) throws {
+        var container = encoder.singleValueContainer()
+        try container.encode(self.rawValue)
+    }
+
+}
```
czechboy0, line 24:
```suggestion
/// Example struct to be used instead of the default generated type.
/// This illustrates how to introduce a type performing additional validation during Decoding that cannot be expressed with OpenAPI
public struct PrimeNumber: Codable, Hashable, RawRepresentable, Sendable {
    public let rawValue: Int
    public init?(rawValue: Int) {
        if !rawValue.isPrime { return nil }
        self.rawValue = rawValue
    }

    public init(from decoder: any Decoder) throws {
        let container = try decoder.singleValueContainer()
        let number = try container.decode(Int.self)
        guard let value = PrimeNumber(rawValue: number) else {
            throw DecodingError.dataCorruptedError(in: container, debugDescription: "The number is not prime.")
        }
        self = value
    }

    public func encode(to encoder: any Encoder) throws {
        var container = encoder.singleValueContainer()
        try container.encode(self.rawValue)
    }
}
```

Tiny nits

Examples/type-overrides-example/Sources/openapi.yaml
```diff
@@ -0,0 +1,46 @@
...
+          description: The name of the user
+          schema:
+            type: string
+      responses:
+        '200':
+          description: A success response with the user.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/User'
```
czechboy0, line 25:
```suggestion
```

Let's drop all of `paths` since we're not generating client/server.

Examples/type-overrides-example/Sources/openapi.yaml
```diff
@@ -0,0 +1,46 @@
...
+      type: object
+      properties:
+        id:
+          $ref: '#/components/schemas/UUID'
+        name:
+          type: string
+        favorite_prime_number:
+          $ref: '#/components/schemas/PrimeNumber'
+        
+
```
czechboy0, line 46:
```suggestion

```

Sources/_OpenAPIGeneratorCore/Config.swift
```diff
@@ -59,6 +59,9 @@ public struct Config: Sendable {
 
     /// A map of OpenAPI identifiers to desired Swift identifiers, used instead of the naming strategy.
     public var nameOverrides: [String: String]
+    
+    /// A map of OpenAPI paths to desired Types
```
czechboy0, line 63:
```suggestion
    /// A map of OpenAPI schema names to desired custom type names.
```

Sources/_OpenAPIGeneratorCore/Config.swift
```diff
@@ -73,6 +76,7 @@ public struct Config: Sendable {
     ///     Defaults to `defensive`.
     ///   - nameOverrides: A map of OpenAPI identifiers to desired Swift identifiers, used instead
     ///     of the naming strategy.
+    ///   - typeOverrides: A map of OpenAPI paths to desired Types
```
czechboy0, line 79:
```suggestion
    ///   - typeOverrides: A map of OpenAPI schema names to desired custom type names.
```

Sources/_OpenAPIGeneratorCore/Config.swift
```diff
@@ -81,6 +85,7 @@ public struct Config: Sendable {
         filter: DocumentFilter? = nil,
         namingStrategy: NamingStrategy,
         nameOverrides: [String: String] = [:],
+        typeOverrides: TypeOverrides = TypeOverrides(),
```
czechboy0, line 88:
```suggestion
        typeOverrides: TypeOverrides = .init(),
```

Sources/_OpenAPIGeneratorCore/Translator/CommonTypes/Constants.swift
```diff
@@ -372,7 +372,6 @@ enum Constants {
         /// The substring used in method names for the multipart coding strategy.
         static let multipart: String = "Multipart"
     }
-
```
czechboy0, line 375: Please undo this change

Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateSchemas.swift
```diff
@@ -76,4 +77,18 @@ extension TypesFileTranslator {
         )
         return componentsSchemasEnum
     }
+    private func diagnoseTypeOverrideForNonExistentSchema() throws {
```
czechboy0, line 80: This is a great idea - please just move it to validateDoc.swift and add doc comments, and invoke it from `validateDoc`, rather than from `translateSchemas`. Thanks!

Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateSchemas.swift
```diff
@@ -76,4 +77,18 @@ extension TypesFileTranslator {
...
+        let nonExistentOverrides = config.typeOverrides.schemas.keys
+            .filter { key in
+                guard let componentKey = OpenAPI.ComponentKey(rawValue: key) else { return false }
+                return !self.components.schemas.contains(key: componentKey)
+            }
+            .sorted()
+        
+        for override in nonExistentOverrides {
+            try diagnostics.emit(
+                .warning(message: "TypeOverride defined for schema '\(override)' that is not defined in the OpenAPI document")
```
czechboy0, line 90:
```suggestion
                .warning(message: "A type override defined for schema '\(override)' is not defined in the OpenAPI document.")
```

Sources/_OpenAPIGeneratorCore/TypeOverrides.swift
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// Type containing overrides for schema types.
```
czechboy0, line 15:
```suggestion
/// A container of schema type overrides.
```

Sources/_OpenAPIGeneratorCore/TypeOverrides.swift
```diff
@@ -0,0 +1,26 @@
...
+public struct TypeOverrides: Sendable {
+    /// A dictionary of overrides for replacing the types generated from schemas with manually provided types
```
czechboy0, line 17:
```suggestion
    /// A dictionary of overrides for replacing named schemas from the OpenAPI document with custom types.
```

Sources/_OpenAPIGeneratorCore/TypeOverrides.swift
```diff
@@ -0,0 +1,26 @@
...
+    public var schemas: [String: String]
+
+    /// Creates a new instance of `TypeOverrides`
```
czechboy0, line 20:
```suggestion
    /// Creates a new instance.
```

Sources/_OpenAPIGeneratorCore/TypeOverrides.swift
```diff
@@ -0,0 +1,26 @@
...
+    /// - Parameter schemas: A dictionary mapping schema names to their override types.
```
czechboy0, line 21:
```suggestion
    /// - Parameter schemas: A dictionary of overrides for replacing named schemas from the OpenAPI document with custom types.
```

Sources/swift-openapi-generator/GenerateOptions+runGenerator.swift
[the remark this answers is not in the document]
czechboy0: Let's make it even shorter, `- Type overrides: "Foo" -> "Bar", "Baz" -> "UUID"`, all one line.

Sources/swift-openapi-generator/GenerateOptions.swift
```diff
@@ -93,6 +93,13 @@ extension _GenerateOptions {
     /// - Parameter config: The configuration specified by the user.
     /// - Returns: The name overrides requested by the user
     func resolvedNameOverrides(_ config: _UserConfig?) -> [String: String] { config?.nameOverrides ?? [:] }
+    /// Returns the type overrides requested by the user.
+    /// - Parameter config: The configuration specified by the user.
+    /// - Returns: The type overrides requested by the user
+    func resolvedTypeOverrides(_ config: _UserConfig?) -> TypeOverrides {
+        if let typeOverrides = config?.typeOverrides { return TypeOverrides(schemas: typeOverrides.schemas ?? [:]) }
+        return TypeOverrides()
+    }
```
czechboy0, line 102:
```suggestion
    func resolvedTypeOverrides(_ config: _UserConfig?) -> TypeOverrides {
        guard let schemaOverrides = config?.typeOverrides?.schemas, !schemaOverrides.isEmpty else {
            return .init()
        }
        return TypeOverrides(schemas: schemaOverrides)
    }
```

Sources/_OpenAPIGeneratorCore/TypeOverrides.swift
```diff
@@ -0,0 +1,26 @@
...
+    public init(schemas: [String: String] = [:]) { self.schemas = schemas }
+
+    /// A Boolean value indicating whether there are no overrides.
+    public var isEmpty: Bool { schemas.isEmpty }
```
czechboy0, line 25:
```suggestion
```

I don't think we need this?

Sources/swift-openapi-generator/UserConfig.swift
```diff
@@ -54,6 +57,13 @@ struct _UserConfig: Codable {
         case filter
         case namingStrategy
         case nameOverrides
+        case typeOverrides
         case featureFlags
     }
+    
+    struct TypeOverrides: Codable {
+        
+        /// A dictionary of overrides for replacing the types generated from schemas with manually provided types
```
czechboy0, line 66:
```suggestion
        /// A dictionary of overrides for replacing the types generated from schemas with manually provided types.
```

Sources/swift-openapi-generator/UserConfig.swift
```diff
@@ -54,6 +57,13 @@ struct _UserConfig: Codable {
         case filter
         case namingStrategy
         case nameOverrides
+        case typeOverrides
         case featureFlags
     }
+    
+    struct TypeOverrides: Codable {
```
czechboy0, line 64:
```suggestion

    /// A container of type overrides.
    struct TypeOverrides: Codable {
```

Tests/OpenAPIGeneratorCoreTests/TestUtilities.swift
```diff
@@ -62,20 +66,25 @@ class Test_Core: XCTestCase {
...
             nameOverrides: nameOverrides,
+            typeOverrides: TypeOverrides(schemas: schemaOverrides),
             featureFlags: featureFlags
         )
     }
 
     func loadSchemaFromYAML(_ yamlString: String) throws -> JSONSchema {
         try YAMLDecoder().decode(JSONSchema.self, from: yamlString)
     }
+    func loadComponentsFromYAML(_ yamlString: String) throws -> OpenAPI.Components {
```
czechboy0, line 85:
```suggestion
    }

    func loadComponentsFromYAML(_ yamlString: String) throws -> OpenAPI.Components {
```

Tests/OpenAPIGeneratorCoreTests/DeclarationHelpers.swift
```diff
@@ -0,0 +1,66 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+@testable import _OpenAPIGeneratorCore
+
+extension Declaration {
```
czechboy0, line 16: I don't think we need these, let's move the testing strategy to snippet-based reference tests here, which will be needed anyway once you move the validation to validateDoc.

Tests/OpenAPIGeneratorCoreTests/Translator/TypesTranslator/Test_typeOverrides.swift
```diff
@@ -0,0 +1,88 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import XCTest
+import OpenAPIKit
+import Yams
+@testable import _OpenAPIGeneratorCore
+
+class Test_typeOverrides: Test_Core {
```
czechboy0, line 19: Let's use SnippetBasedReferenceTests for this please, it's much more maintainable for us.

czechboy0 (changes requested):
Looks great overall, added a bunch of small asks, thank you for your patience 🙂 We're almost there.

Tests/OpenAPIGeneratorCoreTests/Translator/TypesTranslator/Test_typeOverrides.swift
simonbility: Oh nice!
I was looking for something like this but did find it.

Sources/_OpenAPIGeneratorCore/Translator/CommonTypes/Constants.swift
simonbility: FWIW i just ran the code-formatter on this file.

simonbility:
Thanks for the thorough review 🙏 

I appreciate the suggestions of the documentation writing.
I always have a hard time writing these.

If there is anything else i can do please let me know.

Examples/type-overrides-example/Sources/openapi.yaml
```diff
@@ -0,0 +1,26 @@
...
+servers:
+  - url: https://example.com/api
+    description: Example service deployment.
+components:
+  schemas:
+    UUID: # this will be replaced by with Foundation.UUID specified by typeOverrides in open-api-generator-config
+      type: string
+      format: uuid
+    
+    PrimeNumber: # this will be replaced by with ExternalLibrary.PrimeNumber specified by typeOverrides in open-api-generator-config
```
czechboy0, line 14:
```suggestion
    UUID: # this will be replaced by with Foundation.UUID specified by typeOverrides in openapi-generator-config
      type: string
      format: uuid
    
    PrimeNumber: # this will be replaced by with ExternalLibrary.PrimeNumber specified by typeOverrides in openapi-generator-config
```

Sources/_OpenAPIGeneratorCore/Parser/validateDoc.swift
```diff
@@ -97,8 +97,7 @@ func validateContentTypes(in doc: ParsedOpenAPIRepresentation, validate: (String
 
 /// Validates all references from an OpenAPI document represented by a ParsedOpenAPIRepresentation against its components.
 ///
-/// This method traverses the OpenAPI document to ensure that all references
-/// within the document are valid and point to existing components.
+/// This method iterates through all Type overrides in a config and checks if a schem
```
czechboy0, line 100: Can you undo this change? I think it's unrelated, right?

simonbility: Oh yes that was supposed to be placed on the `validateTypeOverrides` method

czechboy0:
Only comment changes requested, kicking CI.

czechboy0:
There are a few failures, please take a look when you have time - might be worth also running them all locally to verify after the fixes, more in: https://github.com/apple/swift-openapi-generator/blob/main/CONTRIBUTING.md#run-ci-checks-locally

simonbility:
@czechboy0 
I think i addressed what i can. (Added Missing License Header, and ran the code-formatter)
The failure when building with Xcode 26 i think might be a Xcode bug. https://github.com/apple/swift-openapi-generator/actions/runs/15705561051/job/44250487271

And as previously mentioned the failures in the example project are expected. https://github.com/apple/swift-openapi-generator/actions/runs/15705561051/job/44250491634

czechboy0:
Actually, it's as green as it can be right now. Merging - thanks @simonbility! 👏

czechboy0:
Released in https://github.com/apple/swift-openapi-generator/releases/tag/1.9.0, thanks @simonbility 👏

simonbility:
🎉 
Feels awesome being able to fix my own problem.
Thanks for all the support 🙏

Issue #756, opened 2025-04-02
Multipart form properties codegen is not typesafe


jpsim:
### Motivation

One of the main benefits I get from using swift-openapi-generator is type safety for client code.

However, for multipart form upload requests, it seems like the property codegen is not typesafe, where all properties just get a generic `HTTPBody` payload.

For example, given this OpenAPI document:

```yaml
/uploadFile:
  post:
    operationId: uploadFile
    requestBody:
      required: true
      content:
        multipart/form-data:
          schema:
            type: object
            properties:
              description:
                type: string
                description: A text description of the uploaded file
              count:
                type: integer
                description: An integer value associated with the upload
              imageFile:
                type: string
                format: binary
                description: The image file to upload
            required:
              - description
              - count
              - imageFile
```

This code is generated for the request body type:

```swift
enum Body: Sendable, Hashable {
    enum MultipartFormPayload: Sendable, Hashable {
        struct DescriptionPayload: Sendable, Hashable {
            var body: OpenAPIRuntime.HTTPBody
            init(body: OpenAPIRuntime.HTTPBody) {
                self.body = body
            }
        }
        case description(OpenAPIRuntime.MultipartPart<Operations.UploadFile.Input.Body.MultipartFormPayload.DescriptionPayload>)
        struct CountPayload: Sendable, Hashable {
            var body: OpenAPIRuntime.HTTPBody
            init(body: OpenAPIRuntime.HTTPBody) {
                self.body = body
            }
        }
        case count(OpenAPIRuntime.MultipartPart<Operations.UploadFile.Input.Body.MultipartFormPayload.CountPayload>)
        struct ImageFilePayload: Sendable, Hashable {
            var body: OpenAPIRuntime.HTTPBody
            init(body: OpenAPIRuntime.HTTPBody) {
                self.body = body
            }
        }
        case imageFile(OpenAPIRuntime.MultipartPart<Operations.UploadFile.Input.Body.MultipartFormPayload.ImageFilePayload>)
        case undocumented(OpenAPIRuntime.MultipartRawPart)
    }
    case multipartForm(OpenAPIRuntime.MultipartBody<Operations.UploadFile.Input.Body.MultipartFormPayload>)
}
```

So at the call site, you're responsible for producing `HTTPBody` values that respect the schema, but you get no compile-type safety if you make a mistake:

```swift
let multipartBody: MultipartBody<Operations.UploadFile.Input.Body.MultipartFormPayload> = [
[...]

PR #775, opened 2025-05-27
Add convenience initializers for primitive-typed multipart parts

jpsim:
Generate additional `init(value:)` initializers for multipart payload structs when the part is defined with primitive types (integer, number, string, or boolean) in the OpenAPI document. This provides a more ergonomic, typesafe API by allowing direct value initialization instead of requiring manual `HTTPBody` construction with no type validation.

The generated initializers automatically convert numeric and boolean values to strings as required by multipart form encoding.

Before/After Examples:

| Type    | Before                          | After                  |
|---------|---------------------------------|------------------------|
| integer | `init(body: HTTPBody("123"))`   | `init(value: 123)`     |
| number  | `init(body: HTTPBody("45.67"))` | `init(value: 45.67)`   |
| string  | `init(body: HTTPBody("hello"))` | `init(value: "hello")` |
| boolean | `init(body: HTTPBody("true"))`  | `init(value: true)`    |

The new initializers:
- For integers: `init(value: Int)` - converts to string internally
- For numbers: `init(value: Double)` - converts to string internally
- For strings: `init(value: String)` - passes through directly
- For booleans: `init(value: Bool)` - converts to string internally

These convenience initializers are also generated for multipart parts with custom headers. When headers are present, the generated initializer includes both headers and value parameters:
- Without headers: `init(value: String)`
- With headers: `init(headers: Headers, value: String)`

This partly resolves https://github.com/apple/swift-openapi-generator/issues/756. Support for non-trivial types (objects, arrays, etc.) is not included in this change as it would require more complex type handling that I couldn't figure out.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateStructBlueprint.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateStructBlueprint.swift
@@ -163,10 +163,10 @@ extension FileTranslator {
             // are not handled by this specific "value" initializer.
             guard valueTypeName == .string else { return nil }
             needsStringConversion = false
-        default:
-            // Other schema types (e.g., .object, .array, .fragment, .reference)
-            // are not supported for this value initializer, even if they might
-            // be considered "built-in" by the TypeMatcher for other purposes.
+        case .object, .array, .all, .one, .any, .not, .reference, .fragment, .null:
+            // Other schema types are not supported for this value initializer,
+            // even if they might be considered "built-in" by the TypeMatcher
+            // for other purposes.
             return nil
         }
 


czechboy0:
> Ah TIL, this is really unfortunate as it appears to mean that https://github.com/apple/swift-openapi-generator/issues/756 is fundamentally impossible! I did not realize that from our discussion in that issue.

Yeah it also only dawned on me that we don't have a spec/RFC to follow here.

> What if instead, we made this clear in the initializer that it assumes JSON scalar serialization by renaming the parameter from value to jsonValue?

Serializing it as JSON would e.g. mean quoting strings, aka converting them to JSON fragments using JSONEncoder.

Although I'm worried about creating a side channel of assuming both sides are Swift. We've historically taken the opposite direction, and tried to stick to the spec as much as possible. I'll let @simonjbeaumont chime in as well.

jpsim:
I just realized there's a cleaner solution here that both enforces type safety for multipart parts without enforcing json serialization or any serialization strategy at all really. Just let the user pass the serialization function.

```swift
init(value: Swift.Int, toHTTPBody: (Swift.Int) -> OpenAPIRuntime.HTTPBody) {
    self.body = toHTTPBody(value))
}
```

czechboy0:
Which type would this be an initializer on?

jpsim:
Here's a concrete example:

```swift
public enum RequestBodies {
    @frozen public enum MultipartIntegerRequest: Sendable, Hashable {
        @frozen public enum multipartFormPayload: Sendable, Hashable {
            public struct countPayload: Sendable, Hashable {
                public var body: OpenAPIRuntime.HTTPBody
                public init(body: OpenAPIRuntime.HTTPBody) {
                    self.body = body
                }
                public init(value: Swift.Int, toHTTPBody: (Swift.Int) -> OpenAPIRuntime.HTTPBody
                    self.body = toHTTPBody(value))
                }
            }
            case count(OpenAPIRuntime.MultipartPart<Components.RequestBodies.MultipartIntegerRequest.multipartFormPayload.countPayload>)
            public struct scorePayload: Sendable, Hashable {
                public var body: OpenAPIRuntime.HTTPBody
                public init(body: OpenAPIRuntime.HTTPBody) {
                    self.body = body
                }
                public init(value: Swift.Int32, toHTTPBody: (Swift.Int32) -> OpenAPIRuntime.HTTPBody
                    self.body = toHTTPBody(value))
                }
            }
            case score(OpenAPIRuntime.MultipartPart<Components.RequestBodies.MultipartIntegerRequest.multipartFormPayload.scorePayload>)
            public struct namePayload: Sendable, Hashable {
                public var body: OpenAPIRuntime.HTTPBody
                public init(body: OpenAPIRuntime.HTTPBody) {
                    self.body = body
                }
                public init(value: Swift.String, toHTTPBody: (Swift.String) -> OpenAPIRuntime.HTTPBody
                    self.body = toHTTPBody(value))
                }
            }
            case name(OpenAPIRuntime.MultipartPart<Components.RequestBodies.MultipartIntegerRequest.multipartFormPayload.namePayload>)
            public struct itemsPayload: Sendable, Hashable {
                public var body: OpenAPIRuntime.HTTPBody
                public init(body: OpenAPIRuntime.HTTPBody) {
                    self.body = body
                }
            }
            case items(OpenAPIRuntime.MultipartPart<Components.RequestBodies.MultipartIntegerRequest.multipartFormPayload.itemsPayload>)
            case undocumented(OpenAPIRuntime.MultipartRawPart)
        }
        case multipartForm(OpenAPIRuntime.MultipartBody<Components.RequestBodies.MultipartIntegerRequest.multipartFormPayload>)
    }
}
```

PR #789, opened 2025-06-19
Fix the test overrides example package

czechboy0:
### Motivation

The CI was failing on the new example.

### Modifications

Fixed the structure and cleaned up the test project a bit.

### Result

CI should start passing now, and the test project is more focused on type overrides.

### Test Plan

Added a unit test target to the test project.


simonjbeaumont (approved):
Thank you!

PR #795, opened 2025-07-16
Fix final layer of error handling in command plugin

### LINKED ISSUE #794: Invalid `openapi.yaml` produces unexpected diagnostic message
### Description

When a typo is made or a wrong property name is used in `openapi.yaml` schema declaration, no relevant diagnostic message is provided when running `swift package generate-code-from-openapi`. The only error message seen is this one:
```
Targets with names FooTarget don't contain any config or OpenAPI document files with expected names.
```
This message is incorrect, since both config and OpenAPI document are provided in the expected location. The actual underlying reason is that the OpenAPI document didn't pass validation, but validation diagnostic is not produced, which is very confusing.

### Reproduction

Create an OpenAPI document that uses `int` and `double` types in the schema instead of `integer` and `number`:
```yaml
openapi: '3.1.0'
info:
  title: GreetingService
  version: 1.0.0
servers:
  - url: https://example.com/api
    description: Example service deployment.
paths:
  /summary:
    get:
      operationId: getSummary
      parameters:
        - name: name
          required: false
          in: query
          description: The name used in the returned greeting.
          schema:
            type: string
      responses:
        '200':
          description: A success response with a greeting.
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Greeting'
components:
  schemas:
    Greeting:
      type: object
      properties:
        message:
          type: string
        floatingPoint:
          type: double
        integerValue:
          type: int
```

### Package version(s)

```json
      "location" : "https://github.com/apple/swift-openapi-generator.git",
      "state" : {
        "revision" : "ea291f05436363665ee67c453f4a72a895f18e75",
        "version" : "1.9.0"
      }
```

### Expected behavior

If `openapi.yaml` doesn't pass validation, a validation error should be produced, ideally with line/column information and an actionable suggestion for how the error can be fixed.

### Environment

```
> swift --version
swift-driver version: 1.120.5 Apple Swift version 6.1.2 (swiftlang-6.1.2.1.2 clang-1700.0.13.5)
Target: arm64-apple-macosx15.0
```

### Additional information

_No response_

simonjbeaumont:
### Motivation

When the generator is run as a command plugin on a correctly configured target but there are problems generating, there's a layer of error handling which is incorrect resulting in a poor user experience. For example, if the OpenAPI document contains a bad type like this...

```yaml
components:
  schemas:
    BadIntHolder:
      type: object
      description: A value that has a bad typed int.
      properties:
        int:
          type: int
          description: A bad type
```

...then the command plugin shows the following output:

```
Considering target 'HelloWorldURLSessionClient':
- Trying OpenAPI code generation.
/Users/Si/work/code/swift-openapi-workspace/packages/swift-openapi-generator/Examples/hello-world-urlsession-client-example/Sources/HelloWorldURLSessionClient/openapi.yaml: error: Found neither a JSONType nor a Arr
ay<JSONType> in Document.components.schemas.Greeting.properties.i.type.

JSONType could not be decoded because:
Inconsistency encountered when parsing `[unknown object]`: Cannot initialize JSONType from invalid String value int.

Array<JSONType> could not be decoded because:
Expected value to be parsable as Sequence but it was not..
Error: /Users/Si/work/code/swift-openapi-workspace/packages/swift-openapi-generator/Examples/hello-world-urlsession-client-example/Sources/HelloWorldURLSessionClient/openapi.yaml: error: Found neither a JSONType no
r a Array<JSONType> in Document.components.schemas.Greeting.properties.i.type.

JSONType could not be decoded because:
Inconsistency encountered when parsing `[unknown object]`: Cannot initialize JSONType from invalid String value int.

Array<JSONType> could not be decoded because:
Expected value to be parsable as Sequence but it was not..
- OpenAPI document path: /Users/Si/work/code/swift-openapi-workspace/packages/swift-openapi-generator/Examples/hello-world-urlsession-client-example/Sources/HelloWorldURLSessionClient/openapi.yaml
- Configuration path: /Users/Si/work/code/swift-openapi-workspace/packages/swift-openapi-generator/Examples/hello-world-urlsession-client-example/Sources/HelloWorldURLSessionClient/openapi-generator-config.yaml
- Access modifier: internal
- Naming strategy: idiomatic
- Name overrides: <none>
- Type overrides: <none>
- Feature flags: <none>
- Output directory: /Users/Si/work/code/swift-openapi-workspace/packages/swift-openapi-generator/Examples/hello-world-urlsession-client-example/Sources/HelloWorldURLSessionClient/GeneratedSources
- Current directory: /Users/Si/work/code/swift-openapi-workspace/packages/swift-openapi-generator/Examples/hello-world-urlsession-client-example
- Plugin source: command
- Is dry run: false
- Additional imports: <none>
- Additional file comments: <none>
- Stopping because target isn't configured for OpenAPI code generation.
error: Targets with names HelloWorldURLSessionClient don't contain any config or OpenAPI document files with expected names. See documentation for details.
```

There are two issues with this:

1. The final error is just plain wrong.
2. Because the diagnostics are printed to stderr and the big info message to stdout they are ordered wrongly—you get the diagnostics _before_ the message saying how the generator will run.

### Modifications

- Fix the categorisation of errors in the command plugin.
- Print the big info message to stderr.

### Result

- Command plugin now correctly categorises misconfiguration errors vs other errors and provides more appropriate output on failure.

The above example now results in:

```
Considering target 'HelloWorldURLSessionClient':
- Trying OpenAPI code generation.
- OpenAPI document path: /Users/Si/work/code/swift-openapi-workspace/packages/swift-openapi-generator/Examples/hello-world-urlsession-client-example/Sources/HelloWorldURLSessionClient/openapi.yaml
- Configuration path: /Users/Si/work/code/swift-openapi-workspace/packages/swift-openapi-generator/Examples/hello-world-urlsession-client-example/Sources/HelloWorldURLSessionClient/openapi-generator-config.yaml
- Access modifier: internal
- Naming strategy: idiomatic
- Name overrides: <none>
- Type overrides: <none>
- Feature flags: <none>
- Output directory: /Users/Si/work/code/swift-openapi-workspace/packages/swift-openapi-generator/Examples/hello-world-urlsession-client-example/Sources/HelloWorldURLSessionClient/GeneratedSources
- Current directory: /Users/Si/work/code/swift-openapi-workspace/packages/swift-openapi-generator/Examples/hello-world-urlsession-client-example
- Plugin source: command
- Is dry run: false
- Additional imports: <none>
- Additional file comments: <none>/Users/Si/work/code/swift-openapi-workspace/packages/swift-openapi-generator/Examples/hello-world-urlsession-client-example/Sources/HelloWorldURLSessionClient/openapi.yaml: error: Found neither a JSONType nor a Array<JSONType> in Document.components.schemas.Greeting.properties.int.type.

JSONType could not be decoded because:
Inconsistency encountered when parsing `[unknown object]`: Cannot initialize JSONType from invalid String value int.

Array<JSONType> could not be decoded because:
Expected value to be parsable as Sequence but it was not..
Error: /Users/Si/work/code/swift-openapi-workspace/packages/swift-openapi-generator/Examples/hello-world-urlsession-client-example/Sources/HelloWorldURLSessionClient/openapi.yaml: error: Found neither a JSONType nor a Array<JSONType> in Document.components.schemas.Greeting.properties.int.type.

JSONType could not be decoded because:
Inconsistency encountered when parsing `[unknown object]`: Cannot initialize JSONType from invalid String value int.

Array<JSONType> could not be decoded because:
Expected value to be parsable as Sequence but it was not..
- OpenAPI code generation failed with error.
error: The generator failed to generate OpenAPI files for target 'HelloWorldURLSessionClient'.
```

## Related issues

- Fixes #794

### DIFF
--- a/Plugins/OpenAPIGeneratorCommand/plugin.swift
+++ b/Plugins/OpenAPIGeneratorCommand/plugin.swift
@@ -82,10 +82,11 @@ extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
                 hadASuccessfulRun = true
             } catch let error as PluginError {
                 if error.isMisconfigurationError {
-                    print("- OpenAPI code generation failed with error.")
+                    print("- Stopping because target isn't configured for OpenAPI code generation.")
                     throw error
                 } else {
-                    print("- Stopping because target isn't configured for OpenAPI code generation.")
+                    print("- OpenAPI code generation failed with error.")
+                    throw error
                 }
             }
         }
--- a/Sources/swift-openapi-generator/GenerateOptions+runGenerator.swift
+++ b/Sources/swift-openapi-generator/GenerateOptions+runGenerator.swift
@@ -52,7 +52,7 @@ extension _GenerateOptions {
         }
         let (diagnostics, finalizeDiagnostics) = preparedDiagnosticsCollector(outputPath: diagnosticsOutputPath)
         let doc = self.docPath
-        print(
+        FileHandle.standardError.write(
             """
             Swift OpenAPI Generator is running with the following configuration:
             - OpenAPI document path: \(doc.path)


simonjbeaumont:
//cc @MaxDesiatov

czechboy0 (approved):
Thanks!

MaxDesiatov (commented):
Does this actually output the validation error? I don't see that changed in the diff, but I'm not familiar with the code to say for sure

simonjbeaumont:
@MaxDesiatov the validation output was always printed (it should have been in your output already, just way up because of the stderr/stdout flushing differences). You can see in the examples in the PR description that they were printed in both cases too. Now they'll be printed in the correct order and you'll see them right before the final error.

Would you mind running your repro against this branch to confirm?

MaxDesiatov:
I see, yes it's at the top, which I personally find unobvious. May I suggest that separating the error messages to be placed on adjacent lines? `- OpenAPI code generation failed with error` with actual error context dozens of lines above feels disconnected. If placing them on adjacent lines requires significant changes, maybe it could say "see the detailed error message above" or smth like that?

simonjbeaumont:
@MaxDesiatov I'm just wanting to confirm things here... [your comment on this PR](https://github.com/apple/swift-openapi-generator/pull/795#issuecomment-3077673544) says:

> I see, yes it's at the top, which I personally find unobvious. May I suggest that separating the error messages to be placed on adjacent lines? - OpenAPI code generation failed with error with actual error context dozens of lines above feels disconnected.

To be clear, we have solved exactly that in this PR. The weird ordering was the use of both stdout and stderr. It should now be the case that the final output of the tool in this failure mode is an error that it couldn't generate and, right above that, will be the errors.

But... in [your comment on the issue](https://github.com/apple/swift-openapi-generator/issues/794#issuecomment-3077467672) you say

> No I never saw this in the output. That was the most misleading thing in the current diagnostic behavior. For some time I thought the generator couldn't discover the OpenAPI document at all.

I'm pretty sure that this PR addresses your issue but I'd like to make sure I haven't missed something before we cut a release 🙏

MaxDesiatov:
Sorry, the error message was there, but it was cut off because my terminal height was set to just a few lines. I didn't notice it because of the ordering. If changing from stdout to stderr fixes the ordering issue, that's much appreciated, thanks!

Plugins/OpenAPIGeneratorCommand/plugin.swift
```diff
@@ -82,10 +82,11 @@ extension SwiftOpenAPIGeneratorPlugin: CommandPlugin {
                 hadASuccessfulRun = true
             } catch let error as PluginError {
                 if error.isMisconfigurationError {
-                    print("- OpenAPI code generation failed with error.")
+                    print("- Stopping because target isn't configured for OpenAPI code generation.")
```
MahdiBM, line 85: This is not quite right though.

When `isMisconfigurationError` is `true`, It's pretty much certain that the user configured their target for openapi, but did it wrong.

When `isMisconfigurationError` is `false`, it can be either a misconfiguration or no-configuration.

Or at least that's what i intended with this code.

simonjbeaumont: That's fair, how about we just tweak this one?


```suggestion
if error.isMisconfigurationError {
                    print("- Stopping because target is misconfigured for OpenAPI code generation.")
```

//cc @MahdiBM @czechboy0

PR #858, opened 2026-01-12
Fix document filtering to handle references with description or summary fields

egrank:
### Motivation

If OpenAPI document filtering is enabled for operations containing references with `description` or `summary` fields, Swift OpenAPI Generator fails with an error as such references cannot be resolved.

For example, this reference can be resolved:

```Yaml
  schema:
    $ref: '#/components/schemas/Item'
```

but not this one: 
```Yaml
  schema:
    $ref: '#/components/schemas/Item'
    description: "Item with description"
```

The cause of the described issue is in the `FilteredDocumentBuilder` code. The required components are registered there, including optional `description` and `summary` properties, while the reference matching for filtering is done against references created only with component keys, so equality check fails if `description` or `summary` are present.

### Modifications

Registering and matching only component keys instead of whole references.

### Result

After this change, there are no more limitations on having description and summary fields for references in filtered operations.

### Test Plan

Extended unit tests with test cases to cover the described issues:

* Added `/things/c` path that exercises references with descriptions across:
    * Path-level and operation-level parameters with description fields
    * Response references with descriptions
    * Headers with schema references containing descriptions
    * Examples with descriptions
* Updated the test assertion helper to verify headers and examples are correctly filtered

The added tests fail without the fix.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Hooks/FilteredDocument.swift
+++ b/Sources/_OpenAPIGeneratorCore/Hooks/FilteredDocument.swift
@@ -328,7 +328,8 @@ private extension FilteredDocumentBuilder {
         switch schema.value {
 
         case .reference(let reference, _):
-            guard requiredSchemaReferences.insert(try OpenAPI.ComponentKey(stringLiteral: reference.requiredName)).inserted else { return }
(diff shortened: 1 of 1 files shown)


czechboy0 (approved):
This is great, thank you @egrank! 🙏

czechboy0 (approved):

Issue #859, opened 2026-01-18
Generate better server code for response content `*/*` (dynamic Content-Type)


brandonbloom:
## Description

When an OpenAPI response uses `content: {'*/*': ...}` to indicate the response media type is intentionally unconstrained / dynamic / extensible, the generated **server serializer** currently emits code that:

1. calls `converter.validateAcceptIfPresent("*/*", in: request.headerFields)`, and
2. sets the outgoing `Content-Type` to `"*/*"` (e.g. via `setResponseBodyAsBinary(..., contentType: "*/*")`).

This causes two practical problems:

- Client compatibility: it effectively forces clients to include `Accept: */*` just to pass validation (a client sending `Accept: application/json` can be rejected even though the operation is intentionally “any content”).
- Bad response header: it emits `Content-Type: */*`, which is not a useful concrete media type to put on the wire.

We’d like wildcard `*/*` to mean “no static content negotiation; content type is chosen dynamically at runtime”, not “alias for `application/octet-stream`”.

Commit hash: `ecf21fdf08d4cf30ca506bb822f5fe580e090efb`

## Reproduction

Minimal spec:

```yaml
# openapi.yaml
openapi: 3.0.3
info:
  title: t
  version: 1.0.0
paths:
  /download:
    get:
      operationId: download
      responses:
        "200":
          description: ok
          content:
            "*/*":
              schema:
                type: string
                format: binary
```

Minimal generator config:

```yaml
# openapi-generator-config.yaml
mode:
  - types
  - server
```

In the generated server serializer, the `*/*` response path includes `validateAcceptIfPresent("*/*")` and emits `Content-Type: */*`.

## Package version(s)

- swift-openapi-generator: commit `ecf21fdf08d4cf30ca506bb822f5fe580e090efb`
- (also involves generated code’s usage of swift-openapi-runtime `Converter.validateAcceptIfPresent` / `setResponseBodyAsBinary`)

## Expected behavior

For response content type exactly `*/*` (server serializers):

1. Skip `validateAcceptIfPresent` entirely.
2. Do not emit `Content-Type: */*`.

However, simply omitting `Content-Type` is not ideal; what we really want is *dynamic* `Content-Type`.

Feature request: plumb dynamic headers/media type through the generated output model for the `*/*` case, so the handler can provide a real `Content-Type` at runtime.

One concrete approach would be to model the `*/*` body case as something that carries header fields too, for example:

[...]

PR #860, opened 2026-01-18
Server: special-case */* responses (skip Accept validation, don’t emit Content-Type: */*)

brandonbloom:
### Motivation

When an OpenAPI response uses `content: {'*/*': ...}` to indicate an intentionally unconstrained/dynamic media type, the generated server serializer currently:

- calls `converter.validateAcceptIfPresent("*/*", in: request.headerFields)`, which effectively forces clients to include `Accept: */*`, and
- emits `Content-Type: */*` (via `setResponseBodyAsBinary(..., contentType: "*/*")`).

Both behaviors are problematic in practice.

This PR is a safe first step: it fixes the immediate incorrect strictness and avoids emitting an invalid/meaningless response `Content-Type`, while keeping the generated public API unchanged.

Related: #859 (kept open; follow-up can explore a richer, potentially source-breaking solution to plumb dynamic `Content-Type`).

### Modifications

- In server response serialization, special-case response content type exactly `*/*` to:
  - skip `validateAcceptIfPresent`, and
  - assign the `HTTPBody` directly without setting `Content-Type: */*`.
- Add a snippet-based regression test verifying the generated server code for a `*/*` response does not contain `validateAcceptIfPresent("*/*")` and does not contain `contentType: "*/*"`.

### Result

Generated servers no longer reject requests that omit `Accept: */*` for wildcard responses, and no longer emit `Content-Type: */*`.

### Test Plan

- `swift test`

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponseOutcome.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Responses/translateResponseOutcome.swift
@@ -402,54 +402,46 @@ extension ServerFileTranslator {
                 var caseCodeBlocks: [CodeBlock] = []
 
                 let contentType = typedContent.content.contentType
-                let isWildcardAnyContentType = contentType.lowercasedTypeAndSubtype == "*/*"
+                let contentTypeForServer = contentType.lowercasedTypeAndSubtype == "*/*"
+                    ? ContentType.applicationOctetStream
+                    : contentType
 
-                if !isWildcardAnyContentType {
-                    let contentTypeHeaderValue = contentType.headerValueForValidation
-                    let validateAcceptHeader: Expression = .try(
-                        .identifierPattern("converter").dot("validateAcceptIfPresent")
-                            .call([
-                                .init(label: nil, expression: .literal(contentTypeHeaderValue)),
-                                .init(label: "in", expression: .identifierPattern("request").dot("headerFields")),
-                            ])
-                    )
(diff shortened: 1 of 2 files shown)


czechboy0:
Hi @brandonbloom,

> Generated servers no longer reject requests that omit Accept: */*

It's surprising that it'd reject it, the idea is that if the client doesn't send an `Accept` header at all, the validation is skipped: https://github.com/apple/swift-openapi-runtime/blob/6e88f56e6cc686b0e27b287d5d01789a08acf0d0/Sources/OpenAPIRuntime/Conversion/Converter%2BServer.swift#L51

Are you seeing a different behavior? Is the server rejecting requests that completely omit `Accept`?

Regarding the response `Content-Type` header, I still think we should send one. Maybe sending `application/octet-stream` is more appropriate, considering we're just piping through raw bytes without knowing what they are? I agree that sending `Content-Type: */*` isn't right.

brandonbloom:
You’re right that if `Accept` is absent, validation is skipped. My wording was imprecise.

The failing case I saw was when `Accept` is present and concrete (for example `application/json`), while generated server code for a `*/*` response validated against `*/*`. That can reject concrete `Accept` values.

I updated generation so server-side `*/*` responses are treated as `application/octet-stream` for validation/serialization:
- validate `Accept` against `application/octet-stream`
- emit `Content-Type: application/octet-stream` (never `*/*`)

I also updated the snippet reference test to cover this case.

Issue #253, opened 2023-09-08
Consider spliting Components into separated files


bartoszirzyk:
Hi,
I've got an openapi.yaml file which has 64k lines. Generator produces Types.swift with 135k lines. It's a big file, hard to navigate and scroll through.  It'd be nice to have each Schemas element in separated file and maybe other generated types. Probably it's not what we want as a default behaviour but having a feature flag to enable such thing would be appreciated.

czechboy0:
Hi @bartoszirzyk - the current behavior is mainly driven by the fact that the recommended way to use the generator is through the build plugin, and the build plugin requires the name of the output files to be known before the OpenAPI document is parsed (due to how SwiftPM plugins work). So unfortunately I don't really see this happening until SwiftPM relaxes that requirement.

Out of curiosity, when do you find you need to look at the generated code, as opposed to looking at the source OpenAPI document and then using autocompletion? The generated code is considered an implementation detail that adopters should _have to_ look at much.

simonjbeaumont:
FWIW I think this is a reasonable ask. Experimentally, it seems that the compiler is having a much harder time with one large file (cf. more, smaller files). An example on my machine shows pretty substantial improvement:

```console
% wc -l Sources/GithubTypesOneFile/*
457133 Sources/GithubTypesOneFile/Types.swift

% swift package clean && swift package update && swift build --skip-update --target GithubTypesOneFile
Build complete! (647.71s)
```

```console
% wc -l Sources/GithubTypesMajorSplit/*
 10492 Sources/GithubTypesMajorSplit/APIProtocol.swift
312924 Sources/GithubTypesMajorSplit/Components.swift
133735 Sources/GithubTypesMajorSplit/Operations.swift
    15 Sources/GithubTypesMajorSplit/Servers.swift
457166 total

% swift package clean && swift package update && swift build --skip-update --target GithubTypesMajorSplit
Build complete! (412.02s)
```


```console
% wc -l Sources/GithubTypesComponentsSplit/*
 10492 Sources/GithubTypesComponentsSplit/APIProtocol.swift
    13 Sources/GithubTypesComponentsSplit/Components.swift
    30 Sources/GithubTypesComponentsSplit/Headers.swift
133735 Sources/GithubTypesComponentsSplit/Operations.swift
   657 Sources/GithubTypesComponentsSplit/Parameters.swift
    15 Sources/GithubTypesComponentsSplit/RequestBodies.swift
[...]

PR #866, opened 2026-02-22
[Generator][Experimental] Add opt-in dependency-aware sharded Types generation

jpsim:
## Summary

This PR adds an **experimental**, fully opt-in sharded `Types` generation mode intended for very large OpenAPI specs.

Default behavior is unchanged. This is an additive path for advanced users who need better generated-code compile times and are willing to adopt custom build integration.

## Why experimental

This feature targets a narrow, high-scale use case and introduces additional generation topology (layering + sharding + inter-file imports). Marking it experimental lets us:
- gather real-world feedback before committing to long-term API stability,
- iterate on naming/layout heuristics with lower compatibility pressure,
- keep default maintenance burden low for the core project.

## Real-world motivation / observed impact

In our production use case (large spec, filtered subset), generated output compilation was a major bottleneck.
With dependency-aware layering + sharding, we observed **>2x faster compilation** of generated output.

Example environment and workload:
- OpenAPI source: ~18MB (filtered to ~300 operations for generation)
- Baseline: ~45s generation + ~6m30s release compile of generated output
- Sharded mode: generated output compile time reduced by more than half (roughly ~3m in this workload)

These results are workload- and build-system-dependent, but they motivated proposing this upstream as an optional path for similarly large deployments.

### Benchmark snapshot (real-world workload)

The chart below shows compile time of generated output in our production project (not generator runtime), measured on the same filtered spec input.

- Baseline: single-file `Types` output
- Variant: dependency-aware sharded output
- Observed: ~6m30s -> ~3m (more than 2x faster)

> Note: this is one workload and build setup; absolute gains will vary by project and toolchain.

Ramp benchmark comparing before vs after compile time for generated swift-openapi output (single-file vs dependency-aware sharded output)

## Scope of this PR

- Add `ShardingConfig` to `Config` (opt-in only), with validation.
- Add dependency-graph-based partitioning primitives used for sharding.
- Add sharded `Types` output path (root + component/type/operation shard files).
- Add CLI/config plumbing for explicit opt-in sharding.
- Add tests for algorithms, config validation, and sharded generation behavior.

## Non-goals

- No change to default single-file output behavior.
- No auto-sharding heuristics in this PR.
- No promise of stable file naming/layout semantics yet (experimental).

## Stability / compatibility

- If `sharding` is not set, behavior is unchanged.
- Experimental sharded output is explicitly subject to iteration.
- Existing users are unaffected unless they opt in.

## Maintenance posture

- Isolated implementation and tests to limit impact on the main generation path.
- Happy to adjust scope, naming, or API shape to align with maintainer preference.
- If preferred, this can be additionally gated/documented under an explicit “Experimental” section.


jpsim:
Here's a little demo with the App Store Connect OpenAPI schema:

```console
$ ./bench.sh
=== Environment ===
swift-driver version: 1.127.14.1 Apple Swift version 6.2.3 (swiftlang-6.2.3.3.21 clang-1700.6.3.2)
CPUs: 14

=== Downloading Apple App Store Connect spec ===

=== Generating types ===
Writing data to file Types.swift...
Writing data to file Operations_L4_2_16.swift...
Baseline: 1 file(s), 687829 lines
Sharded:  659 file(s), 695081 lines

=== Building packages (caching dependencies) ===
Building for debugging...
Build of target: 'ASCTypes' complete! (369.43s)
Building for debugging...
Build of target: 'ASCTypes' complete! (55.58s)

=== Benchmark ===
Benchmark 1: baseline (1 file)
  Time (mean ± σ):     170.237 s ±  0.293 s    [User: 200.128 s, System: 5.721 s]
  Range (min … max):   169.901 s … 170.439 s    3 runs

Benchmark 2: sharded (659 files)
  Time (mean ± σ):     83.448 s ±  4.814 s    [User: 291.162 s, System: 14.737 s]
  Range (min … max):   79.791 s … 88.903 s    3 runs

Summary
  sharded (659 files) ran
    2.04 ± 0.12 times faster than baseline (1 file)

Note: This measures forced full-rebuild time of the generated module.
Dependencies are pre-compiled and cached. Only the generated code recompiles each run.
Sharding enables parallel compilation across 14 cores.
```

<details>
<summary>bench.sh</summary>

```bash
#!/usr/bin/env bash
set -euo pipefail

# Benchmark: sharded vs baseline compilation of generated OpenAPI code
# Prerequisites: swift-openapi-generator (with --sharding), hyperfine, swift 6.x
# Install: brew install hyperfine

DIR=/tmp/asc-sharding-bench.noindex
SPEC_URL="https://developer.apple.com/sample-code/app-store-connect/app-store-connect-openapi-specification.zip"

echo "=== Environment ==="
swift --version | head -1
echo "CPUs: $(sysctl -n hw.ncpu 2>/dev/null || nproc)"

# Download & fix spec (ASC spec has empty enum:[] which produces invalid Swift)
echo -e "\n=== Downloading Apple App Store Connect spec ==="
rm -rf "$DIR" && mkdir -p "$DIR/spec"
curl -sL "$SPEC_URL" -o "$DIR/spec.zip"
unzip -qo "$DIR/spec.zip" -d "$DIR/spec/"
jq 'walk(if type == "object" and (.enum | type) == "array" and (.enum | length) == 0
         then del(.enum) else . end)' \
  "$DIR/spec/openapi.oas.json" > "$DIR/spec/openapi-fixed.json"

# Generate baseline (single file) and sharded types
echo -e "\n=== Generating types ==="
mkdir -p "$DIR/baseline/Sources/ASCTypes" "$DIR/sharded/Sources/ASCTypes"
swift-openapi-generator generate --mode types \
  --output-directory "$DIR/baseline/Sources/ASCTypes" \
  "$DIR/spec/openapi-fixed.json" 2>&1 | grep "Writing" | tail -1
swift-openapi-generator generate --mode types --sharding 4,4,4,2,2,1 \
  --output-directory "$DIR/sharded/Sources/ASCTypes" \
  "$DIR/spec/openapi-fixed.json" 2>&1 | grep "Writing" | tail -1

# Report
BASELINE_FILES=$(find "$DIR/baseline/Sources" -name "*.swift" | wc -l | tr -d ' ')
SHARDED_FILES=$(find "$DIR/sharded/Sources" -name "*.swift" | wc -l | tr -d ' ')
BASELINE_LINES=$(wc -l < "$DIR/baseline/Sources/ASCTypes/Types.swift" | tr -d ' ')
SHARDED_LINES=$(cat "$DIR/sharded/Sources/ASCTypes/"*.swift | wc -l | tr -d ' ')
echo "Baseline: $BASELINE_FILES file(s), $BASELINE_LINES lines"
echo "Sharded:  $SHARDED_FILES file(s), $SHARDED_LINES lines"

# Create Package.swift for both
PKG='// swift-tools-version:5.10
import PackageDescription
let package = Package(
    name: "ASCBenchmark", platforms: [.macOS(.v13)],
    dependencies: [
        .package(url: "https://github.com/apple/swift-openapi-runtime", from: "1.8.2"),
        .package(url: "https://github.com/apple/swift-http-types", from: "1.0.2"),
    ],
    targets: [.target(name: "ASCTypes", dependencies: [
        .product(name: "OpenAPIRuntime", package: "swift-openapi-runtime"),
        .product(name: "HTTPTypes", package: "swift-http-types"),
    ], path: "Sources/ASCTypes")]
)'
echo "$PKG" > "$DIR/baseline/Package.swift"
echo "$PKG" > "$DIR/sharded/Package.swift"

# Build both packages (caches dependencies)
echo -e "\n=== Building packages (caching dependencies) ==="
swift build --package-path "$DIR/baseline" --target ASCTypes 2>&1 | grep -E "^Build|^error:" || true
swift build --package-path "$DIR/sharded" --target ASCTypes 2>&1 | grep -E "^Build|^error:" || true

# Benchmark: forced full rebuild of ASCTypes module only
echo -e "\n=== Benchmark ==="
TOUCH="find $DIR/baseline/Sources -name '*.swift' -exec touch {} +"
TOUCH="$TOUCH && find $DIR/sharded/Sources -name '*.swift' -exec touch {} +"
hyperfine --warmup 1 --runs 3 --prepare "$TOUCH" \
  -n "baseline ($BASELINE_FILES file)" \
  "swift build --package-path $DIR/baseline --target ASCTypes 2>/dev/null" \
  -n "sharded ($SHARDED_FILES files)" \
  "swift build --package-path $DIR/sharded --target ASCTypes 2>/dev/null"

echo -e "\nNote: This measures forced full-rebuild time of the generated module."
echo "Dependencies are pre-compiled and cached. Only the generated code recompiles each run."
echo "Sharding enables parallel compilation across $(sysctl -n hw.ncpu 2>/dev/null || nproc) cores."
```

</details>

So there's benefits to this file sharding even while keeping a single library, but with release mode and WMO you need to split into separate modules to fully benefit there too.

jpsim:
I’m marking this as ready for review to start the discussion with maintainers. @czechboy0 @simonjbeaumont I’d love to hear your thoughts on how we can move forward here!

simonjbeaumont:
> I’m marking this as ready for review to start the discussion with maintainers. @czechboy0 @simonjbeaumont I’d love to hear your thoughts on how we can move forward here!
 
Ack.

First off—thank you for picking this work up! It sadly fell prey to being on my back-burner long enough that it simply dropped off any burner.

To be transparent, it'll take me a little time to get to this. I'll try and find some time next week to do an initial review. In the meantime could you answer a few questions before I dive deep into the code:

1. How much benefit do you think the graph-based sharding is bringing and is the increased complexity worth it? Is there a more naive approach that yields similar benefits?
2. Similarly, does it need to have so many configuration options? I haven't done a thorough review yet, but at first glance I think users might find these a bit unapproachable. In your experience, do you tweak them, or do we think there's a good set of values that will work generally or a number of shards that we could infer from the parsed OpenAPI document automatically?
3. There's some discussion in the PR description about single-module, multiple-files vs. multiple-modules and the interplay of that decision with WMO and release builds. Can we clarify the scope of this feature — are we planning to shard across modules? I originally thought this would just be splitting the generated code into multiple source files within the same module and, in my earlier PoCs, got comparable wins to what you're seeing here.
4. I think, once we've done an initial pass, we'd want to take a look at extending the reference tests to show this sharding in a simple scenario. We can defer it for now, but would you be open to adding these tests later?

Just adding for a little more context. In my PoC I was doing this as a post-processing step in the generator pipeline IIRC. Essentially, for each generated file (e.g. `Types.swift`), shard it, purely based on number of decls, which was configurable. This produced `Types.0.swift`, `Types.1.swift`, etc. This was a pretty straightforward approach and yielded pretty much the same compilation benefits.

It got bogged down because I scope-crept that in an attempt to parallelise the _generation_ itself, which does lend itself quite well to parallelisation.

I wonder if you have any thoughts about this simpler approach?

jpsim:
Thanks for the great questions, Si! I ran benchmarks across 4 configurations to answer them with data. The headline finding: **in release builds with WMO, only separate modules provide parallelism** — file splitting alone barely helps.

Here's why. WMO merges all files in a module back into one compilation unit, so the ~650-file layered split (1 module) only beats naive split by 16% in release. But distributing those files across ~35 dependency-aware modules cuts release time by 58% vs naive split.

All builds on M4 Pro (14 cores), 1 warmup + 2 measured iterations, ~690k lines of generated Swift:

codegen-comparison-debug

codegen-comparison-release

| Scenario | Debug | Release (WMO) |
|---|---|---|
| Monolithic (1 file, 1 module) | 5m24s | 12m01s |
| Naive split (~20 files, 3 modules) | 3m07s | 7m15s |
| Layered split (~650 files, 1 module) | 2m15s | 6m06s |
| Layered modules (~650 files, ~35 modules) | 1m33s | 3m01s |

Now to your questions:

**1. Graph-based vs naive sharding** — In debug, layered split beats naive split (2m15s vs 3m07s, 28% faster) because ~650 independent files parallelize better than ~20, despite using *fewer* modules. In release the gap narrows to 16% since WMO collapses files anyway. The real payoff is that graph-based sharding *enables* multi-module builds — each layer only depends on previous layers, so consumers can distribute files across separate build targets.

**2. Config options** — The current shard counts were found experimentally. I think we can auto-derive sensible defaults from the parsed document (decl count + dependency depth) so users don't need to tune. Happy to iterate here.

**3. Single-module vs multi-module scope** — The generator's scope in this PR is just file splitting within a single module (same as your PoC). The multi-module part is consumer-side config — in our case, Bazel targets that map to the generated dependency layers. But the benchmarks show this is where the release wins live: layered modules is 51% faster than layered split under WMO (3m01s vs 6m06s).

**4. Reference tests** — Absolutely, happy to add these as a follow-up.

**Re: your simpler approach** — Naive splitting captures a solid chunk of the debug win (5m24s → 3m07s). But in release it leaves most of the performance on the table (7m15s vs 3m01s for layered modules). This PR is more of a sketch showing what's possible when you optimize for compilation parallelism end-to-end — not intended to merge as-is. Happy to collaborate on the right incremental path, whether that's starting with naive splitting and layering in graph-based sharding later, or something else entirely.

czechboy0:
JP, what's the main priority for your use case? Build times in debug mode, when developers iterate locally, and only a few files change between rebuilds, or the build times of clean release builds? I suspect the solutions will be different for each, since the bottlenecks are in different places.

jpsim:
Certainly the value from speeding up debug builds for local iteration is the most impactful win I'm after.

However, speeding up release builds is not insignificant either, as demonstrated by the benchmarks I've shared above.

From my experimentation those two use cases share a common optimal solution.

czechboy0:
The tricky part is that with the build plugin approach, it's easy to do sharding inside a single module, but I don't think it's possible to coordinate between multiple modules. That's why I wonder if we first focus on what we can achieve with file splitting in one module (which helps debug builds), as everyone would be able to adopt that with a single config opt-in.

We can then discuss any multi-module splitting separately.

jpsim:
That makes total sense to me, and in fact is what this PR does. Consumers, including us, can then derive module graph from the files produced and the manifest. Or ignore it and still benefit from the single-module shards.

simonjbeaumont:
I'd be in favour of us designing this feature with an incremental approach and in such a way where most people can get a benefit from something simple, but folks who are highly motivated can use a more complex solution to unlock even more benefit.

Specifically to this point. When @czechboy0 and I discussed this before, we discussed that the number and names of generated files was never part of our stable API guarantee, so we could, in theory, enable something by default. E.g. we _could_ explore generating `Types.{0..N}.swift` instead of `Types.swift` by default. I'd like us to revisit this discussion and potentially even this belongs behind a config option.

But I'd like for this simpler mode to be (a) easy to enable, and (b) easy to understand.

I'm not shutting down enabling a much more complex, opt-in workflow, for people using external build systems, so long as (a) it's not that-or-nothing, and (b) we don't have a huge complexity hike in the codebase.

IIRC in my manual sharding experiments, we saw a pretty big win for both debug and release builds with a simpler approach. Granted it's not the full win your showing but it's a nice point on the curve that I feel we should offer with a low cost of adoption.

We can see some of that here from your results:

> | Scenario | Debug | Release (WMO) |
> |---|---|---|
> | Monolithic (1 file, 1 module) | 5m24s | 12m01s |
> | Naive split (~20 files, 3 modules) | 3m07s | 7m15s |
> | Layered split (~650 files, 1 module) | 2m15s | 6m06s |
> | Layered modules (~650 files, ~35 modules) | 1m33s | 3m01s |

We can see with your single-module example it produces a 2x speedup for both debug _and_ release, no? I think this is very much worth offering.

However, I think we're missing some important datapoints here. We only have one measurement for a single-module split, with ~650 files. IMHO this isn't going to produce a user-friendly experience for the casual adopter.

From my previous investigation (included below) a single-module split produces pretty big wins, but tails off with the number of shards.

<details>
<summary>My previous investigation</summary>

> FWIW I think this is a reasonable ask. Experimentally, it seems that the compiler is having a much harder time with one large file (cf. more, smaller files). An example on my machine shows pretty substantial improvement:
> 
> ```console
> % wc -l Sources/GithubTypesOneFile/*
> 457133 Sources/GithubTypesOneFile/Types.swift
> 
> % swift package clean && swift package update && swift build --skip-update --target GithubTypesOneFile
> Build complete! (647.71s)
> ```
> 
> ```console
> % wc -l Sources/GithubTypesMajorSplit/*
>  10492 Sources/GithubTypesMajorSplit/APIProtocol.swift
> 312924 Sources/GithubTypesMajorSplit/Components.swift
> 133735 Sources/GithubTypesMajorSplit/Operations.swift
>     15 Sources/GithubTypesMajorSplit/Servers.swift
> 457166 total
> 
> % swift package clean && swift package update && swift build --skip-update --target GithubTypesMajorSplit
> Build complete! (412.02s)
> ```
> 
> 
> ```console
> % wc -l Sources/GithubTypesComponentsSplit/*
>  10492 Sources/GithubTypesComponentsSplit/APIProtocol.swift
>     13 Sources/GithubTypesComponentsSplit/Components.swift
>     30 Sources/GithubTypesComponentsSplit/Headers.swift
> 133735 Sources/GithubTypesComponentsSplit/Operations.swift
>    657 Sources/GithubTypesComponentsSplit/Parameters.swift
>     15 Sources/GithubTypesComponentsSplit/RequestBodies.swift
>    491 Sources/GithubTypesComponentsSplit/Responses.swift
> 311782 Sources/GithubTypesComponentsSplit/Schemas.swift
>     15 Sources/GithubTypesComponentsSplit/Servers.swift
> 457230 total
> 
> % swift package clean && swift package update && swift build --skip-update --target GithubTypesComponentsSplit
> Build complete! (422.18)
> ```
> 
> ```console
> % wc -l Sources/GithubTypesSchemasSplit/*
>  10492 Sources/GithubTypesSchemasSplit/APIProtocol.swift
>     13 Sources/GithubTypesSchemasSplit/Components.swift
>     30 Sources/GithubTypesSchemasSplit/Headers.swift
> 133735 Sources/GithubTypesSchemasSplit/Operations.swift
>    657 Sources/GithubTypesSchemasSplit/Parameters.swift
>     15 Sources/GithubTypesSchemasSplit/RequestBodies.swift
>    491 Sources/GithubTypesSchemasSplit/Responses.swift
> 100277 Sources/GithubTypesSchemasSplit/Schemas.1.swift
> 100545 Sources/GithubTypesSchemasSplit/Schemas.2.swift
> 110984 Sources/GithubTypesSchemasSplit/Schemas.3.swift
>     15 Sources/GithubTypesSchemasSplit/Schemas.swift
>     15 Sources/GithubTypesSchemasSplit/Servers.swift
> 457269 total
> 
> % swift package clean && swift package update && swift build --skip-update --target GithubTypesSchemasSplit
> Build complete! (140.95s)
> ```
> 
> 
> ```console
> % wc -l Sources/GithubTypes50kSplit/*
>  10492 Sources/GithubTypes50kSplit/APIProtocol.swift
>     13 Sources/GithubTypes50kSplit/Components.swift
>     30 Sources/GithubTypes50kSplit/Headers.swift
>  50029 Sources/GithubTypes50kSplit/Operations.1.swift
>  50135 Sources/GithubTypes50kSplit/Operations.2.swift
>  33596 Sources/GithubTypes50kSplit/Operations.3.swift
>     13 Sources/GithubTypes50kSplit/Operations.swift
>    657 Sources/GithubTypes50kSplit/Parameters.swift
>     15 Sources/GithubTypes50kSplit/RequestBodies.swift
>    491 Sources/GithubTypes50kSplit/Responses.swift
>  50162 Sources/GithubTypes50kSplit/Schemas.1.1.swift
>  50128 Sources/GithubTypes50kSplit/Schemas.1.2.swift
>  50079 Sources/GithubTypes50kSplit/Schemas.2.1.swift
>  50479 Sources/GithubTypes50kSplit/Schemas.2.2.swift
>  55450 Sources/GithubTypes50kSplit/Schemas.3.1.swift
>  55546 Sources/GithubTypes50kSplit/Schemas.3.2.swift
>     15 Sources/GithubTypes50kSplit/Schemas.swift
>     15 Sources/GithubTypes50kSplit/Servers.swift
> 457345 total
> 
> % swift package clean && swift package update && swift build --skip-update --target GithubTypes50kSplit
> Build complete! (116.46s)
> ```
> 
> I think we have other features we can explore too (e.g. only generating the bits that users want to use), which will have much bigger impact, but I think we should keep this on the backlog.
> 

</details>

— source: https://github.com/apple/swift-openapi-generator/issues/253#issuecomment-1729861562

That was with GitHub API and going from 647s to 116s by splitting based on the structure of the OpenAPI _spec_ (rather than any dependencies), i.e. components, operations, schemas, etc., and furhter sharding them according to a max number of decls per file. IMHO this output structure is more approachable than ~650 files and, in this instance, it sharded each section a maximum of 6 times, producing 18 files total, and resulted in a 82% speedup in build times.

There is no dependency analysis needed for this approach whatsoever IIRC. 

I think, in summary, I'd be in support of us offering this "naive" splitting first, but designing the internals and config API in a way to make room for something more complex.

WDYT @jpsim @czechboy0

ThomasDutartre:
Really interesting PR! In my situation, I immediately wondered if there was a better way to handle generation, as a 23,000-line file is not exactly fun to work with.

simonjbeaumont:
> Really interesting PR! In my situation, I immediately wondered if there was a better way to handle generation, as a 23,000-line file is not exactly fun to work with.

Just checking you're aware of the existing filtering feature. For many use cases this can vastly reduce the generated code by specifying only the operations you need.

jpsim:
Hi folks, since I last posted this we've increased the size of our OpenAPI schema and number of endpoints. Now on an M4 Pro applying these optimizations release compile times drop from 15m54s to 3m46s, so this is still very much something we'd like to work on landing.

image

I'd like to introduce you to @nac5504 from our team, who will be working on productionizing this approach in small increments over the next few weeks.

simonjbeaumont:
> Hi folks, since I last posted this we've increased the size of our OpenAPI schema and number of endpoints. Now on an M4 Pro applying these optimizations release compile times drop from 15m54s to 3m46s, so this is still very much something we'd like to work on landing.
> ...
>
> I'd like to introduce you to @nac5504 from our team, who will be working on productionizing this approach in small increments over the next few weeks.

Thanks for the update @jpsim and hi, @nac5504 👋.

@jpsim, just wanted to check you saw my reply here a while back, which ended with:

> I think, in summary, I'd be in support of us offering this "naive" splitting first, but designing the internals and config API in a way to make room for something more complex.
— https://github.com/apple/swift-openapi-generator/pull/866#issuecomment-4022467320

TL;DR, I'd like for us to consider a simpler approach _first_, where we get as much milage out of a single-module sharding approach first. I think the majority of users will get a huge benefit from this, with a reasonable heuristic to make it approachable. Then we can add some tuning parameters to that. Then, we can discuss whether we need to pull in the layered module approach for advanced use cases.

nac5504:
Hi @simonjbeaumont, excited to be working with you! I agree the best first step is adopting the "naive splitting" approach mentioned above, as it seems like we both use some version of this post-generator already.

Our current setup hardcodes the number of output shard files, but you mentioned your earlier PoC split based on a number of declarations per file. What do you think the best config interface should be for this first version?

Do you think the initial config API should be as small as something like `typesFileSplitting: true`, with the generator choosing the heuristic internally? Then, if we receive feedback wanting specific splitting patterns, we can add tuning options from there.

simonjbeaumont:
I think there are many strategies here:
1. Splitting the types file without declaration sharding. E.g. `Types.swift` might be replaced with one file per current top-level type, e.g. `Operations.swift`, `Components.swift`, etc.
2. These files could then be further split without declaration sharding, e.g. by having `Components.swift` declaring just an empty `enum Components {}` and the next layer of types be provided in extensions, e.g. `Component+Schemas.swift`, `Components+Parameters.swift`, etc.
3. Then we could consider some form of sharding, where we follow a similar empty-namespace-and-extensions pattern, e.g. `Components+Schemas.swift` might declare simply `extension Components { enum Schemas {} }` and we provide `Components+Schemas.{0..N}.swift`. How we choose to shard is still a choice to be made. Possibly by choosing N, or by choosing max declarations per-shard, M, and have N grow accordingly.

My hunch is we get most of the win from the above (possibly with exposing some options for N or M).

We should consider _where_ this logic will live. Does it affect the generation or will it be added to `postTransitionHooks:` in `makeGeneratorPipeline`?

We should take a beat and also consider Swift Syntax here as an option for adopters who want something more advanced. When Swift OpenAPI first was released Swift Syntax was too expensive to have as a dependency but given we'll very soon be raising the floor of this package to Swift 6.2, which bundles prebuilt swift-syntax in the toolchain, we should see if this is now a better option. Given the transformations we're talking about are rewriting Swift source files based on declarations, I could imagine a pretty generic _sharder_ tool and/or plugin.

nac5504:
Hi @simonjbeaumont,

Thank you for your advice for how to kick this off. I am hoping to land most of the improvements mentioned in this PR within the next 4 weeks, so I'm looking forward to getting more feedback from you as the PRs go live. I opened #925 as a first pass at the initial strategy you mentioned: splitting the generated types output by the current top-level namespaces, so `Types.swift` remains the root file and `Components` / `Operations` move into `Types+Components.swift` and `Types+Operations.swift`.

I intentionally kept this first PR away from dependency-based sharding/layering. My hope is that the namespace split gives us a reviewable first step that validates the multi-output pipeline, config shape, and CLI/write-path changes without deciding the harder `N` vs `M` sharding API yet.

One important constraint in this PR is that the new splitting config is only supported through the CLI/command-plugin path, not the SwiftPM build-tool plugin path. That seemed aligned with the idea that this is an advanced opt-in configuration for users who are already thinking about generated output shape and build performance. Additionally, the build-tool plugin has to declare its output files before invoking the generator, so supporting dynamic splitting there needs a separate design rather than quietly producing undeclared files. This constraint is something I'm hoping to discuss further how we want to add this feature to SwiftPM users in the future, I opened an [additional draft PR](https://github.com/nac5504/swift-openapi-generator/pull/5) with a potential solution.

For where the logic lives: in #925 it lives within the types translator rather than `postTransitionHooks`. That felt like the right layer for this change because the translator is where we still have the semantic pieces before they are flattened into Swift text: the root API declarations, `Components`, `Operations`, shared imports, top comments, and the configured output file names. The split is really a choice about how the types translator emits its structured output, not a generic pipeline transformation.

A post-render hook would have to work from rendered Swift source, rediscover declaration boundaries, preserve/duplicate imports and file comments, and then produce a new output-file set. That seems both more fragile and more coupled to renderer formatting.

I also spent some time looking at a Swift Syntax based approach. It is promising as a future tool for generic source rewriting, but it did not seem like the right fit for this type of change. This strategy is based on information the generator already has before rendering: which declarations are the root API surface, which declaration is `Components`, and which declaration is `Operations`. Doing the split there avoids turning structured generator output into Swift text and then immediately parsing that text back into a syntax tree to recover the same boundaries.

Update: Thanks for the quick review already, taking a look at your comments now :)

cc @jpsim

PR #867, opened 2026-02-24
Moved to Swift 6 language mode

czechboy0:
### Motivation

Swift 6 language mode allows us to keep the codebase modern and catch more issues at compile time.

### Modifications

Upgrade the codebase to Swift 6 language mode and Swift 6.0+ toolchains.

### Result

A codebase that is closer to modern best practices now.

### Test Plan

Tests passed locally, will verify the rest of platforms on PR CI.


Package.swift
```diff
@@ -154,15 +139,27 @@ let package = Package(
...
-        settings.append(.enableUpcomingFeature("MemberImportVisibility"))
-        target.swiftSettings = settings
-    case .macro, .plugin, .system, .binary: ()  // not applicable
-    @unknown default: ()  // we don't know what to do here, do nothing
+    case .executable, .regular, .test:
+        break
+    default:
+        continue
     }
-}  // --- END: STANDARD CROSS-REPO SETTINGS DO NOT EDIT --- //
```
simonjbeaumont, line 168: Can't drop this. Tooling uses it.

Package.swift
```diff
@@ -154,15 +139,27 @@ let package = Package(
...
+    // Require `any` for existential types.
+    settings.append(.enableUpcomingFeature("ExistentialAny"))
+
+    // https://github.com/swiftlang/swift-evolution/blob/main/proposals/0444-member-import-visibility.md
+    settings.append(.enableUpcomingFeature("MemberImportVisibility"))
+
+    if target.name != "PetstoreConsumerTests" {
+        // https://github.com/swiftlang/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md
+        settings.append(.enableUpcomingFeature("InternalImportsByDefault"))
+    }
```
simonjbeaumont, line 175: Do we want/need this? There's no supported library product in this package.

simonjbeaumont (changes requested):

Package.swift
kukushechkin: I would keep it to be consistent with other repos.

Package.swift
simonjbeaumont: I can see you restored the fence but you've disregarded the "DO NOT EDIT". Your changes will be overwritten. You _can_ append to the settings, but you must do so _outside_ of this block.

kukushechkin: Ag, ok, do you mean the whole block? Could you check if I understood you correctly now what to restore?

simonjbeaumont (approved):

PR #839, opened 2025-10-22
Update OpenAPIKit to v4

iT0ny:
## Context
Duplicate of apple/swift-openapi-generator#830. This PR implements the additional changes/fixes for issues surfaced in the original PR

## Summary
- Move OpenAPIKit to 4.x
- Refresh handling of null values
- Fix tests

## Details
- Bump OpenAPIKit version so we target 4.x, and raise the Yams requirement to 5.1+
- Update wording in `testEmitsComplexOpenAPIParsingError` and `testSchemaWarningsForwardedToGeneratorDiagnostics` (`Tests/OpenAPIGeneratorCoreTests/Parser/Test_YamsParser.swift`, `Tests/OpenAPIGeneratorCoreTests/Translator/TypesTranslator/Test_translateSchemas.swift`)
- Update checks for null values in `Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateRawEnum.swift`
- Update OpenAPI.Content.Encoding usage according to the OpenAPIKit v4 migration guide  in `Tests/OpenAPIGeneratorCoreTests/Translator/TypeAssignment/Test_TypeMatcher.swift`

## Testing
- ✅ `swift test`
- CI run

Fixes #821

PR #875, opened 2026-03-09
Update OpenAPIKit to v6

mattpolzin:
This PR builds upon https://github.com/apple/swift-openapi-generator/pull/839 to update from OpenAPIKit 4.x to OpenAPIKit ~5.0~ 6.0.

### Motivation

Updating to OpenAPIKit v4.x enables efforts to add external loading support to the generator (as a separate future project). This means being able to read in OpenAPI Documents spread across multiple files.

Updating to OpenAPIKit v5.x allows the generator project to support OAS 3.2.0 features and take advantage of various other improvements to OpenAPIKit that required breaking changes to be made.

Updating to OpenAPIKit v6.0 allows this project to turn the `ExternalLoading` trait off (on by default). Turning this trait off saves 50% of the compile time of OpenAPKit for as long as `swift-openapi-generator` is not using the external loading features of OpenAPKit anyway.

### Modifications

- Update OpenAPIKit dependency to v6.x
- Remove code that parsed OAS 3.2.0 documents as 3.1.x documents; OpenAPIKit v5 supports OAS 3.2.0 documents natively.
- Make changes required by move to new OpenAPIKit version (following migration guide).
- Introduce `assumeLookupOnce()` to temporarily maintain the existing invariant that no entries in the Components Object are themselves references; replace all `lookup()` calls with `assumeLookupOnce()`. See the **Note** on this below.
- Fix code that looked for references on JSON Schemas under the assumption that they were wrapped in `OpenAPI.Reference`. This assumption does not hold in OpenAPIKit v5.

**NOTE on `assumeLookupOnce()`**
The new `assumeLookupOnce()` function is intended to allow for gradual migration. The generator had previously used `lookup()` which used to look at the Components Object and either return an entry or throw. Going forward, `lookup()` does recursive lookup until a non-reference is found and `lookupOnce()` acts more like `lookup()` used to act -- however, `lookupOnce()` has a different method signature owing to the fact that it may return a reference instead of the ultimately desired object. The need for these changes arise from the fact that OpenAPIKit v5 now allows entries in the Components Object to themselves be references. This has been supported by the OpenAPI Standard for some time but OpenAPIKit only introduced support for it with v5.

Therefore, `assumeLookupOnce()` was added as a way to throw an error if a reference is found as an entry in the Components Object (behavior already exhibited by the generator project) but the goal should be to remove uses of `assumeLookupOnce()` over time and replace them with either `lookupOnce()` or `lookup()` depending on the needs of each call site. In this way, the generator project will move closer to supporting Component Object reference entries without the overhead of changing every call site all at once.

### Result

Theoretically, nothing functionally changes.

### Test Plan

Tests updated, `swift test` passing, "Compatibility test" suite passing.


Package.swift
```diff
@@ -46,8 +46,8 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-collections", from: "1.1.4"),
 
         // Read OpenAPI documents
-        .package(url: "https://github.com/mattpolzin/OpenAPIKit", from: "3.9.0"),
-        .package(url: "https://github.com/jpsim/Yams", "4.0.0"..<"7.0.0"),
+        .package(url: "https://github.com/mattpolzin/OpenAPIKit", from: "5.1.1"),
+        .package(url: "https://github.com/jpsim/Yams", "5.1.0"..<"7.0.0"),
```
simonjbeaumont: Is this bump necessary? I notice that OpenAPIKit's manifest only uses Yams in tests. Do you know if there are actually any API _breaks_ between Yams 4 and 5. Looks like "no" from the release notes other than dropping Swift versions, which won't be a concern for us.

Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateRawEnum.swift
```diff
@@ -100,7 +100,9 @@ extension FileTranslator {
                 // In nullable enum schemas, empty strings are parsed as Void.
                 // This is unlikely to be fixed, so handling that case here.
                 // https://github.com/apple/swift-openapi-generator/issues/118
-                if isNullable && anyValue is Void {
+                // Also handle nil values in nullable schemas.
+                let isNullValue = anyValue is Void || (anyValue as? String) == nil
+                if isNullable && isNullValue {
```
simonjbeaumont, line 105: Is this new check correct? Won't `(anyValue as? String)` be `nil` for non-nullable non-string values?

mattpolzin: I'm taking a closer look at this now. I inherited this change from the v4 bump.

> non-nullable

From what I can tell, this possibility is ruled out because `isNullable` must be true on line `105`.

> non-string

We are in a case where the "backing type" has been determined to be a string, so I imagine that was the thinking with assuming the allowed values would be strings. I can say for sure that if the JSON Schema is of `string` type, then OpenAPIKit will special-case its parsing of `allowedValues` to only parse `String` (or `String?` if nullable) values. Tracing the calls in the generator project for `translateRawEnum()` I see that it does only pass `.string` as the backing type when OpenAPIKit has determined the schema type to be string, so this change actually should be safe and correct, albeit a little involved to sanity check!

Package.swift
mattpolzin: Ah, I made this change because OpenAPKit increased the minimum version -- doing so allowed building in Swift 6 mode without pre-concurrency annotations on Yams.

There should be no reason I need to make the same bump in _your_ package manifest, I can revert.

mattpolzin:
Thanks for the review. I'll fix the CI failures in follow-up commits (formatting fixed just now).

mattpolzin:
Alright, ready for another pass!

mattpolzin:
Interesting. Swift 6 is failing to identify the correct overload of a function. Fixable, one way or the other.

mattpolzin:
This is rather unfortunate. 
- Swift 6.0 on Linux and Windows has trouble resolving some extensions with helper functions that no other Swift versions have.
- Swift 6.3 unit tests, Compatibilty tests, and others are just not finishing (looks like GitHub runner issues like have become weirdly common lately).

I started in on "supporting" Swift 6.0, but it is unclear how much fiddly stuff would need to be moved out of extensions and into some less idiomatic location just to help the compiler find functions.

simonjbeaumont:
No need to worry about 6.0. I'm busy in the other PR dropping it completely. 

Once that's done we can focus on getting the others passing.

simonjbeaumont:
@mattpolzin I've merged #887 now and updated this PR branch so 6.0 is no longer an issue.

The fact the unit tests are not completing appears to be actually relevant to this PR as all the tests passed fine in #887.

mattpolzin:
Troubling. This represents a massive performance regression. The examples CI jobs that do succeed before timing out still take close to twice as long on this branch as they did on the branch you linked to.

Since it all compiles and works, I think I'll need to just make educated guesses at what might be slowing the compiler down and tweak those things.

mattpolzin:
There's something really subtle going on here that I have not figured out yet. In the most recent CI run, Linux-based "Release Build" for Swift 6.2 finished in `5m 40s` but for Swift 6.1 it timed out.

What's worse, I can't reproduce any of this on my Linux laptop inside a docker container (yet). Using a Swift 6.1 Docker Image I saw a `swift build` of this branch take 1.5 minutes and `swift test` took under 30s. But I am not sure I am correctly reproducing what is happening for the "Unit test" CI workflows because they are fairly complex (based on several layers of YAML configurations) and they may not in fact be analogous to `swift test`.

mattpolzin:
I 'd like to come up with a troubleshooting process that doesn't involve me pushing a hopeful change and seeing if it works after CI runs in GitHub.

mattpolzin:
Well, I've found something to go on I think. As a proxy for what's going on in CI (though the situation is not nearly as severe on my personal laptop) I can observe a meaningful difference in compile time between the `main` branch and this one when building within a Swift docker container.

Interestingly, over half the difference in build time can be observed going as far back as the bump to OpenAPIKit v4, the other half (or less) comes from bumping to OpenAPKit v5.

I'm worried this may have less to do with function overloading slowing the compilation down and more to do with the introduction of concurrency (mainly `Sendable`) support into OpenAPKit given that realization. Still more digging to do, but that's all I've got time for at the moment.

simonjbeaumont:
@mattpolzin yeah, there's a bunch of abstractions in the CI that come from our use of a shared set of GitHub Workflows configs from another repo and the use of matrix jobs. But they really should boil down to just `swift test` at the end. 

Worth noting that the compatibility tests require an environment variable to enable: `SWIFT_OPENAPI_COMPATIBILITY_TEST_ENABLE=true swift test`. They're interesting because they create a little harness with a real OpenAPI doc and go through the same steps as an adopter build (more or less).

Hopefully you have enough to work through locally. I wonder if it's worth having a look at some of the compiler flags, like `-warn-long-expression-type-checking` or `-profile-stats-entities`? 

Would be good if we can understand the impact here and then we can assess the blast radius. I don't _think_ CI should be the issue, but if we can identify a pathological setup in CI that isn't representative of users then maybe we can change CI.

mattpolzin:
This is not very promising news I'm afraid. I have bisected the performance degradation and the vast majority of it occurs between OpenAPIKit versions `4.0.0-alpha.4` and `4.0.0-alpha.5`. This makes me 98% confident that the reason Swift compilation is taking longer is almost entirely due to Swift Concurrency features. Most likely specifically `async let` uses in OpenAPIKit's external dereferencing code.

The numbers (on my personal laptop) of `swift build` of this project are:

| OpenAPIKit Version | User Time | Real Time |
|--------------------|-----------|-----------|
| `v4.0.0-alpha.4`   | 5m24s     | 0m35s     |
| `v4.0.0-alpha.5`   | 8m29s     | 0m53s     |
| `v5.3.0`           | 9m51s     | 1m9s      |

On the plus side, compilation does not take very long given `swift build` is run on a modern CPU with multiple cores available (Real Time above feels not terrible even though it is much worse than it was before concurrency code was added).

I'm not confident any good solution exists other than waiting for Swift to compile faster or removing concurrency (which I added to the external dereferencing code by request of Honza; and, it is definitely a good use-case for concurrency).

---

Any thoughts based on the above?

mattpolzin:
My first thought was add more type annotations. Annotating the async let lines with their types does not make any difference it doesn't look like.

My next thought is conditional compilation -- at least until or unless a project wants to use OpenAPIKit to perform external dereferencing, the massive hit to compilation time may be avoidable. I'll experiment with  that at least enough to know if it makes much of a difference like I think it might.

mattpolzin:
Good news/Bad news.

I added a trait that conditionally compiles the external dereferencing code and that was _entirely_ the slowdown.

| `ExternalLoading` trait | User Compile Time | Real Compile Time |
|-------------------------|-------------------|-------------------|
| enabled                 | 9m58s             | 1m9s              |
| disabled                | 5m15s             | 0m31s             |

The bad news is that package traits are only available for Swift 6.1+ so to disable external loading we will need to bump `swift-openapi-generator` to OpenAPIKit v6, not v5. In v6 I will bump minimum Swift version to 6.1 and add the new trait.

The other bad news, naturally, is that Swift compilation of concurrency features is so slow that it might be considered prohibitive to enable an OpenAPIKit feature that would allow `swift-openapi-generator` to load external OpenAPI references.

simonjbeaumont:
Oof. Thanks for the update, Matt! IIUC that means we have an incremental path forward. Swift OpenAPI Generator already is 6.1+ so we could take on a trait when bumping the OpenAPIKit dependency. 

I agree it's a shame it won't be _as_ straightforward to investigate support for external references. We could explore using traits in Swift OpenAPI Generator too if the cost is prohibitively high to enable by default?

mattpolzin:
It'll be on by default in OpenAPIKit because I don't want to assume that any given project will be driven by a desire to keep compilation time low. But I'll disable it when bringing v6 of OpenAPIKit into the generator project because that'll make it much easier to justify the update (CI won't take substantially longer or time out for you). 

I figure then it's up to y'all when you are ready to sink time into external loading features whether the compilation time cost is appropriate or not.

mattpolzin:
Alright, I've bumped to OpenAPIKit v6 now. Major version 6 does not involve any breaking changes. It bumps the minimum Swift version to 6.1 (already this project's minimum requirement) and it encloses external loading features in an `ExternalLoading` package trait (on by default, but turned off for `swift-openapi-generator` for now to save a lot on compile time).

mattpolzin:
@simonjbeaumont looks like there's some sort of caching issue in CI maybe? The "Examples" job for Swift 6.1 succeeds (resolving OpenAPIKit to version 6.0.0) but the Examples job for Swift 6.2 fails (resolving OpenAPIKit to version 3.9.0).

simonjbeaumont:
Looking at the examples CI, this looks to be because the examples share a common harness and, on each example, do a `swift package unedit ... || swift package edit ...`, with the `||` there to catch the first time it's not in edit mode. Both of these commands include `--skip-update` which I think is the main problem. When we `swift package edit` in the `swift-openapi-generator` branch for this PR it doesn't update the dependencies, which clearly it needs to. Then it tries to declare the dependency on OpenAPIKit disabling traits, but the cached version had no traits, which is a SwiftPM error.

Let me see if I can get a PR up to fix that.

simonjbeaumont:
Thank you so much for driving this effort @mattpolzin and for your patience as we navigated through the changes. I think this is good to land now. ❤️ 🙏

simonjbeaumont (approved):

simonjbeaumont:
Also—thank you to @iT0ny!

PR #880, opened 2026-03-18
Use OpenAPIKit for more validation logic

mattpolzin:
### Motivation

OpenAPIKit provides some validation help in its latest versions that can be used here to reduce the amount of logic this project must maintain. For example, the generator need not crawl the document to validate all content types, this is the behavior of a `Validation<OpenAPIKit.ContentType>`.

### Modifications

- I've removed the reference validations entirely because they are now available in exactly the desired form upstream.
- I replaced the existing implementation of `ContentType` check with one using OpenAPIKit's validation DSL. This mostly serves to avoid the need to maintain logic in this library that OpenAPIKit must maintain anyway (namely, there's no need to find the `ContentType` values to validate them, OpenAPIKit will validate all values of that type anywhere in the Document).

### Result

Removed code, slightly different error messages.

### Test Plan

Tests have been updated to reflect the new (though comparable) errors.

### DIFF
--- a/Tests/OpenAPIGeneratorCoreTests/Parser/Test_validateDoc.swift
+++ b/Tests/OpenAPIGeneratorCoreTests/Parser/Test_validateDoc.swift
@@ -143,10 +143,11 @@ final class Test_validateDoc: Test_Core {
             ],
             components: .noComponents
         )
+        let validator = Validator.blank.validating(validateContentTypes() { contentType in
+            (try? _OpenAPIGeneratorCore.ContentType(string: contentType)) != nil
+        })
         XCTAssertNoThrow(
-            try validateContentTypes(in: doc) { contentType in
-                (try? _OpenAPIGeneratorCore.ContentType(string: contentType)) != nil
-            }
+            try doc.validate(using: validator, strict: false)
         )
     }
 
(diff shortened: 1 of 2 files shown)


simonjbeaumont:
Realise this is backing off another WIP PR but, after a preliminary glance, this seems like a win to me—thanks!

Tests/OpenAPIGeneratorCoreTests/Parser/Test_validateDoc.swift
```diff
@@ -30,16 +30,17 @@ final class Test_validateDoc: Test_Core {
                 "All Operation Ids in Document are unique",
                 "Server Variable\'s enum is either not defined or is non-empty (if defined).",
                 "Server Variable\'s default must exist in enum, if enum is defined.",
-                "JSONSchema reference can be found in components/schemas",
```
mattpolzin, line 33: Note that each of these reference validations has been replaced by the more strict variant that the generator project desires (by using `validatingAllReferencesFoundInComponents()`). The phrasing of the validation description changes from "reference can be found in components/..." to "reference points to this document and can be found in components/..."

Other than the phrasing change that reflects that stricter validation, the diff reflects that the generator-defined "Content type is of form type/subtype" is also being run.

mattpolzin:
@simonjbeaumont absolutely no hurry on this PR, but I wanted to ping you since I left it open for quite a while as a draft but it has now been rebased against latest `main` and should be ready for review.

simonjbeaumont:
@mattpolzin amazing, thank you!

simonjbeaumont (approved):
LGTM!

PR #882, opened 2026-03-20
Drop Swift 5.10 and stop using deprecated PackagePlugin.Path

simonjbeaumont:
### Motivation

In #878 we split the manifest for this project so we could have the 6.1+ manifest declare a dependency on swift-openapi-runtime without default traits to support projects that want to avoid linking the full Foundation transitive dependency. 

The consequence of this is that the newer manifest lifted the `swift-tools-version`. There was an attempt to mitigate that by marking the Swift language version as v5 to decouple it from moving to 6. However, what this meant was we started getting warnings in the plugin code because plugin targets cannot have `swiftSettings` applied to them so they started complaining about the use of the deprecated `PackagePlugin.Path` APIs.

We previously could not move to URL, as requested by the deprecation warning, because we were still supporting Swift 5.10. But 5.10 dropped out of our support window 6 months ago. We haven't been CI'ing it, but we never lifted the tools version in the package manifest.

We can do that now, and then move to using `URL` in the plugin code, which should get rid of the warnings.

### Modifications

Bump the minimum version of Swift to 6.0. However, we still retain a split manifest because traits are only supported in 6.1+.

Then move to using URL instead of PackagePlugin.Path APIs.

### Result

Plugin should stop generating warnings for adopters on 6.1+.

### Test Plan

Integration tests still pass. Built locally and no longer see the warnings.

### Related

Expect this to fix #881

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -20,7 +20,6 @@ jobs:
     name: Unit tests
     uses: apple/swift-nio/.github/workflows/unit_tests.yml@main
     with:
-      linux_5_10_arguments_override: "--explicit-target-dependency-import-check error"
       linux_6_0_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
       linux_6_1_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
       linux_6_2_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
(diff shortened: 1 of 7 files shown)


czechboy0 (approved):

vdkdamian:
Seems like this created an issue for me where the `openapi-generator-config.yaml` file can't be found anymore due to my folder containing a space. It tries to find the path but I see it's percent-encoded.

```
Error: Configuration file not found at path: 
/Volumes/Developer/developer/Documents/Projects/Company/Company%20Project/company-project/Sources/Project/openapi-generator-config.yaml. 
Please ensure the file exists and the path is correct.
```

I'll rename the folder structure for now. This seems to solve it.

**Edit:**
Solved with https://github.com/apple/swift-openapi-generator/issues/899

PR #883, opened 2026-03-20
Add access modifiers to generated imports

arpit-garg-1995:
### Motivation

Starting in Swift 6, users can enable the `InternalImportsByDefault` upcoming feature flag, which makes all imports `internal` by default. When users generate code with an `accessModifier` of `public` or `package`, the generated declarations need to re-export the symbols they depend on. However, blindly applying the access modifier to every import causes a compiler error under `-warnings-as-errors`: `HTTPTypes` types appear only inside function bodies, never in public method signatures, so `public import HTTPTypes` is rejected by Swift 6.

### Modifications

- Add `setsAccessModifier: Bool = true` to `ImportDescription`; set to `false` for `HTTPTypes` in `clientServerImports` so the generator never emits `public import HTTPTypes` or `package import HTTPTypes`.
- Update `importDescriptions(adding:)` in `FileTranslator` to respect the flag when propagating the configured access modifier.
- Apply `swift-format` fixes to `FileTranslator.swift`, `ServerTranslator.swift`, and `Test_TextBasedRenderer.swift`.
- Update Petstore reference snapshots to match new generator output: `public import` for `OpenAPIRuntime` and Foundation types; plain `import` for `HTTPTypes`.
- Fix `Examples/streaming-chatgpt-proxy/Sources/ChatGPT/Middleware.swift` to use `package import` for Foundation and OpenAPIRuntime, matching the generator output for `package`-access targets.

### Result

Format check passes. All unit tests pass with `-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error`. The ChatGPT example builds correctly.

### Test Plan

`swift test -Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error` — 319 tests, 0 failures, verified on macOS and Linux (Swift 6.2 Docker).


simonjbeaumont (changes requested):
Thanks for taking the time to open the PR.

I'm sympathetic to the problem and open to a solution here. But I think this has been solved at the wrong layer.

The PR currently does this at the concrete implementation of a renderer and just unilaterally updates all imports to use a given access modifier.

This needs to be done in the types that conform to `FileTranslator`, e.g. see `ClientFileTranslator`, which is where the structured Swift representation is determined. You'll see there how access modifiers are used to generate the declarations in the file and where the import statements are defined.

This allows for two things, which we'll need before we can land this PR:
- Other renderer implementations that may come in the future will get consistent behaviour because we removed the layering violation.
- We can treat the imports used for the generated API differently from the `additionalImports`.

arpit-garg-1995:
Support access modifiers on generated imports

### Motivation

With the introduction of Swift 6.0's `InternalImportsByDefault` flag, generated declarations need a way to re-export the symbols they depend on. Supporting the configured access modifier on imports ensures they remain visible to consumers of the generated code when necessary.

### Modifications

- Add `accessModifier` property to `ImportDescription`.
- Update `TextBasedRenderer` to conditionally render `public` and `package` access modifiers on imports.
- Introduce `importDescriptions(adding:)` in `FileTranslator` to automatically propagate the configured access modifier to both built-in and additional imports.
- Adopt the new import generation logic in `ClientFileTranslator`, `ServerFileTranslator`, and `TypesFileTranslator`.
- Add unit tests in `Test_TextBasedRenderer` to verify the correct rendering of imports with different access modifiers, attributes, and OS conditions.

### Result

Generated Swift files can now properly emit `public import` or `package import` statements when compiled with Swift `6.0` and above, based on the provided configuration.

### Test Plan

Unit tests added in `Test_TextBasedRenderer`.
Testing with Act: Passed local medium act pull_request workflow

simonjbeaumont:
@arpit-garg-1995 just a heads up that this package only supports Swift 6.1+ now so you don't need to do anything special in the generator to handle older compilers.

arpit-garg-1995:
Addressed the latest feedback in this commit.

### Modifications
- Removed the compiler-version conditional around the generated code path.
- Dropped the special handling for older Swift compilers now that the package only supports Swift 6.1+.

### Result
This simplifies the generator logic and aligns it with the package’s current Swift version support.

### Test Plan
- Ran `swift test`.
- Testing with Act: Passed local medium act pull_request workflow

simonjbeaumont:
> * Ran `swift test`.
> * Testing with Act: Passed local medium act pull_request workflow

Are you sure you got the tests passing locally? They're failing for every Swift version and platform in CI.
 
> @swift-server-bot test this please

Just FYI — this does nothing.

arpit-garg-1995:
Let me check and confirm. Will update soon.

arpit-garg-1995:
Apologies for the earlier noise. The failures are now fixed:

- Added `setsAccessModifier: Bool = true` to `ImportDescription` and set it to `false` for `HTTPTypes`, since its types only appear in function bodies and Swift 6 rejects `public import` for implementation-only imports under `-warnings-as-errors`.
- Updated `importDescriptions(adding:)` to respect the flag, and refreshed the Petstore snapshots accordingly.
- Fixed `Examples/streaming-chatgpt-proxy/Sources/ChatGPT/Middleware.swift` to use `package import` for Foundation and OpenAPIRuntime to match the generator output for `package`-access targets.

Verified locally: format check passes, 319 tests pass with `-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error` on both macOS and Linux (Swift 6.2 Docker).

arpit-garg-1995:
@simonjbeaumont Hi there, I have updated the PR and thoroughly tested it with act, it passed soundness, macOS and linux jobs. Couldn't tested the windows one due to machine constraints. let me know your thoughts on this PR. Thanks!

simonjbeaumont:
Thanks. Just a heads up I'm out on vacation right now and will look at this when I get back.

simonjbeaumont (approved):

arpit-garg-1995:
Hi @simonjbeaumont , thanks for the approval! It looks like CI hasn't run on the latest commit yet, the workflow run needs maintainer approval before the required checks can pass and this can be merged. Could you approve the CI run when you get a chance? Happy to address anything that comes up. Thanks!

PR #898, opened 2026-04-20
Fix extended string delimiter selection in `renderLiteral`

kukushechkin:
Dynamically pick the minimum number of `#` delimiters so that neither the closing sequence nor the escape prefix appears in the content.

### Motivation

When the string content contains `"#` or `\#`, a single-hash raw literal (`#"..."#`) produces unexpected Swift code.

### Modifications

- Dynamically pick the minimum number of `#` delimiters so that neither the closing sequence nor the escape prefix appears in the content.
- Add test verifying correct handling of string delimiters.

### Result

No unexpected Swift code is generated.

### Test Plan

Unit tests.


FranzBusch (approved):

simonjbeaumont (approved):
Thanks @kukushechkin!

PR #900, opened 2026-04-24
Use decoded file system paths in plugin arguments

### LINKED ISSUE #899: Build plugin passes percent-encoded paths when project path contains spaces
### Description

When using `swift-openapi-generator` 1.11.1 through the build plugin, generation fails if the project directory contains a space in its path.

The plugin appears to pass `URL.path()` values as command-line arguments after the 1.11.1 migration from `PackagePlugin.Path` to `URL`. On recent Foundation, `URL.path()` returns a percent-encoded path, so a directory like:

```text
/Users/username/Documents/The Folder/Reproduceit
```

is passed to the generator as:

```text
/Users/username/Documents/The%20Folder/Reproduceit
```

The generator then tries to open that encoded string as a filesystem path and reports that the config file does not exist.

This works on 1.10.4 and fails on 1.11.1.

## Commit Hash

Likely related to the plugin migration in 73997cc62c2193d5046e431c9d546119dda14502.

## Context

The issue was encountered while building an Xcode project that uses the Swift OpenAPI Generator build plugin to generate code from an OpenAPI document and generator configuration file in the app target.

### Reproduction

Create or open a Swift package or Xcode project in a directory whose path contains a space, for example:

```text
/Users/username/Documents/The Folder/Reproduceit
```

Add `swift-openapi-generator` 1.11.1 as a plugin dependency and include the usual files in the target:

```text
Reproduceit/OpenAPI/openapi.yaml
Reproduceit/OpenAPI/openapi-generator-config.yaml
```

Build the target in Xcode, or otherwise run the build plugin.

The build fails during the plugin phase:

```text
PhaseScriptExecution Running swift-openapi-generator /Users/username/Library/Developer/Xcode/DerivedData/Reproduceit-epjtjkyhclqwledtfltpqtlyxzlr/Build/Intermediates.noindex/Reproduceit.build/Debug-iphoneos/Reproduceit.build/Script-7563506F2F9A6B7800C381F6.sh (in target 'Reproduceit' from project 'Reproduceit')
    cd /Users/username/Documents/The\ Folder/Reproduceit
    /bin/sh -c /Users/username/Library/Developer/Xcode/DerivedData/Reproduceit-epjtjkyhclqwledtfltpqtlyxzlr/Build/Intermediates.noindex/Reproduceit.build/Debug-iphoneos/Reproduceit.build/Script-7563506F2F9A6B7800C381F6.sh

Error: Configuration file not found at path: /Users/username/Documents/The%20Folder/Reproduceit/Reproduceit/OpenAPI/openapi-generator-config.yaml. Please ensure the file exists and the path is correct.
Usage: swift-openapi-generator <subcommand>
  See 'swift-openapi-generator --help' for more information.
Command PhaseScriptExecution failed with a nonzero exit code
```

The file exists at the decoded filesystem path:

```text
/Users/username/Documents/The Folder/Reproduceit/Reproduceit/OpenAPI/openapi-generator-config.yaml
```

### Package version(s)

```text
swift-openapi-generator: 1.11.1
```

The same project is safe on:

```text
swift-openapi-generator: 1.10.4
```

### Expected behavior

The build plugin should pass valid filesystem paths to the generator, so paths containing spaces should work the same as they did on 1.10.4.

Expected config path:

```text
/Users/username/Documents/The Folder/Reproduceit/Reproduceit/OpenAPI/openapi-generator-config.yaml
```

## Actual Behavior

The build plugin passes a percent-encoded path:

```text
/Users/username/Documents/The%20Folder/Reproduceit/Reproduceit/OpenAPI/openapi-generator-config.yaml
```

The generator treats that as a literal filesystem path and fails with "Configuration file not found".

### Environment

```console
$ swift --version
swift-driver version: 1.148.6 Apple Swift version 6.3.2 (swiftlang-6.3.2.1.103 clang-2100.1.1.101)
Target: arm64-apple-macosx26.0
```

```console
$ sw_vers
ProductName:		macOS
ProductVersion:		26.5
BuildVersion:		25F5058e
```

```console
$ uname -a
Darwin Users-MacBook-Pro.local 25.5.0 Darwin Kernel Version 25.5.0: Tue Apr 14 21:52:16 PDT 2026; root:xnu-12377.120.99.0.7~25/RELEASE_ARM64_T6041 arm64
```

### Additional information

A local fix is to use `URL.path(percentEncoded: false)` when constructing plugin command arguments, instead of `URL.path()`.

For example, in `Plugins/PluginsShared/PluginUtils.swift`, paths passed to the generator should be decoded filesystem paths:

```swift
doc.path(percentEncoded: false)
config.path(percentEncoded: false)
genSourcesDir.path(percentEncoded: false)
```

This avoids passing `%20` as part of the literal filesystem path.

shindgewongxj:
### Motivation

Fixes #899.

The 1.11.1 build plugin migrated from `PackagePlugin.Path` to `URL` when invoking `swift-openapi-
generator`. The plugin currently passes `URL.path()` values as command-line arguments to the generator.

On current Foundation, `URL.path()` returns a percent-encoded path. For projects located in directories
with spaces, the plugin passes paths such as:

 ```text
 /Users/username/The%20Folder/Reproduceit/Reproduceit/OpenAPI/openapi-generator-config.yaml
 ```

 The generator then treats that string as a literal filesystem path and fails to load the existing file.

### Modifications

Use decoded filesystem paths when constructing plugin arguments by calling path(percentEncoded: false).

This updates the paths passed for:

- the OpenAPI document
- the generator config file
- the plugin output directory

The same filesystem-path formatting is also used in duplicate-file plugin diagnostics.

### Result

Projects whose paths contain spaces, or other percent-encoded filesystem characters, can use the build
plugin successfully again.

This restores the behavior users had before the 1.11.1 plugin path migration.

### Test Plan

- Ran swift build
- Ran swift test --filter Test_GenerateOptions
- Manually verified Foundation path behavior:
  - URL.path() returns /tmp/The%20Folder/file.yaml
  - URL.path(percentEncoded: false) returns /tmp/The Folder/file.yaml


simonjbeaumont:
Thanks for taking the time to file the issue and provide a fix!

It's a good catch but think we should probably add a test to check we don't regress this again. I think the simplest way to do this would be to update the integration test so that one of the targets includes a space in its name.

I put up a branch for you to see how to do that: https://github.com/apple/swift-openapi-generator/compare/main...simonjbeaumont:swift-openapi-generator:sb/integration-test-for-targets-with-space.

Feel free to pull that commit into your branch. 

You can run the integration test locally using this command from the root of your checkout:

```console
% SWIFT_OPENAPI_GENERATOR_REPO_URL=file://$PWD ./scripts/run-integration-test.sh
```

(note: because it clones the current checkout to a temporary directory before running the test, it runs on changes that have been committed and does not include local changes in your Git worktree).

With just the patched integration test, it should reproduce the issue for you:

```console
% SWIFT_OPENAPI_GENERATOR_REPO_URL=file://$PWD ./scripts/run-integration-test.sh
...
Error: Configuration file not found at path: /private/var/folders/06/__qdhxzn7ld3_bsvfcfz0kch0000gn/T/run-integration-test.sh.JRcAp5Vyqn/swift-openapi-generator/IntegrationTest/Sources/Types%20Only/openapi-generator-config.yaml. Please ensure the file exists and the path is correct.
```

From there you can layer on top your commit for testing.

Are you willing to make these changes?

shindgew:
Thanks for the review and the ready-made patch branch!

I've cherry-picked your `sb/integration-test-for-targets-with-space` commit onto this branch (preserving authorship) and added a small follow-up commit to update the downstream module references — SwiftPM sanitizes the renamed target's C99 module name from `"Types Only"` to `Types_Only`, so `import Types` in `MockTransportClient`/`MockTransportServer` and `additionalImports: - Types` in the Client/Server generator configs needed updating for the integration test to compile.

Verified locally with `SWIFT_OPENAPI_GENERATOR_REPO_URL=file://$PWD ./scripts/run-integration-test.sh`:

- Reverting the plugin fix reproduces the failure you described:
  ```
  Error: Configuration file not found at path:
  .../Sources/Types%20Only/openapi-generator-config.yaml
  ```
- With the plugin fix applied, the script completes with `✅ Successfully built integration test package.`

Marking the PR as ready for review.

simonjbeaumont:
I realised we can simplify this further and reduce the changes required to the integration test by keeping the target name in the integration test package as `Types` but just put it in a directory with a space and set the `path:` parameter on the target declaration.

I've kicked our CI for you anyway.

shindgew:
Thanks for the follow-up. I updated the integration test to keep the target name as `Types` and instead point it at path: "Sources/Types Only", so the module name stays unchanged while the plugin still sees a filesystem path containing a space.

I also reverted the downstream Types_Only import/config changes that were only needed for the earlier approach, and kept the formatting fix for the plugin arguments.

simonjbeaumont (approved):
LGTM. Thanks for the contribution!

PR #904, opened 2026-05-11
Remove validations selectively

mattpolzin:
### Motivation

When it recently became important for the Swift OpenAPI Generator to omit exactly one builtin OpenAPIKit validation, it felt like high time OpenAPIKit added the ability to selectively remove default-on validations. This PR moves to a new OpenAPIKit version with that support and it uses the new method to reduce the amount of code the generator needs to maintain.

### Modifications

Swap out full enumeration of validations for selective removal.

### Result

This is just about reducing overhead in the codebase.

### Test Plan

A unit test has been added and all existing unit tests are passing.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -46,7 +46,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-collections", from: "1.1.4"),
 
         // Read OpenAPI documents
-        .package(url: "https://github.com/mattpolzin/OpenAPIKit", from: "6.0.0", traits: []),
+        .package(url: "https://github.com/mattpolzin/OpenAPIKit", from: "6.1.0", traits: []),
         .package(url: "https://github.com/jpsim/Yams", "4.0.0"..<"7.0.0"),
 
         // CLI Tool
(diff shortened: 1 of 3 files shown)


simonjbeaumont (approved):
LGTM—thank you!

Added a question for clarification.

mattpolzin:
> Added a question for clarification.

Not seeing this, is it maybe a draft of a review still?

simonjbeaumont:
> > Added a question for clarification.
> 
> Not seeing this, is it maybe a draft of a review still?

Weird. Not even in the draft -- looks like it's lost. I'll just ask here.

You added an equality test for the set of validations. Does this imply that OpenAPIKit won't add more default validations in a minor release? I kinda presumed that adding another validation might be in scope for a minor, but wanted to check as we wouldn't want a brittle test if so.

mattpolzin:
I previously have been adding new default-on validations in minor releases if the validation constitutes a definitive requirement of the spec, but now that I think about a downstream project wanting to allow OpenAPIKit updates without making their product stricter in unhelpful ways, I am going to avoid adding default-on validations as minor releases. They will be optional additions until the next major release where I will promote them to default-on. 

The idea with the test here is to know when a major OpenAPIKit release does add more validations and spot that ahead of time or to catch regressions that would reduce the number of validations being run somehow without intending to.

Issue #854, opened 2025-12-20
Crash on duplicate schema name with idiomatic naming strategy


marknefedov:
### Description

```
PS C:\Users\mark_\Desktop\kratos-client-swift> swift build
Building for debugging...
Swift OpenAPI Generator is running with the following configuration:
- OpenAPI document path: C:/Users/mark_/Desktop/kratos-client-swift/Sources/kratos-client-swift/openapi.json
- Configuration path: C:/Users/mark_/Desktop/kratos-client-swift/Sources/kratos-client-swift/openapi-generator-config.yaml
- Generator modes: types, client
- Access modifier: internal
- Naming strategy: idiomatic
- Name overrides: <none>
- Type overrides: <none>
- Feature flags: <none>
- Output file names: Types.swift, Client.swift
- Output directory: C:/Users/mark_/Desktop/kratos-client-swift/.build/plugins/outputs/kratos-client-swift/kratos-client-swift/destination/OpenAPIGenerator/GeneratedSources
- Diagnostics output path: <none - logs to stderr>
- Current directory: C:\Users\mark_\Desktop\kratos-client-swift
- Plugin source: build
- Is dry run: false
- Additional imports: <none>
- Additional file comments: <none>Swift/NativeDictionary.swift:792: Fatal error: Duplicate values for key: 'NullTime'
Current stack trace:
0    swiftCore.dll                      0x00007ffe6395bb40 swift_reportError + 94
1    swiftCore.dll                      0x00007ffe639fb620 swift_stdlib_reportFatalErrorInFile + 132
2    swiftCore.dll                      0x00007ffe63657ff0 StaticString.withUTF8Buffer<A>(_:) + 1060
3    swiftCore.dll                      0x00007ffe63657090 _assertionFailure(_:_:file:line:flags:) + 326
4    swiftCore.dll                      0x00007ffe639373e0 swift_dynamicCastForeignClassMetatypeUnconditional + 29378
5    swiftCore.dll                      0x00007ffe636a48f0 Dictionary.init<A>(uniqueKeysWithValues:) + 351
6    swift-openapi-generator-tool.exe   0x00007ff68648fbed <unavailable> + 13302765
7    swift-openapi-generator-tool.exe   0x00007ff6864f7a7f <unavailable> + 13728383
8    swift-openapi-generator-tool.exe   0x00007ff6864bbea4 <unavailable> + 13483684
[...]

PR #907, opened 2026-05-25
Emit a clear error instead of crashing on duplicate generated schema names

adityasingh2400:
When the idiomatic naming strategy maps two distinct OpenAPI schema names to the same generated Swift type name, the generator crashes instead of reporting the problem. This is the crash tracked in #854: a document with both `NullTime` and `nullTime` under `#/components/schemas` produces

```
Swift/NativeDictionary.swift: Fatal error: Duplicate values for key: 'NullTime'
```

Root cause is in `boxRecursiveTypes`, which builds a name-to-node lookup for recursion detection with `Dictionary(uniqueKeysWithValues:)`. That initializer traps on duplicate keys, and the idiomatic strategy can produce two declarations sharing a name even though the OpenAPI names were distinct. The defensive strategy keeps them distinct, which is why the crash only shows up with idiomatic naming.

The fix detects the collision before building the lookup and emits an error diagnostic that names the conflicting type and points at the documented workarounds (switch `namingStrategy` to `defensive`, or add a `nameOverrides` entry). The `Configuring-the-generator` article already notes that idiomatic naming "might produce name conflicts (in that case, switch back to defensive)", so this turns an undiagnosed trap into the guidance that already exists. The lookup construction now also tolerates duplicate keys, so the error diagnostic is the failure path rather than a fatal trap, independent of which diagnostic collector is in use.

Verification: running the generator with `namingStrategy: idiomatic` on a spec containing `NullTime` and `nullTime` previously crashed; it now exits non-zero with

```
error: Multiple schemas in '#/components/schemas' map to the same generated Swift type name 'NullTime', which is not supported. ...
```

I added a regression test in `Test_translateSchemas` that translates two schemas colliding under the idiomatic strategy and asserts the error is emitted. With the previous code that test aborts with the `Duplicate values for key` trap (signal 5); with the fix it passes. A second test confirms that distinct idiomatic names emit no error. The full `OpenAPIGeneratorCoreTests` target (115 tests) and the reference tests (114 tests) pass, so generated output for valid documents is unchanged, including recursive-type boxing.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateBoxedTypes.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateBoxedTypes.swift
@@ -25,7 +25,16 @@ extension TypesFileTranslator {
     func boxRecursiveTypes(_ decls: [Declaration]) throws -> [Declaration] {
 
         let nodes = decls.compactMap(DeclarationRecursionDetector.Node.init)
-        let nodeLookup = Dictionary(uniqueKeysWithValues: nodes.map { ($0.name, $0) })
+        if let duplicateName = nodes.firstDuplicateName {
+            try diagnostics.emit(
+                .error(
+                    message:
+                        "Multiple schemas in '#/components/schemas' map to the same generated Swift type name '\(duplicateName)', which is not supported. This usually happens when the naming strategy collapses two distinct OpenAPI names into one. Switch the namingStrategy to 'defensive', or add a 'nameOverrides' entry to give one of the schemas a different name.",
+                    context: ["name": duplicateName]
+                )
+            )
+        }
+        let nodeLookup = Dictionary(nodes.map { ($0.name, $0) }, uniquingKeysWith: { first, _ in first })
         let container = DeclarationRecursionDetector.Container(lookupMap: nodeLookup)
 
         let boxedNames = try RecursionDetector.computeBoxedTypes(rootNodes: nodes, container: container)
@@ -215,5 +224,16 @@ extension TypesFileTranslator {
         var desc = desc
         desc.isIndirect = true
         return desc
+    }
+}
+
+extension Array where Element == DeclarationRecursionDetector.Node {
+
+    /// The name of the first node whose name was already used by an earlier
+    /// node in the array, or `nil` if all node names are unique.
+    var firstDuplicateName: String? {
+        var seen = Set<String>()
+        for node in self where !seen.insert(node.name).inserted { return node.name }
+        return nil
     }
 }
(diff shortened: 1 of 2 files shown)


Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateBoxedTypes.swift
```diff
@@ -25,7 +25,16 @@ extension TypesFileTranslator {
...
+        if let duplicateName = nodes.firstDuplicateName {
+            try diagnostics.emit(
+                .error(
+                    message:
+                        "Multiple schemas in '#/components/schemas' map to the same generated Swift type name '\(duplicateName)', which is not supported. This usually happens when the naming strategy collapses two distinct OpenAPI names into one. Switch the namingStrategy to 'defensive', or add a 'nameOverrides' entry to give one of the schemas a different name.",
+                    context: ["name": duplicateName]
+                )
+            )
+        }
+        let nodeLookup = Dictionary(nodes.map { ($0.name, $0) }, uniquingKeysWith: { first, _ in first })
```
simonjbeaumont: I like the motivation of surfacing better errors to users but I don't love that we're now iterating over the nodes twice. Can we instead use the `uniquingKeysWith:` closure to capture the duplicates and then surface the diagnostic errors afterwards if duplicates are found.

Given the closure will be called for each duplicate this will also support providing a richer diagnostic when there is more than one duplicate.

simonjbeaumont (changes requested):
Thanks for contributing! I added a comment about a direction that would support the better diagnostics, without sacrificing the performance.

Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateBoxedTypes.swift
adityasingh2400: Good call, done in 6aef8fb. Duplicate detection now lives in the `uniquingKeysWith:` closure that builds the lookup, so we no longer scan the nodes a second time. The closure records each colliding name into a set, and afterward a single error reports all of them, so two distinct collisions surface together instead of stopping at the first. I added a test covering the multi-collision case alongside the existing single-collision one.

adityasingh2400:
Thanks, that is a much cleaner approach. I reworked it to build the lookup with `uniquingKeysWith:` and record any colliding names in that closure, then emit the diagnostics afterwards only when duplicates were found, so we no longer make a second pass over the nodes. The closure runs once per duplicate, so I also used that to give a richer message: a single collision names the one type, and multiple collisions list all of them with guidance to switch to the defensive naming strategy or add nameOverrides. Let me know if you would like the wording adjusted.

simonjbeaumont:
@adityasingh2400 did you mean to close this? Sorry for the delays in reviewing but I'll get to it soon.

adityasingh2400:
Sorry about that @simonjbeaumont, that was my mistake. I was bulk cleaning up stale forks on my profile and this got caught in the sweep, and since the fork was deleted I cannot reopen this one. I have restored the exact same branch including the uniquingKeysWith rework and opened #914 to replace it. No changes beyond what you see here, and no rush on the review. Thanks for picking it up.

PR #910, opened 2026-06-02
Stop boxing types that recurse only through an array or dictionary

### LINKED ISSUE #682: The generator is boxing properties of recursive types that do not need boxing
### Description

👋 I am observing that the generator is boxing properties that do not need boxing

### Reproduction

This OpenAPI spec:

```yaml
components:
  schemas:
    MyStruct:
      type: object
      properties:
        myProperty:
          type: array
          items:
            $ref: '#/components/schemas/MyStruct'
```

Generates boxed code:
```swift
internal struct MyStruct: Codable, Hashable, Sendable {
    /// - Remark: Generated from `#/components/schemas/MyStruct/myProperty`.
    internal var myProperty: [Components.Schemas.MyStruct]? {
        get  {
            storage.value.myProperty
        }
        _modify {
            yield &storage.value.myProperty
        }
    }
    /// Creates a new `MyStruct`.
    ///
    /// - Parameters:
    ///   - myProperty:
    internal init(myProperty: [Components.Schemas.MyStruct]? = nil) {
        storage = .init(value: .init(myProperty: myProperty))
    }
    internal enum CodingKeys: String, CodingKey {
        case myProperty
    }
    internal init(from decoder: any Decoder) throws {
        storage = try .init(from: decoder)
    }
    internal func encode(to encoder: any Encoder) throws {
        try storage.encode(to: encoder)
    }
    /// Internal reference storage to allow type recursion.
    private var storage: OpenAPIRuntime.CopyOnWriteBox<Storage>
    private struct Storage: Codable, Hashable, Sendable {
        /// - Remark: Generated from `#/components/schemas/MyStruct/myProperty`.
        var myProperty: [Components.Schemas.MyStruct]?
        init(myProperty: [Components.Schemas.MyStruct]? = nil) {
            self.myProperty = myProperty
        }
        typealias CodingKeys = Components.Schemas.MyStruct.CodingKeys
    }
}
```

### Package version(s)

```
├── swift-openapi-generator<https://github.com/apple/swift-openapi-generator@1.5.0>
│   ├── swift-algorithms<https://github.com/apple/swift-algorithms@1.2.0>
│   │   └── swift-numerics<https://github.com/apple/swift-numerics.git@1.0.2>
│   ├── swift-collections<https://github.com/apple/swift-collections@1.1.4>
│   ├── openapikit<https://github.com/mattpolzin/OpenAPIKit@3.3.0>
│   │   └── yams<https://github.com/jpsim/Yams@5.1.3>
│   ├── yams<https://github.com/jpsim/Yams@5.1.3>
│   └── swift-argument-parser<https://github.com/apple/swift-argument-parser@1.5.0>
├── swift-openapi-runtime<https://github.com/apple/swift-openapi-runtime@1.7.0>
│   └── swift-http-types<https://github.com/apple/swift-http-types@1.3.1>
├── swift-openapi-urlsession<https://github.com/apple/swift-openapi-urlsession@1.0.2>
│   ├── swift-openapi-runtime<https://github.com/apple/swift-openapi-runtime@1.7.0>
│   │   └── swift-http-types<https://github.com/apple/swift-http-types@1.3.1>
│   ├── swift-http-types<https://github.com/apple/swift-http-types@1.3.1>
│   └── swift-collections<https://github.com/apple/swift-collections@1.1.4>
└── swift-argument-parser<https://github.com/apple/swift-argument-parser@1.5.0>
```

### Expected behavior

This swift code is valid and compiles:
```swift
struct MYStruct {
    let myProperty: [MYStruct]
}
```

So I would expect that the generated code does not have any boxing.

### Environment

swift-driver version: 1.90.11.1 Apple Swift version 5.10 (swiftlang-5.10.0.13 clang-1500.3.9.4)
Target: arm64-apple-macosx14.0

### Additional information

Originally raised in https://github.com/apple/swift-openapi-generator/issues/70#issuecomment-2501931241

eyupcanakman:
### Motivation

The generator boxes a type whose only self-reference is reached through an array or dictionary. Swift arrays and dictionaries keep their elements in heap-allocated storage, so a reference reached only through them cannot form a value-type cycle and the type does not need a `CopyOnWriteBox`. `Supporting-recursive-types.md` already notes that arrays and dictionaries can be considered boxed, so the containing type does not require boxing. Fixes #682.

### Modifications

- In `ExistingTypeDescription.referencedSchemaComponentName`, return `nil` for `.array` and `.dictionaryValue` instead of recursing into them. `.optional` and `.any` still recurse.

### Result

A type that recurses only through an array or dictionary is generated as a plain struct with stored properties, not a boxed type.

### Test Plan

Added two snippet reference tests covering recursion through an array and through a dictionary.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/Recursion/DeclarationRecursionDetector.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/Recursion/DeclarationRecursionDetector.swift
@@ -155,13 +155,15 @@ fileprivate extension Array where Element == String {
 
 extension ExistingTypeDescription {
 
-    /// The name in the `Components.Schemas.` namespace, if the type can appear
-    /// there. Nil otherwise.
+    /// The name in the `Components.Schemas.` namespace this type references
+    /// without an intervening array or dictionary, if any. Nil otherwise.
     var referencedSchemaComponentName: String? {
         switch self {
         case .member(let components): return components.nameIfTopLevelSchemaComponent
-        case .array(let desc), .dictionaryValue(let desc), .any(let desc), .optional(let desc):
-            return desc.referencedSchemaComponentName
+        case .any(let desc), .optional(let desc): return desc.referencedSchemaComponentName
+        // Arrays and dictionaries provide heap indirection, so a reference reached
+        // only through them cannot form a value-type cycle that needs boxing.
+        case .array, .dictionaryValue: return nil
         case .generic: return nil
         }
     }
--- a/Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
+++ b/Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
@@ -1685,6 +1685,172 @@ final class SnippetBasedReferenceTests: XCTestCase {
         )
     }
 
+    func testComponentsSchemasRecursive_directAndThroughArray() throws {
+        try self.assertSchemasTranslation(
+            """
+            schemas:
+              Node:
+                type: object
+                properties:
+                  parent:
+                    $ref: '#/components/schemas/Node'
+                  children:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/Node'
+            """,
+            """
+            public enum Schemas {
+                public struct Node: Codable, Hashable, Sendable {
+                    public var parent: Components.Schemas.Node? {
+                        get  {
+                            self.storage.value.parent
+                        }
+                        _modify {
+                            yield &self.storage.value.parent
+                        }
+                    }
+                    public var children: [Components.Schemas.Node]? {
+                        get  {
+                            self.storage.value.children
+                        }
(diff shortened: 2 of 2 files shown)


simonjbeaumont (commented):
This looks good and is a really nice catch. Thanks!

Could we extend the tests to include the mixed case where the schema recurses through a direct ref AND a container.
  
```
  Node:
    type: object
    properties:
      parent: { $ref: '#/components/schemas/Node' }        # direct -> must still box
      children:
        type: array
        items: { $ref: '#/components/schemas/Node' }        # container -> must not prevent the box
```

This presumably should still box.

eyupcanakman:
Added two mixed-case tests, a `Node` with a direct `parent` ref alongside a `children` array, and the same with a `children` dictionary. Both still box, since the direct ref forms the value cycle and only the container path stopped contributing one.

simonjbeaumont (approved):
Thanks for the contribution!

PR #914, opened 2026-06-06
Emit a clear error instead of crashing on duplicate generated schema names

adityasingh2400:
When the idiomatic naming strategy maps two distinct OpenAPI schema names to the same generated Swift type name, the generator crashes instead of reporting the problem. This is the crash tracked in #854: a document with both `NullTime` and `nullTime` under `#/components/schemas` produces

```
Swift/NativeDictionary.swift: Fatal error: Duplicate values for key: 'NullTime'
```

Root cause is in `boxRecursiveTypes`, which builds a name-to-node lookup for recursion detection with `Dictionary(uniqueKeysWithValues:)`. That initializer traps on duplicate keys, and the idiomatic strategy can produce two declarations sharing a name even though the OpenAPI names were distinct. The defensive strategy keeps them distinct, which is why the crash only shows up with idiomatic naming.

The fix detects the collision before building the lookup and emits an error diagnostic that names the conflicting type and points at the documented workarounds (switch `namingStrategy` to `defensive`, or add a `nameOverrides` entry). The `Configuring-the-generator` article already notes that idiomatic naming "might produce name conflicts (in that case, switch back to defensive)", so this turns an undiagnosed trap into the guidance that already exists. The lookup construction now also tolerates duplicate keys, so the error diagnostic is the failure path rather than a fatal trap, independent of which diagnostic collector is in use.

Verification: running the generator with `namingStrategy: idiomatic` on a spec containing `NullTime` and `nullTime` previously crashed; it now exits non-zero with

```
error: Multiple schemas in '#/components/schemas' map to the same generated Swift type name 'NullTime', which is not supported. ...
```

I added a regression test in `Test_translateSchemas` that translates two schemas colliding under the idiomatic strategy and asserts the error is emitted. With the previous code that test aborts with the `Duplicate values for key` trap (signal 5); with the fix it passes. A second test confirms that distinct idiomatic names emit no error. The full `OpenAPIGeneratorCoreTests` target (115 tests) and the reference tests (114 tests) pass, so generated output for valid documents is unchanged, including recursive-type boxing.

---

Note: this supersedes #907, which I closed by accident during a bulk cleanup of my stale forks and cannot reopen because the original fork was deleted. The branch is identical to the reviewed state of #907 including the uniquingKeysWith rework discussed there.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateBoxedTypes.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateBoxedTypes.swift
@@ -25,7 +25,16 @@ extension TypesFileTranslator {
     func boxRecursiveTypes(_ decls: [Declaration]) throws -> [Declaration] {
 
         let nodes = decls.compactMap(DeclarationRecursionDetector.Node.init)
-        let nodeLookup = Dictionary(uniqueKeysWithValues: nodes.map { ($0.name, $0) })
+        if let duplicateName = nodes.firstDuplicateName {
+            try diagnostics.emit(
+                .error(
+                    message:
+                        "Multiple schemas in '#/components/schemas' map to the same generated Swift type name '\(duplicateName)', which is not supported. This usually happens when the naming strategy collapses two distinct OpenAPI names into one. Switch the namingStrategy to 'defensive', or add a 'nameOverrides' entry to give one of the schemas a different name.",
+                    context: ["name": duplicateName]
+                )
+            )
+        }
+        let nodeLookup = Dictionary(nodes.map { ($0.name, $0) }, uniquingKeysWith: { first, _ in first })
         let container = DeclarationRecursionDetector.Container(lookupMap: nodeLookup)
 
         let boxedNames = try RecursionDetector.computeBoxedTypes(rootNodes: nodes, container: container)
@@ -215,5 +224,16 @@ extension TypesFileTranslator {
         var desc = desc
         desc.isIndirect = true
         return desc
+    }
+}
+
+extension Array where Element == DeclarationRecursionDetector.Node {
+
+    /// The name of the first node whose name was already used by an earlier
+    /// node in the array, or `nil` if all node names are unique.
+    var firstDuplicateName: String? {
+        var seen = Set<String>()
+        for node in self where !seen.insert(node.name).inserted { return node.name }
+        return nil
     }
 }
--- a/Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateBoxedTypes.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateBoxedTypes.swift
@@ -25,31 +25,11 @@ extension TypesFileTranslator {
     func boxRecursiveTypes(_ decls: [Declaration]) throws -> [Declaration] {
 
         let nodes = decls.compactMap(DeclarationRecursionDetector.Node.init)
-        var duplicateNames: Set<String> = []
-        let nodeLookup = Dictionary(
-            nodes.map { ($0.name, $0) },
-            uniquingKeysWith: { existing, _ in
-                duplicateNames.insert(existing.name)
-                return existing
-            }
-        )
-        if !duplicateNames.isEmpty {
-            let sortedNames = duplicateNames.sorted()
-            let message: String
-            let context: [String: String]
-            if sortedNames.count == 1 {
-                let duplicateName = sortedNames[0]
-                message =
-                    "Multiple schemas in '#/components/schemas' map to the same generated Swift type name '\(duplicateName)', which is not supported. This usually happens when the naming strategy collapses two distinct OpenAPI names into one. Switch the namingStrategy to 'defensive', or add a 'nameOverrides' entry to give one of the schemas a different name."
-                context = ["name": duplicateName]
-            } else {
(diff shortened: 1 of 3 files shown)


Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateBoxedTypes.swift
```diff
@@ -25,7 +25,31 @@ extension TypesFileTranslator {
     func boxRecursiveTypes(_ decls: [Declaration]) throws -> [Declaration] {
 
         let nodes = decls.compactMap(DeclarationRecursionDetector.Node.init)
-        let nodeLookup = Dictionary(uniqueKeysWithValues: nodes.map { ($0.name, $0) })
+        var duplicateNames: Set<String> = []
```
simonjbeaumont: This PR as-is is a worthwhile, incremental improvement, so we can move forward with this approach. However, if you're motivated to improve it further in a follow up, we can enhance this to keep track of the schema location and even be able to emit diagnostics that show up in the editor, pointing to the duplicate entries in the OpenAPI document.

Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateBoxedTypes.swift
```diff
@@ -25,7 +25,31 @@ extension TypesFileTranslator {
     func boxRecursiveTypes(_ decls: [Declaration]) throws -> [Declaration] {
```
simonjbeaumont, line 25: Regardless of whether we enhance to provide a richer diagnostic including schema location, this is not the right place for this check. This function should do only what it says (boxing recursive types).

Further down in this file you'll find `translateSchemas(_:multipartSchemasNames:)`. This logic should move there either as a pass over the computed `[Declaration]` list, or as part of the `flatMap`.

We should also check to see if `TypeAssigner` is a logical place to put this. I can't remember how stateful that is, but if it is stateful, it seems like the best place to keep track of which types have been assigned and emit the diagnostics.

adityasingh2400:
Thanks for the review. I moved the duplicate-name check out of translateBoxedTypes and into translateSchemas, where it now runs as a pass over the computed declarations before boxing. translateBoxedTypes is back to only boxing recursive types, and it keeps the first node on a name collision so it cannot trap if a non-throwing diagnostic collector lets execution continue past the error. The clear error message is unchanged and the existing tests still pass. I left the richer schema-location diagnostic and the TypeAssigner idea as possible follow ups.

Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateBoxedTypes.swift
```diff
@@ -25,7 +25,11 @@ extension TypesFileTranslator {
...
 
         let nodes = decls.compactMap(DeclarationRecursionDetector.Node.init)
-        let nodeLookup = Dictionary(uniqueKeysWithValues: nodes.map { ($0.name, $0) })
+        // Duplicate generated names are detected and reported earlier, in
+        // `translateSchemas`. Keep the first node on a collision here so this
+        // step never traps even if a non-throwing diagnostic collector lets
+        // execution continue past that error.
```
simonjbeaumont, line 31:
```suggestion
```

Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/translateSchemas.swift
```diff
@@ -75,4 +76,33 @@ extension TypesFileTranslator {
...
+            let duplicateName = sortedNames[0]
+            message =
+                "Multiple schemas in '#/components/schemas' map to the same generated Swift type name '\(duplicateName)', which is not supported. This usually happens when the naming strategy collapses two distinct OpenAPI names into one. Switch the namingStrategy to 'defensive', or add a 'nameOverrides' entry to give one of the schemas a different name."
+            context = ["name": duplicateName]
+        } else {
+            let nameList = sortedNames.map { "'\($0)'" }.joined(separator: ", ")
+            message =
+                "Multiple schemas in '#/components/schemas' map to the same generated Swift type names \(nameList), which is not supported. This usually happens when the naming strategy collapses distinct OpenAPI names into one. Switch the namingStrategy to 'defensive', or add 'nameOverrides' entries to give the schemas different names."
+            context = ["names": nameList]
+        }
```
simonjbeaumont, line 105: Seems over the top to special case the single entry list. Let's just use one path.
```suggestion
        let nameList = sortedNames.map { "'\($0)'" }.joined(separator: ", ")
        message =
            "Multiple schemas in '#/components/schemas' map to the same generated Swift type names \(nameList), which is not supported. This usually happens when the naming strategy collapses distinct OpenAPI names into one. Switch the namingStrategy to 'defensive', or add 'nameOverrides' entries to give the schemas different names."
        context = ["names": nameList]
```

Tests/OpenAPIGeneratorCoreTests/Translator/TypesTranslator/Test_translateSchemas.swift
simonjbeaumont, line 1: Please add a test that shows that duplicate basenames in different namespaces do not produce an error. E.g. `A.T` and `B.T` are fine. The tests you added show only top-level types in the Schemas dictionary.

simonjbeaumont (commented):
Thanks. This is nearly ready to land. Just needs a couple more tests.

PR #925, opened 2026-07-14
Add namespace-based types file splitting

nac5504:
## Motivation

This is the first stage of a [recently proposed effort](https://github.com/apple/swift-openapi-generator/pull/866) to add output file splitting to the generator. Rather than landing the full dependency-sharding design all at once, this PR introduces a simple splitting strategy that is easy to reason about and yields substantial results already.

The generator now uses namespace-based splitting by default. Root declarations remain in `Types.swift`, `Components` and `Operations` are emitted in `Types+Components.swift` and `Types+Operations.swift`, and second-level component namespaces such as `Schemas`, `Parameters`, `RequestBodies`, `Responses`, and `Headers` are emitted in their own files. This keeps generated source files smaller while preserving the existing Swift namespace and API structure.

## Commit Outline (11)

_The commit messages contain more specific technical decisions that were made during this process._ 

#### 1. Add multi-file generator output plumbing ([a2013a8](https://github.com/apple/swift-openapi-generator/commit/a2013a83532350fb3b7da00de4e3f4924f903e39))

- Adds the multi-output generator path so one generation run can produce more than one Swift file.
- Initially kept existing single-file behavior intact while introducing the multi-output path; the final API shape is simplified in Review Patch 2 below.

#### 2. Add file splitting output config ([d2949a8](https://github.com/apple/swift-openapi-generator/commit/d2949a89b612bc8b2c7403a8f69e4f787add6451))

- Adds the `output.types.fileSplitting` configuration model and initial `namespace` strategy option.
- Exposes the opt-in through both YAML config and the direct CLI flag.

#### 3. Add namespace-based types file splitting ([eb817f7](https://github.com/apple/swift-openapi-generator/commit/eb817f73533a6264050ed4c5b9aa7ccf4f6f8c82))

- Emits `Types.swift`, `Types+Components.swift`, and `Types+Operations.swift` when namespace splitting is enabled.
- Preserves the root API surface while moving `Components` and `Operations` into their own generated files.

#### 4. [Review Patch 1] Simplify multi-file output model ([b002759](https://github.com/apple/swift-openapi-generator/commit/b002759c8a198806f9f9539d8bf54ee62c2cd1b3))

- Addresses initial review feedback by flattening rendered output to an array instead of adding a wrapper type.
- Removes brittle single-file conveniences now that structured output can contain multiple files.
- Removes empty namespace-specific config options and the dead output-file-name conditional in the tool write path.

#### 5. [Review Patch 2] Address multi-output generator API review feedback ([3979529](https://github.com/apple/swift-openapi-generator/commit/3979529e94ac6f3abbb4e05e3b87f7df57f31e35))

- Removes the single-file core generator entry point so `runGenerator` returns the pipeline final multi-file output directly.
- Updates the tool and tests to use the simplified multi-output API.
- Keeps the output-file-name composition helper internal and renames its variadic argument to avoid confusion with file extensions.

#### 6. [Review Patch 3] Make namespace types splitting the default ([791a86c](https://github.com/apple/swift-openapi-generator/commit/791a86cf2c0cf486355ff948a0b8244ff0447d95))

- Makes namespace-based type splitting unconditional, using a default depth of two.
- Emits `Types.swift`, `Types+Components.swift`, `Types+Operations.swift`, and the second-level component namespace files.
- Removes the proposed file-splitting configuration, updates focused tests and compatibility expectations, and splits the Petstore reference output into the new default file layout.

#### 7. [Review Patch 4] Fix Swift formatting ([c85e262](https://github.com/apple/swift-openapi-generator/commit/c85e262b8736d867081de015e8c4ac909563bf5c))

- Applies the repository’s Swift formatting rules to the multi-output implementation and its tests.
- Resolves the `Soundness / Format check` failure without changing generator behavior.

#### 8. [Review Patch 5] Preserve generated import access ([d84a5a1](https://github.com/apple/swift-openapi-generator/commit/d84a5a1227f801896094236e26c72c9384459a6f))

- Retains configured `public` or `package` access modifiers on imports in every namespace-split types file.
- Prevents public generated declarations from referencing internally imported types when `InternalImportsByDefault` is enabled.
- Updates focused coverage and regenerates the Petstore reference outputs to preserve the accepted proposal's source-compatibility guarantee.

#### 9. [Review Patch 6] Address additional review feedback ([791cb18](https://github.com/apple/swift-openapi-generator/commit/791cb18b81350f4acf17d1efa9a2a3f0b42655b7))

- Replaces order-dependent filename arrays and the filename-composition helper with the typed `OutputFileName` enum.
- Uses sets for filename membership while preserving canonical ordering at observable output boundaries.
- Renames the renderer factory parameter to `makeRenderer`.
- Clarifies generated-file stability in the documentation and removes temporary guidance.
- Removes redundant tests, strengthens multi-file assertions, and cleans up unused test scaffolding.

#### 10. [Review Patch 7] Fix DocC link capitalization ([7c04647](https://github.com/apple/swift-openapi-generator/commit/7c04647f0317677385829ff56af59ac8a3497ab5))

- Fixes the case-sensitive DocC link to the generator API-stability article.

#### 11. [Review Patch 8] Stop treating Swift warnings as errors in CI ([a75b940](https://github.com/apple/swift-openapi-generator/commit/a75b940f561373d6fd46b5e85707675dc3939c95))

- Removes `-Xswiftc -warnings-as-errors` from the Linux Swift 6.1, 6.2, and 6.3 unit-test jobs in the pull-request and main workflows.
- Keeps explicit target dependency import checking configured as an error.
- Leaves unused-import diagnostics visible as warnings while their proper resolution is handled separately.

## Usability

Namespace-based type splitting is enabled by default and does not require a YAML configuration option or CLI flag.

Generating types now automatically emits the applicable files from this set:

- `Types.swift`
- `Types+Components.swift`
- `Types+Operations.swift`
- `Types+Components+Schemas.swift`
- `Types+Components+Parameters.swift`
- `Types+Components+RequestBodies.swift`
- `Types+Components+Responses.swift`
- `Types+Components+Headers.swift`

Programmatic callers receive all generated files through the multi-output API:

```swift
let files = try runGenerator(input: input, config: config, diagnostics: diagnostics)
```

## Compatibility

Namespace splitting changes the generated source-file layout by default, but preserves the generated Swift namespace and API structure.

- Existing generator configurations require no new options.
- Integrations that assume type output is contained in a single `Types.swift` file must handle the additional generated files.
- Programmatic `runGenerator` callers receive the generator pipeline output array directly.
- The number and names of generated files are not considered stable API and may change in future releases.
- The SwiftPM and Xcode build-tool plugins declare the complete generated output set and support the default split layout.

## Future Direction

This PR intentionally limits splitting to a deterministic, depth-two namespace layout. The multi-output generator pipeline provides the foundation for future work such as deeper namespace splitting or dependency-aware file grouping without changing the generated Swift API structure.

This PR does not introduce a user-configurable splitting strategy. Any future configuration or alternative layout should be considered separately based on demonstrated use cases and performance results.

### Unit Tests

The following tests cover the final default depth-two namespace-splitting implementation:

- [`Test_Config.testOutputFileNameRawValues`](https://github.com/apple/swift-openapi-generator/blob/791cb18b81350f4acf17d1efa9a2a3f0b42655b7/Tests/OpenAPIGeneratorCoreTests/Test_Config.swift) verifies the hardcoded generated filenames.
- [`Test_Config.testGeneratorModeOutputFileNames`](https://github.com/apple/swift-openapi-generator/blob/791cb18b81350f4acf17d1efa9a2a3f0b42655b7/Tests/OpenAPIGeneratorCoreTests/Test_Config.swift) verifies mode-specific filename sets and the complete output catalog.
- [`Test_TypesFileTranslatorFileSplitting.testDefaultGenerationProducesDepth2NamespaceFiles`](https://github.com/apple/swift-openapi-generator/blob/791cb18b81350f4acf17d1efa9a2a3f0b42655b7/Tests/OpenAPIGeneratorCoreTests/Hooks/Test_TypesFileTranslatorFileSplitting.swift) verifies declaration routing, generated imports, and the complete namespace-split output.
- [`Test_GenerateOptions.testBuildPluginWritesEveryDeclaredOutputForUnrequestedModes`](https://github.com/apple/swift-openapi-generator/blob/791cb18b81350f4acf17d1efa9a2a3f0b42655b7/Tests/OpenAPIGeneratorTests/Test_GenerateOptions.swift) verifies build-plugin output declarations and empty placeholders for unrequested modes.
- [`SnippetBasedReferenceTests.testAdditionalFileComments`](https://github.com/apple/swift-openapi-generator/blob/791cb18b81350f4acf17d1efa9a2a3f0b42655b7/Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift) verifies additional comments are applied to every split types file.
- [The compatibility test multi-output harness](https://github.com/apple/swift-openapi-generator/blob/791cb18b81350f4acf17d1efa9a2a3f0b42655b7/Tests/OpenAPIGeneratorReferenceTests/CompatabilityTest.swift) verifies the expected output count and compiles the generated package.
- [`FileBasedReferenceTests.testPetstore`](https://github.com/apple/swift-openapi-generator/blob/791cb18b81350f4acf17d1efa9a2a3f0b42655b7/Tests/OpenAPIGeneratorReferenceTests/FileBasedReferenceTests.swift) verifies every generated Petstore output against its split reference file.

### Benchmark Testing

I ran an extensive benchmark suite against representative OpenAPI specs of different sizes and shapes, seen in the charts below. Each chart row compares the current single-file output against the new `namespace` splitting mode, over 5 attempts excluding outliers due to VM compute inconsistencies. Whiskers on the build and generator time charts show +/-1 standard deviation across the attempts. An observability layer was patched locally on top of the generator to observe durations spent in each stage, shown in the last chart.

Post-Generator Swift Build Times

These benefits come at little cost, as the generator time does not increase with statistical significance. Namespace splitting is very close to the single-file baseline across the suite, and is even slightly faster in several rows after outlier filtering.

Generator Time

The phase breakdown within the generator supports the same read. Namespace splitting adds some bookkeeping to route declarations into multiple output files and preserve imports, but that work is not large enough to materially move end-to-end generator time.

Generator Stage Breakdown

Overall, the benchmark story matches the design goal: namespace splitting produces a source layout that is friendlier to downstream Swift compilation without adding a meaningful generator-time cost or substantial added complexity.


Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
```diff
@@ -48,7 +48,7 @@ struct GeneratorPipeline {
     typealias TranslatedOutput = StructuredSwiftRepresentation
 
     /// An output of the rendering phase, usually written to disk.
-    typealias RenderedOutput = RenderedSwiftRepresentation
+    typealias RenderedOutput = RenderedSwiftOutputs
```
simonjbeaumont: I think this change is probably not necessary and the rest of this file can remain mostly unchanged if we just change the definition of the `RenderedSwiftRepresentation` type alias to be an array of files, instead of a single file.

Sources/_OpenAPIGeneratorCore/Layers/RenderedSwiftRepresentation.swift
```diff
@@ -20,7 +20,22 @@ public import struct Foundation.Data
 #endif
 
 /// An in-memory file that contains the generated Swift code.
-typealias RenderedSwiftRepresentation = InMemoryOutputFile
+public typealias RenderedSwiftRepresentation = InMemoryOutputFile
```
simonjbeaumont: The generator pipeline is intended to think in terms of phases from input to output. If the output of a phase is an array, that's fine, let's just change it rather than add another layer of abstraction. That probably means we can just change `RenderedSwiftRepresentation` to be `[InMemoryOutputFile]`.

simonjbeaumont: Separately, why has this now become `public`? And, more generally, there's a lot of `public` in here, which I suspect is unnecessary -- but maybe I'm missing some other purpose that's not used in this PR?

Sources/_OpenAPIGeneratorCore/Layers/StructuredSwiftRepresentation.swift
```diff
@@ -1087,8 +1087,25 @@ struct NamedFileDescription: Equatable, Codable {
...
+    var files: [NamedFileDescription]
+
+    /// The contents of the only file in the representation.
+    var file: NamedFileDescription {
+        get {
+            precondition(files.count == 1, "Expected a single structured Swift file, got \(files.count).")
+            return files[0]
+        }
+        set { files = [newValue] }
+    }
```
simonjbeaumont: It makes sense to me that `StructuredSwiftRepresentation` now wraps `files` instead of `file`. Do we need this additional computed `file` property, which seems like a brittle construction?

Sources/_OpenAPIGeneratorCore/OutputOptions.swift
```diff
@@ -0,0 +1,88 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// Configuration for generated output files.
+public struct OutputOptions: Sendable, Codable, Equatable {
+
+    /// Options that only affect `Types.swift` generation.
+    public var types: TypesOutputOptions?
```
simonjbeaumont: Do we foresee a need to have this level of distinction in the config? Are we anticipating output options that only affect types that we wouldn't want to also apply to the client and server outputs?

Sources/_OpenAPIGeneratorCore/Layers/StructuredSwiftRepresentation.swift
```diff
@@ -1087,8 +1087,25 @@ struct NamedFileDescription: Equatable, Codable {
 /// A file with contents made up of structured Swift code.
 struct StructuredSwiftRepresentation: Equatable, Codable {
 
-    /// The contents of the file.
-    var file: NamedFileDescription
+    /// The contents of the files.
+    var files: [NamedFileDescription]
+
+    /// The contents of the only file in the representation.
+    var file: NamedFileDescription {
```
nac5504: Since this type now supports multiple files, should we consider deprecating file entirely and make callers go through files? That would keep the model simpler: one stored shape, one access pattern, and no single-file convenience whose semantics can fail at runtime.

Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
```diff
@@ -99,7 +122,7 @@ func makeGeneratorPipeline(
     parser: any ParserProtocol = YamsParser(),
     validator: @escaping (ParsedOpenAPIRepresentation, Config) throws -> [Diagnostic] = validateDoc,
     translator: any TranslatorProtocol = MultiplexTranslator(),
-    renderer: any RendererProtocol = TextBasedRenderer.default,
+    renderer: @escaping () -> any RendererProtocol = { TextBasedRenderer.default },
```
nac5504: Uses a fresh renderer for each file because TextBasedRenderer owns a stateful StringCodeWriter and would accumulate rendered swift for multiple file outputs

Sources/_OpenAPIGeneratorCore/GeneratorMode.swift
```diff
@@ -39,15 +39,23 @@ extension GeneratorMode {
...
+        case .client: return Self.outputFileName("Client")
+        case .server: return Self.outputFileName("Server")
         }
     }
 
     /// The Swift file names for all supported generator mode values.
     public static var allOutputFileNames: [String] { GeneratorMode.allCases.map(\.outputFileName) }
 
+    /// Returns a Swift output file name composed from the provided name components.
+    public static func outputFileName(_ name: String, _ extensionNames: String...) -> String {
```
nac5504: Helpful to extract to create split output filenames, such as `Types+Components.swift`, `Types+Operations.swift`, or `Types+Slice1.swift`.

Sources/_OpenAPIGeneratorCore/OutputOptions.swift
```diff
@@ -0,0 +1,88 @@
...
+        self.namespace = namespace
+    }
+}
+
+/// Options for the namespace file splitting strategy.
+public struct NamespaceTypesFileSplittingOptions: Sendable, Codable, Equatable {
+
+    /// Creates namespace file splitting options.
+    public init() {}
+}
```
simonjbeaumont: This is a nothing right now.

Sources/swift-openapi-generator/runGenerator.swift
```diff
@@ -105,19 +104,19 @@ extension _Tool {
...
-            isDryRun: isDryRun
+        let outputs = try _OpenAPIGeneratorCore.runGeneratorOutputs(
+            input: .init(absolutePath: doc, contents: docData),
+            config: config,
+            diagnostics: diagnostics
         )
+        for output in outputs {
+            try replaceFileContents(
+                inDirectory: outputDirectory,
+                fileName: output.baseName == config.mode.outputFileName ? outputFileName : output.baseName,
```
simonjbeaumont: Does this condition ever do anything any more?

simonjbeaumont (commented):
Thanks for getting the ball rolling here. I'm broadly in support of us moving in this direction and have left some initial comments to get us going.

As a courtesy note: I'm about to head OOO for a couple of weeks. It's possible @czechboy0 will be able to keep the ball rolling but I cannot promise that.

Sources/_OpenAPIGeneratorCore/Renderer/RendererProtocol.swift
```diff
@@ -18,13 +18,12 @@
...
     /// - Parameters:
-    ///   - code: A structured representation of the Swift code.
+    ///   - file: A structured representation of the Swift file.
     ///   - config: The configuration of the generator.
     ///   - diagnostics: The collector to which to emit diagnostics.
     /// - Returns: A raw file with Swift contents.
     /// - Throws: An error if an issue occurs during rendering.
-    func render(structured code: StructuredSwiftRepresentation, config: Config, diagnostics: any DiagnosticCollector)
-        throws -> InMemoryOutputFile
+    func render(file: NamedFileDescription, config: Config, diagnostics: any DiagnosticCollector) throws -> InMemoryOutputFile
```
nac5504: now renders one file at a time, instead of `StructuredSwiftRepresentation` which contains multiple files now

Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
nac5504: done!

Sources/_OpenAPIGeneratorCore/OutputOptions.swift
nac5504: Deleted this for the scope of this PR. The vision is that in the future once more complicated splitting configs are added, there will be `SplittingOptions` knobs that will be passed. This stub was for consistency.

Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
```diff
@@ -83,7 +83,30 @@ struct GeneratorPipeline {
 /// - Returns: The raw contents of the generated Swift file.
 public func runGenerator(input: InMemoryInputFile, config: Config, diagnostics: any DiagnosticCollector) throws
     -> InMemoryOutputFile
-{ try makeGeneratorPipeline(config: config, diagnostics: diagnostics).run(input) }
+{
```
nac5504: Would you also be in favor of removing the existing runGenerator endpoint which returns one file, and instead just have the multi-output version? The reason I did this was to preserve existing callsites of current users.

Sources/_OpenAPIGeneratorCore/OutputOptions.swift
nac5504: I do think splitting client/server could be useful later for very large specs, but I see it as a separate strategy rather than this namespace strategy. Client/Server outputs are mostly flat operation methods that reference Operations types, while `Types.swift` owns the large generated declaration graph. So for this first slice I scoped the config under output.types to reflect the strategy we actually support today. I could also foresee some other type of output option in the future that doesn't fall within the realm of output file splitting for types.

nac5504:
@czechboy0 Would love to get your thoughts on this first proposal also! Happy to talk through design decisions, as I will be putting substantial time towards this effort over the next month. Thank you!

Sources/_OpenAPIGeneratorCore/GeneratorMode.swift
simonjbeaumont: Fine to have. Does it need to be public? And the use of the term "extension" in the name is confusing when considering file names. IIUC this is the _Swift extension_ (e.g. `Components`) and the file extension will always be `.swift`. Open to ideas on how you make that clearer.

Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
simonjbeaumont: Yes, I'd expect this to be the case. The pipeline runs as a series of transformations over inputs and outputs. It's possible for one stage to fan out or in, and the output of the whole pipeline should, therefore, be the output type of the final stage.

simonjbeaumont (approved):
I think this is really starting to take shape. Thanks for the hard work @nac5504!

I'd like to give @czechboy0 a chance to look at this before we land it, but I've marked it as approved meaning approved-in-principle, notwithstanding any feedback he may have.

nac5504:
Thanks for the support @simonjbeaumont! Just addressed some of your comments, let me know if you see anything else worth cleaning up :) 

Additionally, am I expected to update documentation for this contribution?

simonjbeaumont:
Back from leave -- thanks for your patience. 

> Additionally, am I expected to update documentation for this contribution?

Yes -- take a look at https://github.com/apple/swift-openapi-generator/blob/main/Sources/swift-openapi-generator/Documentation.docc/Articles/Configuring-the-generator.md

nac5504:
Thank you @simonjbeaumont @czechboy0, this PR should be ready for further review following the proposal passing.

Sources/_OpenAPIGeneratorCore/GeneratorMode.swift
```diff
@@ -39,14 +39,38 @@ extension GeneratorMode {
...
+        switch self {
+        case .types:
+            return [
+                outputFileName, Self.outputFileName(outputFileName, "Components"),
+                Self.outputFileName(outputFileName, "Operations"),
+                Self.outputFileName(outputFileName, "Components", "Schemas"),
+                Self.outputFileName(outputFileName, "Components", "Parameters"),
+                Self.outputFileName(outputFileName, "Components", "RequestBodies"),
+                Self.outputFileName(outputFileName, "Components", "Responses"),
+                Self.outputFileName(outputFileName, "Components", "Headers"),
```
czechboy0: This is a good start, we might expand this with eg "pathItems" as well in the future, full list https://swagger.io/specification/v3.2/#components-object

Sources/_OpenAPIGeneratorCore/GeneratorMode.swift
```diff
@@ -39,14 +39,38 @@ extension GeneratorMode {
...
-        case .client: return "Client.swift"
-        case .server: return "Server.swift"
+        case .types: return Self.outputFileName("Types")
+        case .client: return Self.outputFileName("Client")
+        case .server: return Self.outputFileName("Server")
+        }
+    }
+
+    /// The Swift file names emitted for this generator mode.
+    public var outputFileNames: [String] {
```
czechboy0: I don't like how we rely on the order of files in the array here. Could you create an enum, something like:

```swift
enum OutputFileName: String, Hashable {
  case client = "Client.swift"
  case server = "Server.swift"
  case types = "Types.swift"
  case typesComponents = "Types+Components.swift"
  ...
}
```

And then this declaration would return something like:

```swift
public var outputFileNames: Set<OutputFileName> {
  // ...
}
```

This way, in https://github.com/apple/swift-openapi-generator/pull/925/changes#diff-9d028e92fec1344c33903ea7548594680ad2922b1d587abf174d4dd007a4476eR91 we'll be able to stop relying on the exact order of files in the array.

Sources/_OpenAPIGeneratorCore/Translator/TypesTranslator/TypesFileTranslator.swift
```diff
@@ -43,20 +43,66 @@ struct TypesFileTranslator: FileTranslator {
...
+                            )
+                        )
+                    ]
+                )
+            )
+        }
+        return StructuredSwiftRepresentation(
+            files: [
+                .init(
+                    name: fileNames[0],
```
czechboy0: Let's design out the reliance on the order in the array, more in https://github.com/apple/swift-openapi-generator/pull/925/changes#r3802652538

Sources/_OpenAPIGeneratorCore/GeneratorMode.swift
```diff
@@ -39,14 +39,38 @@ extension GeneratorMode {
...
+        case .client, .server: return [outputFileName]
         }
     }
 
     /// The Swift file names for all supported generator mode values.
-    public static var allOutputFileNames: [String] { GeneratorMode.allCases.map(\.outputFileName) }
+    public static var allOutputFileNames: [String] { GeneratorMode.allCases.flatMap(\.outputFileNames) }
+
+    /// Returns a Swift output file name composed from the base name and type-name suffixes.
+    static func outputFileName(_ baseName: String, _ suffixes: String...) -> String {
```
czechboy0: Eh, I think it's better if we just hardcode these as the raw values in an enum, more in https://github.com/apple/swift-openapi-generator/pull/925/changes#r3802652538

Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
```diff
@@ -99,7 +99,7 @@ func makeGeneratorPipeline(
     parser: any ParserProtocol = YamsParser(),
     validator: @escaping (ParsedOpenAPIRepresentation, Config) throws -> [Diagnostic] = validateDoc,
     translator: any TranslatorProtocol = MultiplexTranslator(),
-    renderer: any RendererProtocol = TextBasedRenderer.default,
+    renderer: @escaping () -> any RendererProtocol = { TextBasedRenderer.default },
```
czechboy0:
```suggestion
    makeRenderer: @escaping () -> any RendererProtocol = { TextBasedRenderer.default },
```

Nit, now that it's a factory closure, let's rename the parameter.

Sources/_OpenAPIGeneratorCore/GeneratorPipeline.swift
```diff
@@ -128,7 +128,11 @@ func makeGeneratorPipeline(
         ),
         renderSwiftFilesStage: .init(
             preTransitionHooks: [],
-            transition: { input in try renderer.render(structured: input, config: config, diagnostics: diagnostics) },
+            transition: { input in
+                try input.files.map { namedFile in
+                    try renderer().render(namedFile: namedFile, config: config, diagnostics: diagnostics)
```
czechboy0:
```suggestion
                    try makeRenderer().render(namedFile: namedFile, config: config, diagnostics: diagnostics)
```

Sources/swift-openapi-generator/Documentation.docc/Articles/Configuring-the-generator.md
```diff
@@ -110,6 +110,21 @@ additionalFileComments:
...
+- `Types+Operations.swift`
+- `Types+Components+Schemas.swift`
+- `Types+Components+Parameters.swift`
+- `Types+Components+RequestBodies.swift`
+- `Types+Components+Responses.swift`
+- `Types+Components+Headers.swift`
+
+The file layout does not change generated Swift symbol names. For example, schema types remain nested under
+`Components.Schemas`. When invoking the generator directly or checking generated sources into a repository, retain all
+of the emitted files.
```
czechboy0: This is a temporary comment that won't make sense to new users, so let's just remove it.

Sources/swift-openapi-generator/Documentation.docc/Articles/Configuring-the-generator.md
```diff
@@ -110,6 +110,21 @@ additionalFileComments:
...
+
+- `Types.swift`
+- `Types+Components.swift`
+- `Types+Operations.swift`
+- `Types+Components+Schemas.swift`
+- `Types+Components+Parameters.swift`
+- `Types+Components+RequestBodies.swift`
+- `Types+Components+Responses.swift`
+- `Types+Components+Headers.swift`
+
```
czechboy0, line 123:
```suggestion

> Important: The number and names of generated files are _not_ considered to be stable, and can change at any time. For details, check out <doc:api-stability-of-the-generator>.
```

Sources/swift-openapi-generator/GenerateOptions.swift
```diff
@@ -48,6 +48,7 @@ struct _GenerateOptions: ParsableArguments {
     @Option(
         help: "When specified, writes out the diagnostics into a YAML file instead of emitting them to standard error."
     ) var diagnosticsOutputPath: URL?
+
```
czechboy0: Revert?

Tests/OpenAPIGeneratorCoreTests/Hooks/Test_TypesFileTranslatorFileSplitting.swift
```diff
@@ -0,0 +1,172 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftOpenAPIGenerator open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftOpenAPIGenerator project authors
```
czechboy0:
```suggestion
// Copyright (c) 2026 Apple Inc. and the SwiftOpenAPIGenerator project authors
```

Tests/OpenAPIGeneratorCoreTests/Test_GeneratorPipeline.swift
```diff
@@ -0,0 +1,66 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftOpenAPIGenerator open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftOpenAPIGenerator project authors
```
czechboy0:
```suggestion
// Copyright (c) 2026 Apple Inc. and the SwiftOpenAPIGenerator project authors
```

Tests/OpenAPIGeneratorCoreTests/Test_GeneratorPipeline.swift
czechboy0, line 1: I don't think we need these two tests, feel free to delete them.

Tests/OpenAPIGeneratorReferenceTests/SnippetBasedReferenceTests.swift
```diff
@@ -6462,7 +6462,7 @@ final class SnippetBasedReferenceTests: XCTestCase {
         let document = try YAMLDecoder().decode(OpenAPI.Document.self, from: documentYAML)
         let translation = try translator.translateFile(parsedOpenAPI: document)
         try XCTAssertSwiftEquivalent(
-            XCTUnwrap(translation.file.contents.topComment),
+            XCTUnwrap(translation.files.first?.contents.topComment),
```
czechboy0: Can you add a test assertion that `files.count == 1` as well, to ensure this `.first` isn't hiding a bug?

Tests/OpenAPIGeneratorTests/Test_GenerateOptions.swift
```diff
@@ -141,5 +179,14 @@ final class Test_GenerateOptions: XCTestCase {
             )
         } catch { XCTFail("Expected ArgumentParser.ValidationError, but got: \(type(of: error)) - \(error)") }
     }
+
     #endif
+
+    private func makeTemporaryConfig(_ contents: String) throws -> URL {
```
czechboy0: Could you make this function take a closure and clean up the files both on the success and failure paths at the end? To avoid the test files sitting in /tmp indefinitely.

czechboy0 (changes requested):
Looks great overall, @nac5504 - thank you! I'm really liking the large Types.swift file being split up.

Added a few minor comments, once those are addressed I'm happy with this landing.

nac5504:
Thank you so much for the additional feedback @czechboy0! I just addressed everything and amended with a 9th commit, could you take another look? Thanks!

czechboy0 (approved):

czechboy0:
Thank you, @nac5504 👏🙏

czechboy0:
Some of the CI is failing, please take a look @nac5504

nac5504:
@czechboy0 Any advice on the preferred way to handle the unused public-import warnings here? Split generated files preserve the same public imports, but some are unused in individual files and fail Linux CI under -warnings-as-errors. Should we downgrade UnusedImportAccess in CI, or calculate import visibility per generated file?

czechboy0:
Hmm let's remove warnings as errors here and file a separate issue. Fixing this properly could be non-trivial.

nac5504:
@czechboy0 added a 10th commit for a potential fix, curious what you think

czechboy0:
Let's break this out into a separate PR. It's a quite risky change that I'd like to consider in isolation. In case we need to roll it back, I'd prefer for us not to have to roll back the file splitting too.

nac5504:
That makes sense @czechboy0. Just reverted the 10th commit to only fix the documentation CI error. Would you like me to create a separate PR to remove warnings as errors first?

> let's remove warnings as errors here

czechboy0:
No you can change it in this PR, we want to get the CI to green here.

nac5504:
@czechboy0 understood. can you run CI again? It should be resolved

Issue #551, opened 2024-03-25
A way to group features or services to allow middleware to be applied to groups


dannflor:
### Motivation

Using the OpenAPI generator loses a lot of expressivity on the server side. Once you're in generated-code-land, you lose access to the functionality you would be able to have with vanilla routes e.g. applying middleware to groups. 
```swift
/// Before
final class AdminMiddleware: AsyncMiddleware
{
    func respond(to request: Request, chainingTo next: AsyncResponder) async throws -> Response {
        if request.auth.has(User.self) {
            if try await (UserService(request.db).user(from: request.auth).isAdmin {
                return try await next.respond(to: request)
            }
        }
        throw Abort(.forbidden, reason: "Admins only, fools")
    }
}
let adminProtected = app.grouped(AdminMiddleware())
adminProtected.group("api", configure: apiController)
func apiController(api: RoutesBuilder) {
    api.get("users") { req async throws -> [User] in
        try await UserService(req.db).allUsers
    }
    api.delete("user", ":name") { req async throws -> User in
        if let username = req.parameters.get("name") {
            try await UserService(req.db).deleteUser(username)
        }
    }
}

/// After
struct APIProtocolImpl: APIProtocol {
    // I am using a middleware with swift-dependency to have access to the request
    @Dependency(\.request) var req
    
    func getUsers(_ input: Operations.getUsers.Input) async throws -> Operations.getUsers.Output {
        // I would need to copy this block for every single route that would be behind AdminMiddleware
        guard try await req.auth.has(User.self), (UserService(req.db).user(from: req.auth).isAdmin else {
            return  .forbidden(.init("Admins only, fools"))
        }
        return try await .ok(.init(body: .json(
            .init(UserService(req.db).allUsers)
        )))
    }
}
```
Repeating the middleware logic for every single route that should be behind it can quickly grow completely untenable if you have middleware that is supposed to be in front of dozens or hundreds of endpoints.
However you *can* apply middleware to the routesBuilder you pass to the transport:
```
let transport = VaporTransport(
    routesBuilder: app.grouped(OpenAPIRequestInjectionMiddleware()).grouped(AdminMiddleware())
)
```
But we're artificially limited to one document, which means I can only apply a middleware if I'm willing to put it in front of *all* routes. I've heard that this project intends to implement authorization as part of the API spec, which would alleviate the problem in this specific example, but this isn't the only situation I would want to arbitrarily group routes for a middleware.

### Proposed solution

[...]

Issue #37, opened 2023-05-25
Support for security scheme/auth


czechboy0:
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#security-scheme-object

Let's do a proposal first, as there are a few ways this could be implemented.

For now, the workaround is to write your own middleware that injects the appropriate headers to every outgoing request.

Example: https://github.com/apple/swift-openapi-generator/tree/main/Examples/auth-client-middleware-example

Jonovono:
@czechboy0 I'm able to get the client middleware to inject headers to the request, but whats the best way to extract that on the server?

What I have tried so far is having a server middleware that verifies the token:

```
    func intercept(
        _ request: Request,
        metadata: ServerRequestMetadata,
        operationID: String,
        next: (Request, ServerRequestMetadata) async throws -> Response
    ) async throws -> Response {
        if protectedOperations.contains(operationID) {
            if let authToken = request.headerFields.first(where: { $0.name == "Authorization" })?.value
            {

                if validateJWTToken(authToken) {
                    var newHeaderFields = request.headerFields
                    newHeaderFields.append(HeaderField(name: "uid", value: "the-user-id"))

                    let newRequest = Request(
                        path: request.path,
                        query: request.query,
                        method: request.method,
                        headerFields: newHeaderFields,
                        body: request.body
                    )
                    let response = try await next(newRequest, metadata)
                    return response
```

All that works great, but from the auth jwt token I extract a uid, i'd like to pass that along to the subsequent handlers. How could I achieve this?

In the code above, I tried adding it to the headers, but then in my handler I don't seem to see it:

```
    func authorizeUser(_ input: Operations.authorizeUser.Input) async throws -> Operations.authorizeUser.Output {
        print("INPUT", input.headers)
        return .ok(.init())
    }

```

Any ideas? I might be doing it completely wrong

czechboy0:
@Jonovono you could add the value to a TaskLocal value on the server, which allows you to access it from the handler. Or inject a token storage object both into the middleware and the object implementing APIProtocol on the server.

Jonovono:
Thanks @czechboy0 , I have went the TaskLocal route and seems to be working good!

mapedd:
Right now, for default handlers generated by OpenAPI for Vapor. there's no way to access request object to authorize it. Is there any solution that being worked on?

czechboy0:
[...]

PR #927, opened 2026-07-16
Emit per-operation security metadata behind a feature flag (SOAR-0015)

vdsingh:
Implements **SOAR-0015** (proposal included in this PR) and addresses #37.

> **Depends on** apple/swift-openapi-runtime#203, which adds the `SecurityRequirement` value types this generated code references. That PR should land (and be tagged in a release) first.

### Motivation

The generator discards the OpenAPI `security` / `securitySchemes` sections entirely, so a `ClientMiddleware` — which only receives an `operationID` — cannot tell which operations require authentication, or which scheme. Adopters must either apply credentials unconditionally or hand-maintain a list that drifts from the document (#37, and the per-operation targeting gap in #551).

### Modifications

Adds a `securityMetadata` feature flag. When enabled, `TypesFileTranslator` emits, per operation, the security requirements it currently drops:

```swift
public enum Operations {
    public enum getGreeting {
        public static let id: String = "getGreeting"
        public static let securityRequirements: [OpenAPIRuntime.SecurityRequirement] = [
            .init(schemes: [.init(name: "OAuth2", kind: .oauth2, scopes: ["read"])])
        ]
    }
}
public enum OperationSecurity {
    public static let requirementsByOperationID: [String: [OpenAPIRuntime.SecurityRequirement]] = ...
}
```

Semantics mirror OpenAPI: outer array = OR of alternatives, each requirement's `schemes` = AND-group, `[]` = explicitly public; an operation without `security` inherits the document-level requirements. AND-groups are sorted by scheme name for deterministic output. Unsupported cases (`mutualTLS`, non-`bearer`/`basic` HTTP) raise a diagnostic. With the flag off, generated output is byte-identical.

Also includes the SOAR-0015 proposal, the `Proposals.md` entry, and the corresponding checkboxes in the supported-features article.

### Result

Adopters can enable `securityMetadata` and drive a conditional-authentication middleware from `OperationSecurity.requirementsByOperationID` — applying credentials only where required and failing fast when a required one is missing.

### Test Plan

Adds reference tests in `SnippetBasedReferenceTests` for inherited document-level security, operation-level AND/OR groups (apiKey + http-bearer + oauth2), explicit `security: []`, the `OperationSecurity` namespace, and that nothing is emitted when the flag is off. Full suite passes (330 tests). Also dogfooded end-to-end against a real 130-operation document, generating and compiling the metadata (114 authenticated / 16 public) into an app.

cc @czechboy0 @simonjbeaumont - flagging per the proposal process.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/FeatureFlags.swift
+++ b/Sources/_OpenAPIGeneratorCore/FeatureFlags.swift
@@ -26,8 +26,14 @@
 /// 0.2 is tagged. (This is for pre-1.0 versioning, would be 1.0 and 2.0 after
 /// 1.0 is released.)
 public enum FeatureFlag: String, Hashable, Codable, CaseIterable, Sendable {
-    // needs to be here for the enum to compile
-    case empty
+    /// Emit per-operation OpenAPI security requirements as generated metadata.
+    ///
+    /// When enabled, each `Operations.<Name>` namespace gains a
+    /// `securityRequirements` property, and a top-level `OperationSecurity`
+    /// namespace exposes a `requirementsByOperationID` map, both typed as
+    /// `OpenAPIRuntime.SecurityRequirement`. A client middleware can consume
+    /// these to apply authentication only to the operations that require it.
+    case securityMetadata
 }
 
 /// A set of enabled feature flags.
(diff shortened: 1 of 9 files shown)


czechboy0:
Hi @vdsingh,

If you're still interested in working on this issue, let's discuss next steps in #37.

Thanks! 🙏

PR #928, opened 2026-08-03
Support nullable referenced schemas in generated types.

### LINKED ISSUE #906: Support nullable referenced schemas using anyOf with type: null
### Description

When an OpenAPI 3.1 schema uses `anyOf` with a `$ref` and a `type: "null"` branch, Swift OpenAPI Generator does not appear to generate the expected optional Swift property.

Minimal schema:

```yaml
openapi: 3.1.0
info:
  title: Repro
  version: 1.0.0
paths: {}
components:
  schemas:
    Parent:
      type: object
      properties:
        child:
          anyOf:
            - $ref: '#/components/schemas/Child'
            - type: 'null'
      required:
        - child
    Child:
      type: object
      properties:
        name:
          type: string
      required:
        - name
```

### Expected behavior

The generated Swift type for `Parent` should allow `child` to be decoded/encoded as a nullable referenced object, for example as an optional `Child`-like generated property.

### Actual behavior

The generator did not produce a usable optional property for this schema shape in my project. I had to switch to this workaround:

```yaml
child:
  allOf:
    - $ref: '#/components/schemas/Child'
  nullable: true
```

That workaround generates a usable optional Swift property, but it also emits warnings because `nullable` is not part of the OpenAPI 3.1 schema vocabulary:

```text
Found 'nullable' property. This property is not supported by OpenAPI v3.1.0.
OpenAPIKit has translated it into 'type: ["null", ...]'.
```

### Why this matters

OpenAPI 3.1 supports JSON Schema's `null` type, so `anyOf`/`oneOf` with a `type: "null"` branch is a common way to express a nullable `$ref`. It would be useful if Swift OpenAPI Generator supported this directly, or documented the recommended schema shape for nullable referenced objects.

### Environment

- Swift OpenAPI Generator: 1.12.0
- OpenAPI document version: 3.1.0
- Swift toolchain: Xcode toolchain, Swift 6.x

femivita:
### Motivation

_[OpenAPI 3.1 supports nullable referenced schemas using anyOf with a $ref and a null branch. The generator previously treated this pattern as unsupported, preventing valid schemas from generating the expected optional Swift property.]_

### Modifications

_[

- Added support for detecting anyOf: [$ref, null] schemas.
- Updated type matching to resolve these schemas as optional referenced types.
- Updated optionality detection.
- Updated schema validation to validate the referenced branch while ignoring the null branch.
- Added regression tests for type matching, type assignment, and schema validation.
- Resolves [Issue #906](https://github.com/apple/swift-openapi-generator/issues/906).

]_

### Result

_[Nullable referenced schemas now generate optional Swift types correctly.
For example:
```
yaml
child:
  anyOf:
    - $ref: "#/components/schemas/Child"
    - type: "null"
```
 
generates:
```
swift
var child: Components.Schemas.Child?
Other anyOf schemas retain their existing behavior.
```
]_

### Test Plan

_[

- Added regression tests covering nullable reference type matching and assignment.
- Added schema validation coverage for nullable references.
- Ran the full test suite with swift test.
- Result: 120 tests passed.
- Ran swift-format lint on all modified files.
- Ran git diff --check successfully.

]_

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeMatcher.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeMatcher.swift
@@ -69,6 +69,9 @@ struct TypeMatcher {
             for: schema.value,
             test: { (schema) -> TypeUsage? in
                 if let builtinType = _tryMatchBuiltinNonRecursive(for: schema) { return builtinType }
+                if case let .any(of: schemas, _) = schema, let reference = nullableReference(in: schemas) {
+                    return try TypeAssigner(context: context).typeName(for: reference).asUsage.asOptional
+                }
                 guard case let .reference(ref, _) = schema else { return nil }
                 return try TypeAssigner(context: context).typeName(for: ref).asUsage
             },
@@ -94,6 +97,7 @@ struct TypeMatcher {
             for: schema.value,
             test: { schema in
                 if _tryMatchBuiltinNonRecursive(for: schema) != nil { return true }
+                if case let .any(of: schemas, _) = schema, nullableReference(in: schemas) != nil { return true }
                 guard case .reference = schema else { return false }
                 return true
             },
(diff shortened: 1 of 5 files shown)


czechboy0:
Hi @femivita,

Thanks! 🙏

Issue #929, opened 2026-08-03
Track proposal for namespace-based generated types file splitting


nac5504:
### Description

Track the public proposal and review discussion for opt-in namespace-based file splitting of generated `Types.swift` output.

Related PRs:
- #925 implements the initial namespace-based splitting support (approved).
- #866 contains related prior discussion and proposal context.

The intended proposal scope is limited to the implementation in #925:

- Add an opt-in `namespace` file-splitting strategy for generated types.
- Keep default generation unchanged: one `Types.swift` file.
- When enabled, emit:
  - `Types.swift` for root declarations
  - `Types+Components.swift` for the generated `Components` namespace
  - `Types+Operations.swift` for the generated `Operations` namespace
- Keep generated adopter-facing API names unchanged.
- Leave more advanced sharding strategies, such as dependency-aware or count-based splitting, to future proposals if benchmarks justify the added complexity.

PR #930, opened 2026-08-03
SOAR-0015: Add namespace file splitting proposal

nac5504:
## Summary

- Adds SOAR-0015 for default namespace-based splitting of generated `Types.swift` output.
- Links the proposal to issue #929 and implementation PR #925.
- Adds SOAR-0015 to the proposals index.

## Notes

The proposal is scoped to the namespace split from PR #925: `Types.swift`, `Types+Components.swift`, and `Types+Operations.swift`, `Types+Components+Schemas.swift`, etc. More advanced sharding remains listed only as future direction.

### DIFF
--- a/Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0015.md
+++ b/Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0015.md
@@ -110,14 +110,13 @@ Supporting build-tool plugin integration is outside this proposal. That follow-u
 
 ### API stability
 
-Existing users are unaffected unless they explicitly opt in:
+Existing users are unaffected unless they explicitly opt in. Without `output.types.fileSplitting`, types generation continues to produce the same single `Types.swift` output.
 
-- Existing configs continue to generate the same single `Types.swift` file by default.
-- Generated adopter-facing symbol names are not intended to change.
-- Runtime public API, runtime "Generated" SPI, transports, and middleware are not intended to change.
-- The new config file and CLI surface are generator API additions.
-- Programmatic generator callers now work with the generator pipeline's rendered output files.
-- Existing configured output file names continue to apply to the mode's primary output file.
+The main observable change for opted-in users is the generated output shape: a types generation run can produce multiple files instead of one. Direct CLI users and custom build integrations that enable this option need to write or compile every generated output file, not just the primary `Types.swift` file. Programmatic generator callers likewise need to handle the returned output-file collection.
+
+This does not intentionally change generated adopter-facing Swift symbols. The same generated declarations remain available to client and server code, but some declarations now live in `Types+Components.swift` and `Types+Operations.swift` instead of the primary file.
+
+Runtime public API, runtime "Generated" SPI, transports, and middleware are not intended to change. The new config file and CLI surface are generator API additions, and existing configured output file names continue to apply to the primary types output file.
 
 ### Future directions
 
--- a/Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0015.md
+++ b/Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0015.md
@@ -1,0 +1,141 @@
+# SOAR-0015: Namespace-Based Types File Splitting
+
+Add opt-in namespace-based splitting for generated `Types.swift` output.
+
+## Overview
+
+- Proposal: SOAR-0015
+- Author(s): [Nick Candello](https://github.com/nac5504)
+- Status: **Awaiting Review**
+- Issue: [apple/swift-openapi-generator#929](https://github.com/apple/swift-openapi-generator/issues/929)
+- Implementation:
+    - [apple/swift-openapi-generator#925](https://github.com/apple/swift-openapi-generator/pull/925)
+- Affected components:
+    - generator
+- Related links:
+    - Prior discussion: [apple/swift-openapi-generator#866](https://github.com/apple/swift-openapi-generator/pull/866)
+
+### Introduction
+
+Introduces opt-in file splitting for generated `Types.swift` file. Generator stage now outputs 3 types files: root declarations remain in `Types.swift`, the generated `Components` namespace moves to `Types+Components.swift`, and the generated `Operations` namespace moves to `Types+Operations.swift`.
+
+### Motivation
+
+Large generated `Types.swift` files can dominate downstream Swift compilation time for sizeable OpenAPI specs. In [apple/swift-openapi-generator#866](https://github.com/apple/swift-openapi-generator/pull/866), various methods of splitting generated schema declarations were explored, and all of them demonstrated real compile-time improvements to differing degrees.
+
+The benchmark results from PR #925 show meaningful build-time wins across the board for API specs of varying sizes. In the post-generator Swift build measurements, Stripe's API improved from 4m49s to 2m13s, GitHub REST improved from 9m12s to 5m54s, and Jira improved from 59.5s to 37.7s. The benchmark suite also did not show a meaningful generator-time tradeoff.
+
+### Proposed solution
+
+Add a types output file-splitting configuration and support one strategy: `namespace`. This keeps the public option explicit while leaving the default output unchanged.
+
+Namespace-based splitting has a few useful properties:
+
+- Existing generation behavior remains unchanged unless users opt in.
+- The generated Swift symbol names remain the same; only the generated file layout changes.
+- The split follows namespaces that already exist in generated code, making the output predictable and easy to review.
+
+Users can enable the strategy in configuration:
+
+```yaml
+generate:
+  - types
+output:
+  types:
+    fileSplitting:
+      strategy: namespace
+```
+
+They can also enable the same strategy from the command line:
+
+```sh
+swift-openapi-generator generate openapi.yaml \
+  --mode types \
+  --types-file-splitting namespace
+```
+
+With this config enabled, the generator emits:
+
+- `Types.swift` for root declarations.
+- `Types+Components.swift` for the generated `Components` namespace.
+- `Types+Operations.swift` for the generated `Operations` namespace.
+
+#### Benchmark results
+
+The benchmark suite in PR #925 compares current single-file output against the new `namespace` splitting mode across representative OpenAPI specs. Each row uses five attempts with outliers excluded for VM compute inconsistencies; build and generator time charts use whiskers for +/- one standard deviation.
+
+The main benchmark result is that post-generator Swift build time improves for several large specs, while generator time remains close to the single-file baseline. The generator-stage breakdown suggests declaration routing is not a meaningful generator-time cost.
+
+![Post-generator Swift build times](https://gist.githubusercontent.com/nac5504/9452377b295faeca1d1198dd6d8ac07d/raw/e4580d68735097dec960548db3db3ca3da7ea833/cli-initiated-compile-build-times.svg)
+
+![Generator time](https://gist.githubusercontent.com/nac5504/9452377b295faeca1d1198dd6d8ac07d/raw/e40ef57a323e6e792b4964af647f3bfe63178e92/generator-time.svg)
+
+![Generator stage breakdown](https://gist.githubusercontent.com/nac5504/9452377b295faeca1d1198dd6d8ac07d/raw/c2da7a2272d8c71a1f8305da3c0dc940f474a7f6/generator-stage-breakdown.svg)
+
+If the proposal renderer does not embed external images, the benchmark chart sources are:
+
+- [Post-generator Swift build times](https://gist.githubusercontent.com/nac5504/9452377b295faeca1d1198dd6d8ac07d/raw/e4580d68735097dec960548db3db3ca3da7ea833/cli-initiated-compile-build-times.svg)
+- [Generator time](https://gist.githubusercontent.com/nac5504/9452377b295faeca1d1198dd6d8ac07d/raw/e40ef57a323e6e792b4964af647f3bfe63178e92/generator-time.svg)
+- [Generator stage breakdown](https://gist.githubusercontent.com/nac5504/9452377b295faeca1d1198dd6d8ac07d/raw/c2da7a2272d8c71a1f8305da3c0dc940f474a7f6/generator-stage-breakdown.svg)
+
+### Detailed design
+
+The implementation adds the configuration and pipeline plumbing needed for a single generator run to emit more than one Swift file. The default behavior remains unchanged: without an explicit types file-splitting configuration, types generation still emits one `Types.swift` file.
+
+The new configuration lives under generated output options:
+
+- `output.types.fileSplitting.strategy`, currently supporting `namespace`.
+
+The same model is used by YAML configuration, direct command-line invocation, and programmatic callers that construct `_OpenAPIGeneratorCore.Config` directly.
+
+At the pipeline boundary, `StructuredSwiftRepresentation` now contains an array of named structured Swift files. Rendering maps over that array, rendering each `NamedFileDescription` independently into an `InMemoryOutputFile`. The public `runGenerator` entry point returns all generated output files from the pipeline:
+
+```swift
+let files = try runGenerator(input: input, config: config, diagnostics: diagnostics)
+```
+
+For the `namespace` strategy, `TypesFileTranslator` routes declarations before rendering rather than parsing rendered Swift source after the fact:
+
+- `Types.swift` contains the root declarations, including `APIProtocol`, the API protocol extension, and server declarations.
+- `Types+Components.swift` contains the generated `Components` namespace.
+- `Types+Operations.swift` contains the generated `Operations` namespace.
+
+Output file names are planned from the splitting configuration using the configured primary types file name as the base. Existing configured output file names therefore continue to apply to the primary types output, while the two split files use the same `+Namespace` suffix convention.
+
+#### Build-tool plugin boundary
+
+SwiftPM and Xcode build-tool plugins must declare generated output files before invoking the generator executable. PR #925 therefore rejects build-tool plugin invocations when `output.types.fileSplitting` is configured.
+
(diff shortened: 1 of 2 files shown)


Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0015.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Detailed design
+
+The implementation adds the configuration and pipeline plumbing needed for a single generator run to emit more than one Swift file. The default behavior remains unchanged: without an explicit types file-splitting configuration, types generation still emits one `Types.swift` file.
+
+The new configuration lives under generated output options:
+
+- `output.types.fileSplitting.strategy`, currently supporting `namespace`.
+
+The same model is used by YAML configuration, direct command-line invocation, and programmatic callers that construct `_OpenAPIGeneratorCore.Config` directly.
```
czechboy0: Please remove these implementation details. The design details should be about the conceptual approach and user-visible concepts.

There are many possible implementations, so no need to lock that in the proposal, in fact it can distract from where users should provide feedback - on the API and config file, not on the implementation.

Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0015.md
```diff
@@ -0,0 +1,140 @@
...
+
+- `Types.swift` contains the root declarations, including `APIProtocol`, the API protocol extension, and server declarations.
+- `Types+Components.swift` contains the generated `Components` namespace.
+- `Types+Operations.swift` contains the generated `Operations` namespace.
+
+Output file names are planned from the splitting configuration using the configured primary types file name as the base. Existing configured output file names therefore continue to apply to the primary types output, while the two split files use the same `+Namespace` suffix convention.
+
+#### Build-tool plugin boundary
+
+SwiftPM and Xcode build-tool plugins must declare generated output files before invoking the generator executable. PR #925 therefore rejects build-tool plugin invocations when `output.types.fileSplitting` is configured.
```
czechboy0: Hmm this is unfortunate - we should discuss more what our options are.

Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0015.md
```diff
@@ -0,0 +1,140 @@
...
+swift-openapi-generator generate openapi.yaml \
+  --mode types \
+  --types-file-splitting namespace
+```
+
+With this config enabled, the generator emits:
+
+- `Types.swift` for root declarations.
+- `Types+Components.swift` for the generated `Components` namespace.
+- `Types+Operations.swift` for the generated `Operations` namespace.
```
czechboy0: I wonder if we should go one level deeper - one file for `#/components/schemas`, one file for `#/components/parameters`, etc.

Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0015.md
czechboy0: Not being able to support this in the build plugin might actually be an argument for switching this to be the default and making this not even configurable. What do folks think?

Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0015.md
nac5504: This is exactly the follow-up PR I am drafting. It provides a config option to specify if you want the first-level namespaces split or the second-level.

czechboy0: We should discuss this more - whether and how much can actually be configurable, considering the limitation on build plugins.

https://github.com/apple/swift-openapi-generator/pull/930#discussion_r3706666275

simonjbeaumont: What speedup do we get from the next level. My original PoC centred around each section in components being split into its own extension, e.g. `Components+Schemas.swift` and `Components+Parameters.swift` and then further sharding these decl-wise, e.g. `Components+Schemas.{0..N}.swift`.

If we think there's a meaningful gain from the next level in the general case I think we just do it unilaterally. People are already opting into sharding and the fewest levers they have to pull to get maximum benefit the better. So I'd be inclined to have these phases:

1. Shard based on namespace, but to the next level of granularity.
2. Further shard these based on number of decls.
3. Something more advanced.

(1) I think we could defend with few options. (2) we need to discuss as we probably need to move to a prebuild plugin.

Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0015.md
simonjbeaumont: I think we should be open to moving to a prebuild plugin if it unlocks significant benefits. If we need a dynamic number of files generated then that's what prebuild plugins are advertised as the solution for.

Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0015.md
czechboy0: I like 1 at the next level - out of the box, no configuration. I'm hopeful that can bring a lot of benefits already without any user action (for build plugin users; users generating manually will just need to check in a few more files).

nac5504: I ran a test suite of first-level vs. [second-level](https://github.com/nac5504/swift-openapi-generator/pull/6) splitting against baseline, and I didn't see any real improvements past the current proposal. Here are the plots:
image
image
image

The depth 2 setup outputs the following files:
```
Types.swift
Types+Components.swift
Types+Components+Headers.swift
Types+Components+Parameters.swift
Types+Components+RequestBodies.swift
Types+Components+Responses.swift
Types+Components+Schemas.swift
Types+Operations.swift
```
These outputs are typically dominated in size by `Types+Components+Schemas.swift`, which is why I believe the further splitting has no effect. The decision for which depth should be the default splitting config should likely be based on readability of the output files rather than build time performance.

czechboy0: In earlier tests you also had the GitHub OpenAPI doc - can you test that one as well?

> These outputs are typically dominated in size by Types+Components+Schemas.swift, which is why I believe the further splitting has no effect.

This is only true for OpenAPI docs that _don't_ have much content in `#/components/parameters`, `#/components/requestBodies`, and `#/components/responses`. But I've definitely worked with OpenAPI docs that have quite a lot of content there, and I believe GitHub has some as well.

I'm quite surprised by the error bars showing that this change made performance _worse_ for the Stripe API, in debug builds (where the files are compiled in parallel). How could that be? Is that just measurement noise?

nac5504: Here is the full suite including Github's doc:
image
image
image
image

> I'm quite surprised by the error bars showing that this change made performance worse for the Stripe API, in debug builds

I'm not sure which graph you're referring to here? The generator time or the swift compile time? Both splitting configs were statistically the same, the slightly longer build time for depth 2 splitting is just noise from the testing setup. In general, there could be some measurement noise since these runs exist on MacOS VMs and don't have full CPU usage, but results are averaged from 5 trials with outliers removed so it should be accurate.

nac5504: > This is only true for OpenAPI docs that don't have much content in #/components/parameters, #/components/requestBodies, and #/components/responses. But I've definitely worked with OpenAPI docs that have quite a lot of content there, and I believe GitHub has some as well.

Looking at the size breakdown of Github's API spec, and it is still very heavily favored towards the schemas:

| File | LOC | Size |
| -- | -- | -- |
|Types+Components+Headers.swift | 30 | 1.5 KiB|
|Types+Components+Parameters.swift | 608 | 35.5 KiB|
|Types+Components+RequestBodies.swift | 15 | 557 B|
|Types+Components+Responses.swift | 717 | 33.9 KiB|
|Types+Components+Schemas.swift | 323,319 | 20.8 MiB|
|Types+Components.swift | 13 | 496 B|
|Types+Operations.swift | 174,576 | 8.8 MiB|
|Types.swift | 23,993 | 1.8 MiB|
|Total | 523,271 | 31.4 MiB|

I'd imagine a spec with lots of headers, parameters, request bodies, responses, would certainly benefit from this depth 2 sharding. I'd be in favor of making this the default config to benefit those specs also.

nac5504: @simonjbeaumont @czechboy0 Just to clarify before I rewrite the proposal and the implementation PR, I will update to:

- Remove the splitting config entirely, and have the depth-2 splitting as the default
- Add a pre-build plugin to support SwiftPM build plugin users

Is there anything else worth changing?

simonjbeaumont: > * Remove the splitting config entirely, and have the depth-2 splitting as the default

Yes.

> * Add a pre-build plugin to support SwiftPM build plugin users

No. But call it out as a potential requirement in the future directions if we were to adopt further, non-determnistic sharding.

> Is there anything else worth changing?

Nah, once you've done the above I think we can kick off the forums review and have some folks chime in there.

simonjbeaumont (commented):
@nac5504 thanks for your work on this so far 🙏 

I've kicked off the forums thread[^1]. 

Please update the status to `In Review` and include a link to the thread in the metadata.

Please also subscribe to the forums thread so you can engage with any feedback that might come in.

[^1]: https://forums.swift.org/t/proposal-soar-0015-namespace-based-types-file-splittingt/88830

Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0015.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Introduction
+
+This proposal splits generated `Types.swift` into multiple output files at the existing generated namespace boundaries to improve post-generator Swift build times.
+
+### Motivation
+
+Large generated `Types.swift` files can dominate downstream Swift compilation time for sizeable OpenAPI documents. Currently, adopters cannot ask the generator to produce smaller types files; working around the issue requires post-processing generated source, which is brittle and unsupported. In [apple/swift-openapi-generator#866](https://github.com/apple/swift-openapi-generator/pull/866), several ways of splitting generated declarations demonstrated compile-time improvements to differing degrees.
+
+The benchmark results from PR #925 show meaningful post-generator Swift build-time improvements for several large API documents. Stripe's API improved from 4m49s to 2m13s, GitHub REST improved from 9m12s to 5m54s, and Jira improved from 59.5s to 37.7s. The benchmark suite did not show a meaningful generator-time tradeoff.
```
czechboy0: Missing link for #925

czechboy0: > The benchmark suite did not show a meaningful generator-time tradeoff.

How should I interpret this? That the benchmark didn't show a generator _slowdown_? Or something else?

Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0015.md
```diff
@@ -0,0 +1,109 @@
...
+
+Generated namespaces provide stable, understandable splitting boundaries without introducing arbitrary declaration counts or dependency-grouping rules. Initially, this proposal introduced a more trivial verson of namespace splitting at the first namespace depth (i.e. `Types.swift`, `Types+Operations.swift`, `Types+Components.swift`), however enabling second-layer namespace splitting enables docs with heavy diversity within `Components` to see substantial benefits. Further proposals may split generated components even more in efforts to reduce default compile-time for package users.
```
czechboy0:
```suggestion
Generated namespaces provide stable, understandable splitting boundaries without introducing arbitrary declaration counts or dependency-grouping rules. Initially, this proposal introduced a more trivial verson of namespace splitting at the first namespace depth (i.e. `Types.swift`, `Types+Operations.swift`, `Types+Components.swift`), however enabling second-layer namespace splitting enables docs with heavy diversity within `Components` to see substantial benefits.
```

I'd remove the sentence from here and only mention it in Future Directions.

Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0015.md
```diff
@@ -0,0 +1,109 @@
...
+- `Types+Components+Headers.swift`
+
+Declarations are assigned to files according to the generated namespace that owns them:
+
+- Root declarations remain in `Types.swift`.
+- The `Components` namespace container is emitted in `Types+Components.swift`.
+- The `Operations` namespace is emitted in `Types+Operations.swift`.
+- The supported second-level `Components` namespaces are each emitted in their corresponding `Types+Components+<Namespace>.swift` file.
+
+All files are compiled into the same module, so adopters continue to use the same generated symbols and namespace hierarchy. Only the source-file layout changes. A configured primary types filename remains the base name for the output set, with namespace names appended using the `+Namespace` convention. For example, a primary filename of `GeneratedTypes.swift` produces `GeneratedTypes+Components.swift` and the corresponding namespace-derived files.
```
czechboy0: > All files are compiled into the same module

That's not really up to the generator - it simply emits Swift files. Adopters who use manual invocation of the CLI can in theory then shift the files around various modules, making this not fully accurate.

czechboy0: > A configured primary types filename remains the base name for the output set, with namespace names appended using the `+Namespace` convention. For example, a primary filename of `GeneratedTypes.swift` produces `GeneratedTypes+Components.swift` and the corresponding namespace-derived files.

What is this referring to? Is this proposal also proposing the ability to _configure_ the base name of the generated `Types.swift`? We don't have that ability today.

Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0015.md
```diff
@@ -0,0 +1,109 @@
...
+
+![Generator stage breakdown](https://gist.githubusercontent.com/nac5504/9452377b295faeca1d1198dd6d8ac07d/raw/c2da7a2272d8c71a1f8305da3c0dc940f474a7f6/generator-stage-breakdown.svg)
+
+### Detailed design
+
+The generator's types-output path changes from producing one source file to producing a named collection of source files. The file layout mirrors the generated Swift namespace hierarchy. A declaration belongs to the file for its nearest supported namespace boundary, while declarations outside those boundaries remain in the primary types file.
+
+Each generated file is a source-level fragment of the same generated API. Moving a declaration between these files does not rename it, change its access level, or introduce an additional Swift namespace.
+
+Types generation now produces a collection of output files. Programmatic callers of `runGenerator` must consume the full collection rather than assuming that types generation produces only one file:
```
czechboy0: Please remove any mention of `runGenerator`, it's an implementation detail, not public API, and users never interact with it.

Moving most of the contents from Proposed solution to Detailed design will help here.

Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0015.md
```diff
@@ -0,0 +1,109 @@
...
+
+### API stability
+
+This proposal does not change the runtime public API, runtime "Generated" SPI, or the APIs implemented by transports and middleware.
+
+Existing and newly generated adopter code remains source compatible: symbol names, namespace nesting, and intended access are unchanged. Only the source-file layout changes, so integrations that assume a single `Types.swift` file must instead write, track, or compile the complete output set.
+
+The generator API does change: `runGenerator` returns `[InMemoryOutputFile]` instead of one `InMemoryOutputFile`, requiring programmatic callers to handle a collection. Existing configuration files and CLI arguments remain valid, but types generation now emits multiple files by default. The command and build-tool plugins automatically handle all generated files without additional configuration. Custom names for Types.swift are also applied to the derived filenames.
```
czechboy0:
```suggestion
Existing configuration files and CLI arguments remain valid, but types generation now emits multiple files by default. The command and build-tool plugins automatically handle all generated files without additional configuration.
```

Let's avoid discussing internal APIs.

czechboy0: > Custom names for Types.swift are also applied to the derived filenames.

Can you elaborate on this again? Which feature does this refer to?

Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0015.md
nac5504: My apologies, I was conflating our repo's post-generator pipeline with the generator's internal setup. I'll remove all mentions of that, thank you for the feedback!

czechboy0:
Great, thanks @nac5504 - I added my +1 on the forums thread.

nac5504:
Here is the next feature I would like to land after this proposal concludes: https://github.com/nac5504/swift-openapi-generator/pull/8. Any early feedback would be greatly appreciated!

Sources/swift-openapi-generator/Documentation.docc/Proposals/SOAR-0015.md
```diff
@@ -0,0 +1,100 @@
+# SOAR-0015: Namespace-Based Types File Splitting
+
+Split generated types into deterministic, depth-two namespace files by default.
+
+## Overview
+
+- Proposal: SOAR-0015
+- Author(s): [Nick Candello](https://github.com/nac5504)
+- Status: **In Review**
```
czechboy0:
```suggestion
- Status: **Ready for Implementation**
```

czechboy0:
Hi all,

(temporarily taking over as release manager due to Si's OOO).

Thank you everyone for the feedback - it was broadly supportive, so this proposal is now **Ready for Implementation**.

czechboy0 (approved):

czechboy0:
Thanks again @nac5504 - the proposal was accepted. I'll get this PR merged, and please tag @simonjbeaumont and me in the implementation PR once it's ready and fully aligned with the accepted proposal.

PR #931, opened 2026-08-05
Make nullable schemas referenced with $ref generate optional types

k-kohey:
### Motivation

When a response body schema is referenced with `$ref`, the generated type is not optional even if the referenced schema is nullable.

```yaml
components:
  schemas:
    MyObject:
      type: [object, null]
      properties:
        id:
          type: string
      required:
        - id
  responses:
    MyResponse:
      description: OK
      content:
        application/json:
          schema:
            $ref: '#/components/schemas/MyObject'
```

Current output:

```swift
@frozen public enum Body: Sendable, Hashable {
    case json(Components.Schemas.MyObject)
}
```

Expected output:

```swift
@frozen public enum Body: Sendable, Hashable {
    case json(Components.Schemas.MyObject?)
}
```

Writing the same schema inline already produces `...?`, so the behavior differs depending on whether `$ref` is used. This also contradicts the documented rule in `Handling-nullable-schemas.md`: "The nullability of a schema is propagated through references."

### Modifications

Removed the `.reference` branch in `TypeAssigner.typeUsage(usingNamingHint:withSchema:components:inParent:)`, which resolved the type name without applying nullability. Every schema now goes through `_typeUsage`, which already applies `withOptional(isOptional(...))` to references, so inline and referenced schemas share a single path.

Request bodies are unaffected, since `translateRequestBody` explicitly drops optionality with `withOptional(false)`.

### Result

Nullable schemas referenced with `$ref` are generated as optional types. This covers both `type: [object, null]` and `nullable: true`, as well as transitive references through another schema. The output for non-nullable references is unchanged.

### Test Plan

Added tests to `SnippetBasedReferenceTests`.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeAssigner.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/TypeAssignment/TypeAssigner.swift
@@ -96,7 +96,9 @@ struct TypeAssigner {
             // it:
             let unboxedSchema = schema.flattenToJsonSchema()
             switch unboxedSchema.value {
-            case let .reference(reference, _): associatedType = try typeName(for: reference).asUsage
+            case let .reference(reference, _):
+                associatedType = try typeName(for: reference).asUsage
+                    .withOptional(try TypeMatcher(context: context).isOptional(unboxedSchema, components: components))
             default:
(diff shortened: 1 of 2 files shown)


czechboy0:
Hi @k-kohey,

Thanks! 🙏

Issue #817, opened 2025-08-27
codegen fails to produce anyOf/oneOf with a null type


BrentMifsud:
### Description

The code generator skips `anyOf` or `oneOf` that have nullable listed:

```
warning: Schema "null" is not supported, reason: "schema type", skipping [context: foundIn=Operations.OperationName.Output.Ok.Body.JsonPayload.DataPayloadPayload (#/paths/path/to/resource/GET/responses/200/content/json/DataPayload)/my_type]
```

### Reproduction

```json
"my_type": {
    "anyOf": [
        {
            "$ref": "#/components/schemas/MyType"
        },
        {
            "type": "null"
        }
    ]
}
```

```json
"my_type": {
    "oneOf": [
        {
            "$ref": "#/components/schemas/MyType"
        },
        {
            "type": "null"
        }
    ]
}
```

### Package version(s)

<img width="275" height="70" alt="Image" src="https://github.com/user-attachments/assets/dfde9ccb-52af-49ca-93b2-5bdfa7fcbae8" />

### Expected behavior

ideally it should produce an optional. But an enum would also work.

### Environment

swift 6.2
Xcode 26.0 beta 6

### Additional information

The only workaround I've found is to do something like this, but the library our back end is using to produce the openapi spec does not seem to support something like this:

```
"my_type": {
    "$ref": "#/components/schemas/MyType",
    "type": ["object", "null"]
},
```

czechboy0:
Hi @BrentMifsud,

I believe this is a duplicate, we already have one or two issues tracking it, I believe you already commented on them as well.

This is known - I'd recommend removing any explicit null types from the OpenAPI doc, and instead just rely on the `required` property in your object schemas. If a property is present, it'll be non-optional. If it's absent, it'll be optional - and that'll handle automatically both the case where the field is missing from the JSON and if it's present as null. You don't need to explicitly allow null in the OpenAPI doc, in fact that causes the issue you're seeing.

czechboy0:
In your example, I'd just use:

```
"my_type": {
    "$ref": "#/components/schemas/MyType"
},
```

czechboy0:
Related, #419, #513.

vdsingh:
Hey there 👋🏽 . Is there any update on this issue? For context, in my project we're using Pydantic and [the OpenAPI document is generated](https://pydantic.dev/docs/validation/latest/concepts/json_schema/) from that. I.e., we can't just update our open api document and fix things that way.

[...]

Issue #565, opened 2024-04-18
bug/regression: latest yams results in incomplete generation of types


dpasirst:
### Description

I'm not exactly sure where this bug is arising from (e.g. swift-openapi-generator or one of its dependencies like OpenAPIKit).

In short, using `swift-openapi-generator` with when the `yams` version is 5.0.X, the correct swift code (Types) is generated. When yams is 5.1.X, the generated code is missing fields. This is with an OpenAPI 3.1 json.

### Reproduction

I have attached an example openapi.json along with the difference in the resulting generated swift code. Specifically, in looking at `Components.Schemas.NewAccount` and `Components.Schemas.Account` you will see that the one created with yams 5.0.X contains a `phoneNumberPayload` as part of each.  There is also a defined `PhoneNumber` struct and curiously it is not linked or used instead of `phoneNumberPayload` but that is perhaps a different improvement. For now, the `phoneNumberPayload` is usable with yams 5.0.X.

However, when building with yams 5.1.X, the generated `NewAccount` and `Account` do not contain `phoneNumberPayload`. It is dropped but should be there.

I have attached the example openapi.json and the example generated outputs.

[openapi.json](https://github.com/apple/swift-openapi-generator/files/15029882/openapi.json)
[generated_types_yams5.0.swift.txt](https://github.com/apple/swift-openapi-generator/files/15029884/generated_types_yams5.0.swift.txt)
[generated_types_yams5.1.swift.txt](https://github.com/apple/swift-openapi-generator/files/15029886/generated_types_yams5.1.swift.txt)

I'm not sure if this is related but Yams 5.1.0 release notes the following breaking change:

> Change how empty strings are decoded into nullable properties.
key: "" previously decoded into
struct Value: Codable { let key: String? } as Value(key: nil)
whereas after this change it decodes as Value(key: "").
This could be a breaking change if you were relying on the previous
semantics.

If it is helpful, the openapi.json was generated via Rust using Aide with Axum.
```rust
#[derive(Default, Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash, JsonSchema)]
pub enum Role {
    #[default]
    #[serde(rename="user")]
    User,
    #[serde(rename="admin")]
    Admin,
}

#[derive(Default, Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash, JsonSchema)]
pub struct PhoneNumber {
    /// country code must contain only a set of digits 0-9
[...]

PR #933, opened 2026-08-06
Support type: null in anyOf/oneOf/allOf as a nullability marker

ThomasDutartre:
### Motivation

In OpenAPI 3.1, a nullable value is commonly expressed by adding a `{ "type": "null" }` branch to an `anyOf`/`oneOf`/`allOf` (e.g. `anyOf: [$ref, {type: null}]`). Today the generator treats the `null` branch as an unsupported member, so the whole composition is skipped and emits `Schema "null" is not supported ... skipping`. This blocks common specs (FastAPI, Figma, and others) and is tracked in #906, #817, #419, #565, #513, #286.

OpenAPIKit already propagates the nullability onto the enclosing composition's core context when one of its subschemas is `null`, so the generator only needs to stop rejecting these schemas and skip the `null` member. This is the minimal, non-collapsing approach previously suggested by the maintainers, which keeps API evolution deterministic (adding a member expands the wrapper instead of changing its shape).

#### Prior maintainer guidance (the "filter" approach)

This PR intentionally implements the filtering approach that @czechboy0 described earlier, rather than the collapsing/optional-root approach of #557 and #558:

- #557 (comment): _"we should simply filter out null from anyOf/allOf schemas, as OpenAPIKit will already represent the nullability on the parent schema"_ — https://github.com/apple/swift-openapi-generator/pull/557#issuecomment-2034640640
- #558 (comment): _"does that mean we just need to filter out the null case in the generator and the rest should work?"_ — https://github.com/apple/swift-openapi-generator/pull/558#issuecomment-2034565928
- #558 (comment): _"would you be interested in opening a PR for the filtering of nulls ..."_ — https://github.com/apple/swift-openapi-generator/pull/558#issuecomment-2034659980

#### Minimal example that should be supported

```json
{
  "components": {
    "schemas": {
      "Child": { "type": "string" },
      "Parent": {
        "type": "object",
        "properties": {
          "child": {
            "anyOf": [
              { "$ref": "#/components/schemas/Child" },
              { "type": "null" }
            ]
          }
        }
      }
    }
  }
}
```

### Modifications

- Added `JSONSchema.isNullType` helper.
- `isSchemaSupported`: filter out the `null` branch of `allOf`/`anyOf`/`oneOf` before checking support, so the composition is generated instead of skipped.
- `translateAllAnyOneOf`: skip the `null` branch when generating members/cases (no member/case is emitted for it).
- `translateArray`: preserve the element's optionality (`[Element?]`), consistent with how builtin element types are already emitted, so arrays that mix values and `null` are supported.

### Result

- `anyOf`/`oneOf`/`allOf` containing a `type: null` branch are now generated as an optional wrapper (e.g. `var b: A.bPayload?`) instead of being skipped.
- The wrapper is not collapsed: `oneOf: [$ref, null]` yields a single-case optional enum, and adding a case simply expands it — API evolution stays deterministic.
- Arrays of nullable elements are emitted as `[Element?]`, preserving order while allowing `null` entries.
- Non-nullable compositions and arrays are unchanged.

### Test Plan

- Added snippet tests in `SnippetBasedReferenceTests` for `oneOf`/`anyOf` with `null` (single and multiple refs) and for an array with a nullable `oneOf` element.
- `swift test`: 329 tests, 0 failures.
- Validated end-to-end against a real-world spec: generation succeeds with no `Schema "null" is not supported` warnings, and nullable properties/array elements are emitted as optionals.

### DIFF
--- a/Sources/_OpenAPIGeneratorCore/Extensions/OpenAPIKit.swift
+++ b/Sources/_OpenAPIGeneratorCore/Extensions/OpenAPIKit.swift
@@ -109,4 +109,10 @@ extension JSONSchema {
 
     /// Returns a human-readable description of the schema.
     var prettyDescription: String { value.prettyDescription }
+
+    /// A Boolean value indicating whether the schema is the `null` type.
+    var isNullType: Bool {
+        if case .null = value { return true }
+        return false
+    }
 }
--- a/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateAllAnyOneOf.swift
+++ b/Sources/_OpenAPIGeneratorCore/Translator/CommonTranslations/translateAllAnyOneOf.swift
@@ -40,6 +40,8 @@ extension TypesFileTranslator {
     func translateAllOrAnyOf(typeName: TypeName, openAPIDescription: String?, type: AllOrAnyOf, schemas: [JSONSchema])
         throws -> Declaration
     {
+        // A `null` branch only marks the composition as nullable, it has no member.
+        let schemas = schemas.filter { !$0.isNullType }
         let properties: [(property: PropertyBlueprint, isKeyValuePair: Bool)] = try schemas.enumerated()
             .map { index, schema in
                 let key = "value\(index+1)"
@@ -129,6 +131,8 @@ extension TypesFileTranslator {
         discriminator: OpenAPI.Discriminator?,
         schemas: [JSONSchema]
     ) throws -> Declaration {
+        // A `null` branch only marks the oneOf as nullable, it has no case.
+        let schemas = schemas.filter { !$0.isNullType }
         let cases: [(String, [String]?, Bool, Comment?, TypeUsage, [Declaration])]
         if let discriminator {
             // > When using the discriminator, inline schemas will not be considered.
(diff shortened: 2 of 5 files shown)


saikatdas0790:
Any movement on this? Would appreciate this getting merged in

czechboy0:
Hi @ThomasDutartre,

Thanks! 🙏

ThomasDutartre:
> Hi @ThomasDutartre,
> 
> we're formalizing how we organize issues and PRs, and we now expect an issue to be assigned to a contributor before a PR is opened. Details in https://github.com/apple/swift-openapi-generator/blob/main/CONTRIBUTING.md#how-to-contribute-your-work. So I'll close this PR for now.
> 
> If you're still interested in working on this issue, let's discuss next steps in #906.
> 
> Thanks! 🙏

I literally quoted all the linked issues, even [your comment](https://github.com/apple/swift-openapi-generator/pull/557#issuecomment-2034640640) that suggested this approach

czechboy0:
It's not about quoting, it's about assigning an issue to a specific person before a PR is opened. Otherwise we get a lot of duplicated PRs and it's quite time-consuming on the maintainers to resolve and ensure everyone gets the credit for the work they did.

PR #947, opened 2026-09-03
Ignore UnusedImportAccess warnings on generated access-level imports

### LINKED ISSUE #946: Add annotations to ignore unused import warnings in generated files
### Description

Example below. The update from 1.13.0 to 1.13.1 broke our CI (where all warnings as errors is set) as there is a new warning in the generated code.

```
...../destination/OpenAPIGenerator/GeneratedSources/Types+Components+RequestBodies.swift:4:25: error: package import of 'Foundation' was not used in package declarations
```

For now I can disable that option on our build but it would be great (if the warnings cannot be removed) if they could be specifically ignored with `.enableExperimentalFeature("SourceWarningControl")`?

### Reproduction

Generated code made with 1.13.0 vs 1.13.1 changes layout and now some imports are unused in some files.

### Package version(s)

```
swift-openapi-generator.git@1.13.1>(traits: default)
```

### Expected behavior

I expect generated code to be warning-free, or have known warnings specifically suppressed.

### Environment

macOS + swift from Xcode

### Additional information

_No response_

tirdesh:
### Motivation

Fixes #946.

The same imports go into every generated file, so an access-level import is often unused in a
given one. Since 1.13.1 changed the file layout, that surfaces as:

```
error: package import of 'Foundation' was not used in package declarations [#UnusedImportAccess]
```

which breaks any package built with `-warnings-as-errors`.

### Modifications

The renderer emits `@diagnose(UnusedImportAccess, as: ignored)` for access-level imports
behind `#if hasFeature(SourceWarningControl)`, following the swift-crypto pattern
@simonjbeaumont pointed to on the issue. Only an access-level import can trip the diagnostic,
so the attribute follows the access modifier.

### Result

On a compiler with SE-0522 the warning is gone, and nothing in the adopter's package has to
change. Older compilers take the `#else` branch and get the imports they get today.

### Test Plan

`swift test`: 334 tests, 0 failures. The new test fails before the change, and the Petstore
references are regenerated rather than hand-edited.

Both branches are checked against real toolchains. `hasFeature(SourceWarningControl)` is
false on 6.3.3, so released compilers take the `#else` branch. On
`swift-DEVELOPMENT-SNAPSHOT-2026-08-30-a` it is true, and there an unused `package import`
under `-warnings-as-errors` fails with the error above while the same import carrying the
attribute compiles.


czechboy0 (commented):
In Swift Crypto, they add the SourceWarningControl experimental feature in Package.swift, but in this PR we don't: https://github.com/apple/swift-crypto/commit/15b0fb5b45380caaf5e51be2377eede9a1adf629#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677eR44

Can you check if it's required or not to make it correctly suppress the warnings?

alexookah (approved):
LGTM

PR #951, opened 2026-09-21
Box schemas whose estimated inline size exceeds a limit

jpsim:
### Motivation

Generated schemas can sometimes be quite large and lead to heavier binaries, stack pressure and even crashes due to stack overflows.

### Modifications

Add a `--max-inline-schema-size <bytes>` cli option and corresponding `maxInlineSchemaSize` config key. It uses heuristics to estimate each codegened schema's `MemoryLayout<T>.size`, and applies the same boxing mechanism that already exists for recursive types to these large types. This also plays nicely with the boxed types due to recursion, so we don't box unnecessarily. This is entirely opt-in, so by default users get no boxing for these large types, maintaining the same behavior as today.

The size estimator runs bottom up, checking inner-most type references first, walking up the hierarchy until the size limit is exceeded, which leads to boxing far fewer types to stay under the limit.

While I was here, I also fixed the [Mirror](https://developer.apple.com/documentation/swift/mirror) implementation for boxed types to have them reflect their stored properties, minimizing the impact of boxing on consumers. This also applies the same mirroring improvements to recursive schemas.

Please let me know if you'd like this to go through a formal SOAR proposal.

### Result

Measured in Ramp's iOS app:
* Binary size of output codegen when compiled dropped by 25%
* Largest type memory size went from 19 KB to 1 KB
* Types over 4 KB went from 87 to 0
* Only 40 types were boxed, even though 250 types exceeded the 1KB limit

Here's the impact before and after boxing in our app:

image

### Test Plan

* Added new snippet tests
* No impact to callsites in our iOS app


czechboy0:
Hi @jpsim, yeah could you file an issue about the problem you're seeing first and we can discuss next steps there. If you could share more numbers about the depth of the OpenAPI types you're working with, etc, that'd be great. Or if your OpenAPI doc is public, sharing it would help too.

czechboy0 (changes requested):
Marking as blocked on the proposal

